PR #553, opened 2025-03-24
feat(kmod): get kmod version

Koichi98:
## Description
Created a version.txt file containing version information and modified the Makefile and debian/rules to reference it for retrieving the kernel module version. Also, made changes to display the version information when the kernel module is installed.
This is a PR separated from #545 to make it smaller.

## Related links
PR #545

## How was this PR tested?

Checked that the version is acquired correctly by running insmod + dmesg.

- [x] Autoware (required)
- [ ] sample application

## Notes for reviewers
I wanted to modify dkms.conf to retrieve the version from version.txt as well, but DKMS identifies the build directory using the pattern {PACKAGE_NAME}-{VERSION_NAME}, so I had no choice but to directly provide the version name. There might be a way to pass it through debian/rules, but that remains a TODO item.


sykwer:
Great!

PR #626, opened 2025-04-21
test(heaphook): add unit test for posix_memalign function

Enigamict:
## Description
implementation of the unit test of the posix_memalign function included in agnocast_heaphook

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

### DIFF
--- a/agnocast_heaphook/src/lib.rs
+++ b/agnocast_heaphook/src/lib.rs
@@ -587,4 +587,37 @@ mod tests {
 
         unsafe { free(new_ptr) };
     }
+
+    #[test]
+    fn test_posix_memalign_normal() {
+        // Arrange
+        let start = MEMPOOL_START.load(Ordering::SeqCst);
+        let end = MEMPOOL_END.load(Ordering::SeqCst);
+        let alignment = 64;
+        let size = 512;
+        let mut ptr: *mut c_void = std::ptr::null_mut();
+
+        // Act
+        let r = posix_memalign(&mut ptr, alignment, size);
+
+        // Assert
+        assert_eq!(r, 0, "posix_memalign should return 0 on success");
+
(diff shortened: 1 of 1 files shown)


veqcc (dismissed):

agnocast_heaphook/src/lib.rs
```diff
@@ -587,4 +587,50 @@ mod tests {
...
+
+        unsafe {
+            for i in 0..size {
+                *((ptr as *mut u8).add(i)) = (i % 255) as u8;
+            }
+            for i in 0..size {
+                assert_eq!(
+                    *((ptr as *const u8).add(i)),
+                    (i % 255) as u8,
+                    "posix_memalign memory should be writable and readable at position {}",
```
veqcc, line 629: Is it a property only for posix_memalign?
I think it also applies to other functions.

Enigamict: Sorry, that's right 
other functions behave the same way as well, so it's not only posix_memalign as you pointed out 
Should I fix the error message?

veqcc: I doubt the assertion itself is needed :thinking:

Enigamict: I see🫨
OK, Is this assertion test itself not necessary?
Is this enough to prove that posix_memalign is working?
```
        assert_eq!(
            ptr as usize % alignment,
            0,
            "posix_memalign memory should be aligned to the specified boundary"
        );
```

veqcc: I think its unnecessary. You can remove it.

veqcc (approved):

atsushi421 (approved):

Koichi98 (approved):

PR #642, opened 2025-05-20
test(heaphook): fix unit test for test_aligned_alloc function

Enigamict:
## Description
The test_aligned_alloc function to properly comply with the C standard specification for aligned_alloc(). In the previous implementation, the sizes used in the tests were not multiples of the alignment values.

> The function aligned_alloc() is the same as memalign(), except for the added restriction that size should be a multiple of alignment.

(https://linux.die.net/man/3/aligned_alloc)
## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

### DIFF
--- a/agnocast_heaphook/src/lib.rs
+++ b/agnocast_heaphook/src/lib.rs
@@ -622,26 +622,31 @@ mod tests {
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


veqcc (approved):
Great :+1:

PR #647, opened 2025-06-02
Remove unnecessary padding

r1ru:
## Description
現状のコードでは、 ポインタを書き込むために追加で `ALIGNMENT(64)`バイトの領域を確保しており、かつ `tlsf_allocate`では`ALIGNMENT(64)`がアラインメント制約として指定されています:

https://github.com/tier4/agnocast/blob/main/agnocast_heaphook/src/lib.rs#L223
https://github.com/tier4/agnocast/blob/main/agnocast_heaphook/src/lib.rs#L180-L186

`ALIGNMENT=1`に変更し、追加で確保する領域を `POINTER_SIZE`バイトに修正することでメモリの無駄を削減しました。

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

## Notes for reviewers

- `malloc` 関数はRustのデフォルトのグローバルアロケータで使用されており、x64環境では `MIN_ALIGN`バイトにアラインメントされたアドレスを返すことが仮定されているため、これを満たす必要があります:
https://doc.rust-lang.org/beta/src/std/sys/alloc/unix.rs.html#13-15


agnocast_heaphook/src/lib.rs
```diff
@@ -242,9 +248,20 @@ fn tlsf_reallocate_wrapped(ptr: usize, size: usize) -> *mut c_void {
...
+    // See: https://doc.rust-lang.org/beta/src/std/sys/alloc/unix.rs.html#13-15
+    let alignment = MIN_ALIGN;
+    debug_assert!(alignment.is_power_of_two() && alignment >= 8);
+
     // return value from internal alloc
-    let start_addr: usize = tlsf_reallocate(original_start_addr_ptr, ALIGNMENT + size) as usize;
-    let aligned_addr: usize = start_addr + ALIGNMENT;
+    //
+    // It is our responsibility to satisfy alignment constraints.
+    // We avoid using `Layout::align` because doing so requires us to remeber the alignment.
```
veqcc, line 259:
```suggestion
    // We avoid using `Layout::align` because doing so requires us to remember the alignment.
```

agnocast_heaphook/src/lib.rs
```diff
@@ -20,7 +20,8 @@ extern "C" {
 }
 
 const POINTER_SIZE: usize = std::mem::size_of::<&usize>();
-const ALIGNMENT: usize = 64; // must be larger than POINTER_SIZE
+const ALIGNMENT: usize = 1;
```
veqcc, line 23: Is `ALIGNMENT` still used?

agnocast_heaphook/src/lib.rs
```diff
@@ -20,7 +20,8 @@ extern "C" {
...
+const MIN_ALIGN: usize = 16; // See: https://doc.rust-lang.org/src/std/sys/alloc/mod.rs.html
```
veqcc, line 24: If you use this fixed value, then agnocast-heaphook msut clarify that it only supports `target_arch = "x86_64"` and should fail on other architectures while initializing.

agnocast_heaphook/src/lib.rs
```diff
@@ -219,15 +220,20 @@ fn tlsf_deallocate(ptr: std::ptr::NonNull<u8>) {
...
     // aligned address returned to user
-    let aligned_addr: usize = if alignment == 0 {
-        start_addr + ALIGNMENT
-    } else {
-        start_addr + ALIGNMENT + alignment - ((start_addr + ALIGNMENT) % alignment)
-    };
+    //
+    // It is our responsibility to satisfy alignment constraints.
+    // We avoid using `Layout::align` because doing so requires us to remeber the alignment.
+    // This is because `Tlsf::{reallocate, deallocate}` functions require the same alignment.
```
veqcc, line 234: Could you elaborate the comments? I couldn't understand it well.
For example, why `Tlsf::deallocate` needs alignment?

r1ru: Because the doc says so:
> The memory block must have been allocated with the same alignment ([Layout::align](https://doc.rust-lang.org/nightly/core/alloc/layout/struct.Layout.html#method.align)) as align.

https://docs.rs/rlsf/latest/rlsf/struct.Tlsf.html#method.deallocate

agnocast_heaphook/src/lib.rs
veqcc: Could you explain why `ALIGNMENT = 1` works fine for `Layout::from_size_align(size, ALIGNMENT)` ?

r1ru: Because 1 != 0 and 2^0 = 1.

agnocast_heaphook/src/lib.rs
veqcc: Currently, we have checked only `x86_64`, but we want to use Agnocast on `aarch64` in future.

agnocast_heaphook/src/lib.rs
veqcc: I understand it needs to be a power of 2.
In this PR, you introduced `MIN_ALIGN = 16`, but you are still using `ALIGNMENT = 1` in some functions. I suspect it works fine.

r1ru: Only the  `malloc` function needs to use `MIN_ALIGN`. As explained above, this is because Rust's global allocator expects `malloc` to return an address aligned to `MIN_ALIGN` bytes. That is why we specify `MIN_ALIGN` in our `malloc` implementation:

```rust
#[no_mangle]
pub extern "C" fn malloc(size: usize) -> *mut c_void {
    if should_use_original_func() {
        return unsafe { (*ORIGINAL_MALLOC.get_or_init(init_original_malloc))(size) };
    }

    // The default global allocator assumes `malloc` returns 16-byte aligned address (on x64 platforms).
    // See: https://doc.rust-lang.org/beta/src/std/sys/alloc/unix.rs.html#13-15
    tlsf_allocate_wrapped(MIN_ALIGN, size)
}
```

veqcc: Then, I have another question:
Isn't it necessary to allign with at least `MIN_ALIGN` when `calloc` or `alligned_alloc` with smaller alignement (e.g. 8) ?

agnocast_heaphook/src/lib.rs
r1ru: Fixed: https://github.com/tier4/agnocast/pull/647/commits/69b888d8746b174d9824560cbbdb1d9db722d5d8

agnocast_heaphook/src/lib.rs
r1ru: We have to ensure the  **user pointer** to be  `MIN_ALIGN`-byte aligned. We cannot ensure this by replacing `ALIGNMENT` with `MIN_ALIGN`, because the user pointer is calculated as:
```rust
let aligned_addr: usize = (start_addr + POINTER_SIZE + alignment - 1) & !(alignment - 1);
```
Even if `start_addr` is `MIN_ALIGN`-byte aligned, `aligned_addr` is not.

We need to satisfy the following two requirements:

1. The alignment passed to `deallocate` and `reallocate` must be the same as the one specified at `allocate`.
2. The `malloc`, `calloc` and `realloc` functions must return `MIN_ALIGN`-byte aligned addresses.

We use `ALIGNMENT` to satisfy 1, and `MIN_ALIGN` to satisfy 2. The reason we need to use `MIN_ALIGN` is specific to Rust's requirements, so I believe it is appropriate to keep it separate from `ALIGNMENT`.

veqcc: Understood.
Then, could you rename `ALIGNMENT` to make its purpose clear? Like `LAYOUT_ALIGNMENT`.
That's all my review, thank you for your patient explanations!!

r1ru: fixed: https://github.com/tier4/agnocast/pull/647/commits/e9b1e0398fcfd3e7b67d31e36f9e493101618705

veqcc (dismissed):

agnocast_heaphook/src/lib.rs
```diff
@@ -20,7 +20,16 @@ extern "C" {
 }
 
 const POINTER_SIZE: usize = std::mem::size_of::<&usize>();
-const ALIGNMENT: usize = 64; // must be larger than POINTER_SIZE
+const LAYOUT_ALIGN: usize = 1;
```
atsushi421, line 23:
```suggestion
const LAYOUT_ALIGN: usize = 1; // Minimun value that is a power of 2.
```

agnocast_heaphook/src/lib.rs
```diff
@@ -215,19 +224,24 @@ fn tlsf_reallocate(ptr: std::ptr::NonNull<u8>, size: usize) -> *mut c_void {
 
 fn tlsf_deallocate(ptr: std::ptr::NonNull<u8>) {
     let mut tlsf = TLSF.get().unwrap().lock().unwrap();
-    unsafe { tlsf.deallocate(ptr, ALIGNMENT) }
+    unsafe { tlsf.deallocate(ptr, LAYOUT_ALIGN) }
 }
 
 fn tlsf_allocate_wrapped(alignment: usize, size: usize) -> *mut c_void {
+    // the alignment must be greater than 8 to ensure that `start_addr_ptr` is 8-byte aligned.
+    let alignment = alignment.max(8);
```
atsushi421, line 232: Could you define a new variable and use it to avoid magic number?

r1ru: fixed: https://github.com/tier4/agnocast/pull/647/commits/35f83443a1e8bf09f2b58c0a697e8012df20475e

atsushi421 (dismissed):

r1ru:
以下のチェックでエラーになりましたが、これは明示的にしておいたほうが良いと思うため、  `#[allow(clippy::if_same_then_else)]`を追加しました。

https://github.com/tier4/agnocast/actions/runs/15600176579/job/43938492378

Koichi98 (approved):

veqcc (approved):

PR #652, opened 2025-06-23
Fix for memory limit error

Enigamict:
## Description
In the previous implementation, when memory allocation failed inside tlsf_allocate(), a panic! was triggered. However, this caused a problem where the panic occurred while holding a lock, leading to a deadlock situation that required forcibly terminating the process.
```
let ptr: std::ptr::NonNull<u8> = tlsf.allocate(layout).unwrap_or_else(|| {
        panic!("[ERROR] [Agnocast] memory allocation failed: use larger AGNOCAST_MEMPOOL_SIZE");
});
```

This update changes the behavior so that when the TLSF allocator fails to allocate memory due to exceeding the memory limit, it now returns NULL instead of panicking. This avoids triggering a panic while holding a lock.

This behavior aligns with that of the standard malloc function, allowing the caller to handle allocation failures appropriately.

## Example Test case

The following example shows that when the allocation size exceeds the memory limit, it correctly displays "FAILED".
```
running 6 tests
test tests::test_calloc_normal ... ok
test tests::test_malloc_normal ... FAILED
test tests::test_realloc_normal ... ok
test tests::test_aligned_alloc_normal ... ok
test tests::test_memalign_normal ... ok
test tests::test_posix_memalign_normal ... ok

failures:

---- tests::test_malloc_normal stdout ----
thread 'tests::test_malloc_normal' panicked at src/lib.rs:538:9:
allocated memory should not be null
```

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application


agnocast_heaphook/src/lib.rs
```diff
@@ -188,21 +193,15 @@ fn init_tlsf() {
...
-    let ptr: std::ptr::NonNull<u8> = tlsf.allocate(layout).unwrap_or_else(|| {
-        panic!("[ERROR] [Agnocast] memory allocation failed: use larger AGNOCAST_MEMPOOL_SIZE");
-    });
-
-    ptr.as_ptr() as *mut c_void
+fn tlsf_allocate(size: usize) -> Result<NonNull<u8>, AgnocastError> {
+    if size > isize::MAX as usize {
+        Err(AgnocastError::OutOfMemory)
+    } else {
+        debug_assert!(size <= isize::MAX as usize && LAYOUT_ALIGN == 1);
```
veqcc, line 200: `size <= isize::MAX as usize` is not necessary because it's in the else branch of `if size > isize::MAX as usize { ...`

agnocast_heaphook/src/lib.rs
```diff
@@ -4,12 +4,17 @@ use std::{
...
+    ptr::NonNull,
     sync::{
         atomic::{AtomicBool, AtomicUsize, Ordering},
         Mutex, OnceLock,
     },
 };
 
+enum AgnocastError {
+    OutOfMemory,
+}
```
veqcc, line 16: I think `AgnocastError` is not necessary only for `OutOfMemory`.
Why don't you just return null in `tlsf_allocate`?

Enigamict: I Fixed.
That's true. No need for the enum with just one error type. Reverted to simply returning NULL

agnocast_heaphook/src/lib.rs
Enigamict: I Fixed.

Enigamict:
I found that the memory limit check was still present in realloc, so I’ve updated it accordingly.

agnocast_heaphook/src/lib.rs
```diff
@@ -188,40 +189,24 @@ fn init_tlsf() {
...
-
-    let new_ptr: std::ptr::NonNull<u8> = unsafe {
-        tlsf.reallocate(ptr, layout).unwrap_or_else(|| {
-            panic!("[ERROR] [Agnocast] memory allocation failed: use larger AGNOCAST_MEMPOOL_SIZE");
-        })
-    };
-
-    new_ptr.as_ptr() as *mut c_void
+fn tlsf_reallocate(ptr: std::ptr::NonNull<u8>, size: usize) -> Option<NonNull<u8>> {
+    if size > isize::MAX as usize {
```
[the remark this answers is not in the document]
Enigamict: This is intended to prevent abnormally large allocations that exceed isize::MAX (as an overflow safeguard).
Is this realloc and malloc check mostly meaningless?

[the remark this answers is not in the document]
Enigamict: `from_size_align()`performs a more comprehensive internal overflow check than a simple`isize::MAX`comparison, and may return `Err` if the size is too large. Removed redundant manual checks and rely on proper verification of `from_size_align`. 
Thank you.

agnocast_heaphook/src/lib.rs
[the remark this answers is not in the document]
Enigamict: ありがとうございます。修正しました！

agnocast_heaphook/src/lib.rs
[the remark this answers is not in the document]
veqcc: Thank you!!

Enigamict:
@veqcc 
Yes, that's the plan. I'll include the unit tests in a separate PR.

veqcc (approved):

agnocast_heaphook/src/lib.rs
```diff
@@ -427,6 +416,11 @@ pub extern "C" fn posix_memalign(memptr: &mut *mut c_void, alignment: usize, siz
         };
     }
 
+    let ptr = tlsf_allocate_wrapped(alignment, size);
+
+    if ptr.is_null() {
+        return libc::ENOMEM;
+    }
     *memptr = tlsf_allocate_wrapped(alignment, size);
```
atsushi421, line 424:
```suggestion
    *memptr = ptr;
```

atsushi421 (approved):

Koichi98 (approved):

veqcc (approved):
@Enigamict 
Since this PR block another PR to move on, we will merge it with minor fixes.
Thank you for your contribution!! :+1:

PR #653, opened 2025-06-30
refactor(heaphook): modify tlsf_{allocate, reallocate}_wrapped to return Option<NonNull<u8>>

r1ru:
## Description

- `tlsf_{allocate, reallocate}_wrapped`関数が`Option<NonNull<u8>>`を返すように修正し、Cの型に依存しないようにしました
- ポインタを`usize`ではなく`NonNull<u8>`として扱うように修正しました

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

### DIFF
--- a/agnocast_heaphook/src/lib.rs
+++ b/agnocast_heaphook/src/lib.rs
@@ -221,8 +221,10 @@ fn tlsf_allocate_wrapped(alignment: usize, size: usize) -> Option<NonNull<u8>> {
     // We avoid using `Layout::align` because doing so requires us to remember the alignment.
     // This is because `Tlsf::{reallocate, deallocate}` functions require the same alignment.
     let aligned_addr = (original_addr + POINTER_SIZE + alignment - 1) & !(alignment - 1);
+
+    // SAFETY: `aligned_addr` must be non-zero.
     debug_assert!(aligned_addr % alignment == 0 && aligned_addr != 0);
-    let aligned_ptr = NonNull::new(aligned_addr as *mut u8).unwrap();
+    let aligned_ptr = unsafe { NonNull::new_unchecked(aligned_addr as *mut u8) };
 
     // store the original pointer
     unsafe { *aligned_ptr.as_ptr().byte_sub(POINTER_SIZE).cast() = original_ptr };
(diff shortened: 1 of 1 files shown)


agnocast_heaphook/src/lib.rs
```diff
@@ -195,83 +195,74 @@ fn tlsf_allocate(size: usize) -> Option<NonNull<u8>> {
...
+    // aligned pointer returned to user
     //
     // It is our responsibility to satisfy alignment constraints.
     // We avoid using `Layout::align` because doing so requires us to remember the alignment.
     // This is because `Tlsf::{reallocate, deallocate}` functions require the same alignment.
-    let aligned_addr: usize = (start_addr + POINTER_SIZE + alignment - 1) & !(alignment - 1);
-    debug_assert!(aligned_addr % alignment == 0);
+    let aligned_addr = (original_addr + POINTER_SIZE + alignment - 1) & !(alignment - 1);
+    debug_assert!(aligned_addr % alignment == 0 && aligned_addr != 0);
+    let aligned_ptr = NonNull::new(aligned_addr as *mut u8).unwrap();
```
veqcc, line 225: Could you return `None` instead of `unwrap()` ?

Koichi98:
The code changes seem to be fine once the unwrap issue mentioned above is fixed.

I've updated the PR title to align with [this standard](https://github.com/commitizen/conventional-commit-types/blob/c3a9be4c73e47f2e8197de775f41d981701407fb/index.json). (While not strictly adhering to it, I believe it's best to follow it as much as possible for better clarity when viewing a list of PRs later.)

agnocast_heaphook/src/lib.rs
r1ru: Then, should I replace `debug_assert` with `assert`?
I think returning `None` instead of panicking is not a good idea, because if `aligned_ptr` ever becomes zero, it would indicate a bug in the internal memory allocator. It's better not to hide such a bug.

r1ru: > Then, should I replace debug_assert with assert?

Actually, this change might be unnecessary, since if `aligned_ptr` ever becomes zero, `NonNull::new(aligned_addr as *mut u8).unwrap()` will panic.

r1ru: fixed: https://github.com/tier4/agnocast/pull/653/commits/3b3d8cf77e3791739ef1d962992a0a447282ce85

veqcc (approved):

agnocast_heaphook/src/lib.rs
veqcc: Thank you!

atsushi421 (approved):

Koichi98 (approved):

PR #657, opened 2025-07-10
fix(heaphook): fix memory-related functions to match C semantics

r1ru:
## Description
Fixed memory-related functions to match the semantics of C:

- Modified to use `Layout` to ensure the alignment is a power of two.
- Fixed `realloc(ptr, size)` so that when ptr is not `NULL` and size is zero, it behaves equivalently to `free(ptr)`.
- Fixed `aligned_alloc` so that the size is a multiple of the alignment.
- Fixed `posix_memalign` so that the alignment is a multiple of two and also a multiple of `sizeof(void *)`.

See:
https://linux.die.net/man/3/posix_memalign
https://linux.die.net/man/3/malloc

I will create the unit tests in a separate PR.

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

### DIFF
--- a/agnocast_heaphook/src/lib.rs
+++ b/agnocast_heaphook/src/lib.rs
@@ -382,18 +382,31 @@ pub unsafe extern "C" fn realloc(ptr: *mut c_void, new_size: usize) -> *mut c_vo
         (true, false) => {
             // The default global allocator assumes `realloc` returns 16-byte aligned address (on x64 platforms).
             // See: https://doc.rust-lang.org/beta/src/std/sys/alloc/unix.rs.html#53-54
-            match Layout::from_size_align(new_size, MIN_ALIGN) {
-                Ok(new_layout) => match NonNull::new(ptr.cast()) {
-                    Some(non_null_ptr) => match tlsf_reallocate_wrapped(non_null_ptr, new_layout) {
+            let new_layout = match Layout::from_size_align(new_size, MIN_ALIGN) {
+                Ok(layout) => layout,
+                Err(_) => return ptr::null_mut(),
+            };
+
+            match NonNull::new(ptr.cast()) {
+                Some(non_null_ptr) => {
+                    // If size is equal to zero, and ptr is not NULL, then the call is equivalent to free(ptr).
+                    if new_layout.size() == 0 {
+                        tlsf_deallocate_wrapped(non_null_ptr);
+                        return ptr::null_mut();
+                    }
+
+                    match tlsf_reallocate_wrapped(non_null_ptr, new_layout) {
                         Some(non_null_ptr) => non_null_ptr.as_ptr().cast(),
                         None => ptr::null_mut(),
-                    },
-                    None => match tlsf_allocate_wrapped(new_layout) {
+                    }
+                }
+                None => {
+                    // If ptr is NULL, then the call is equivalent to malloc(size).
+                    match tlsf_allocate_wrapped(new_layout) {
                         Some(non_null_ptr) => non_null_ptr.as_ptr().cast(),
                         None => ptr::null_mut(),
-                    },
-                },
-                Err(_) => ptr::null_mut(),
+                    }
+                }
             }
         }
         (false, _) => (*ORIGINAL_REALLOC.get_or_init(init_original_realloc))(ptr, new_size),
--- a/agnocast_heaphook/src/lib.rs
+++ b/agnocast_heaphook/src/lib.rs
@@ -422,7 +422,7 @@ pub extern "C" fn posix_memalign(memptr: &mut *mut c_void, alignment: usize, siz
     }
 
     // `alignment` must be a power of two and a multiple of `sizeof(void *)`.
-    if !alignment.is_power_of_two() || alignment % size_of::<*mut c_void>() != 0 {
+    if !alignment.is_power_of_two() || alignment & (size_of::<*mut c_void>() - 1) != 0 {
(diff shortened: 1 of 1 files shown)


agnocast_heaphook/src/lib.rs
```diff
@@ -310,14 +304,19 @@ pub extern "C" fn malloc(size: usize) -> *mut c_void {
...
+    match tlsf_allocate_wrapped(layout) {
         Some(non_null_ptr) => non_null_ptr.as_ptr().cast(),
-        None => std::ptr::null_mut(),
+        None => ptr::null_mut(),
     }
 }
 
 #[inline]
-fn is_shared(ptr: *mut c_void) -> bool {
+fn is_shared(ptr: *mut u8) -> bool {
```
veqcc, line 319: Could you explain what is this change for?

agnocast_heaphook/src/lib.rs
```diff
@@ -232,23 +219,28 @@ fn tlsf_allocate_wrapped(alignment: usize, size: usize) -> Option<NonNull<u8>> {
...
-    // The default global allocator assumes `realloc` returns 16-byte aligned address (on x64 platforms).
-    // See: https://doc.rust-lang.org/beta/src/std/sys/alloc/unix.rs.html#53-54
-    let alignment = MIN_ALIGN;
-    debug_assert!(alignment.is_power_of_two() && alignment >= POINTER_ALIGN);
+    // the original pointer returned by the internal allocator
+    let size = new_layout.size();
+    let new_layout = Layout::from_size_align(POINTER_SIZE + size + alignment, LAYOUT_ALIGN).ok()?;
+    let mut tlsf = TLSF.get().unwrap().lock().unwrap();
+    let original_ptr = unsafe { tlsf.reallocate(original_ptr, new_layout) }?;
+    let original_addr = original_ptr.as_ptr() as usize;
```
veqcc, line 236: `original_ptr` is used twice in different meanings.
The current code may work fine, but please name them separately to make it clear :+1:

veqcc:
Though I have made minor comments, the overall changes looks good. Great work :+1:

agnocast_heaphook/src/lib.rs
r1ru: fixed: https://github.com/tier4/agnocast/pull/657/commits/fa40a7f84ec7842babd0dfa913616b657ebda921

agnocast_heaphook/src/lib.rs
r1ru: Later, I will turn this function into a method like the following:

```rust
impl AgnocastSharedMemory {
    fn is_shared(ptr: *mut u8) -> bool {
        ...
    }
    ...
}
```
So it's better to use `u8` instead of `c_void`, because `u8` is more general.

veqcc: Agreed :+1:

veqcc (dismissed):

r1ru:
Sorry, I made a minor change to the comments:
https://github.com/tier4/agnocast/pull/657/commits/e0e647cbbfc24d377eadca5d1eafe14eb64abcb6

r1ru:
@veqcc @atsushi421 
I fixed a division-by-zero bug in `aligned_alloc`:
https://github.com/tier4/agnocast/pull/657/commits/d041fb03af6a59cdf9a702201654d953440efb84
Since I discovered this bug while writing unit tests, I decided to includ the tests in this PR:
https://github.com/tier4/agnocast/pull/657/commits/a1270b5b525896e5893bea16a9145f34113b68c2

agnocast_heaphook/src/lib.rs
```diff
@@ -433,7 +472,17 @@ pub extern "C" fn memalign(alignment: usize, size: usize) -> *mut c_void {
         };
     }
 
-    match tlsf_allocate_wrapped(alignment, size) {
+    // `alignment` must be a power of two.
+    if !alignment.is_power_of_two() {
+        return ptr::null_mut();
+    }
```
veqcc, line 478: I think this is unnecessary because `Layout::from_size_align(size, alignment)` will return `Err` if `alignment` is not a power of two.

agnocast_heaphook/src/lib.rs
```diff
@@ -419,7 +448,17 @@ pub extern "C" fn aligned_alloc(alignment: usize, size: usize) -> *mut c_void {
         };
     }
 
-    match tlsf_allocate_wrapped(alignment, size) {
+    // `alignment` should be a power of two and `size` should be a multiple of `alignment`.
+    if !alignment.is_power_of_two() || size % alignment != 0 {
+        return ptr::null_mut();
+    }
```
veqcc, line 454: Inserting the following check after `layout` is enough.
`%` is relatively costly
```
if size & (alignment - 1) > 0 {
  return ptr::null_mut();
}
```

agnocast_heaphook/src/lib.rs
r1ru: fixed: https://github.com/tier4/agnocast/pull/657/commits/6388048701df25d02aa31ab49978b42e1fa42cba

agnocast_heaphook/src/lib.rs
r1ru: fixed: https://github.com/tier4/agnocast/pull/657/commits/4f32c2ba538a62be0bff6a5d42dcffee5bd2b17c

Koichi98 (approved):

PR #660, opened 2025-07-16
feat(agnocastlib): single threaded executor for cie

sykwer:
## Description
This is the first pull request to begin integrating the [CallbackIsolatedExecutor](https://github.com/tier4/callback_isolated_executor) into Agnocast.  
The `CallbackIsolatedAgnocastExecutor` is designed as "a set of `SingleThreadedAgnocastExecutor` instances, each dedicated to a single callback group."

The challenge in implementing this model lies in the current Agnocast design, which assumes only one executor per process. This assumption makes it impossible to poll only Agnocast callbacks belonging to a specific callback group.

This pull request enables a `SingleThreadedAgnocastExecutor` to be dedicated to a specific callback group.

## Related links

[TIER IV Internal](https://tier4.atlassian.net/wiki/spaces/CRL/pages/3815211404/CallbackIsolatedAgnocastExecutor)

## How was this PR tested?

- [ ] Autoware (required)
- [x] sample application

### DIFF
--- a/src/agnocastlib/include/agnocast/agnocast_executor.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_executor.hpp
@@ -26,6 +26,10 @@ class AgnocastExecutor : public rclcpp::Executor
 
   void wait_and_handle_epoll_event(const int timeout_ms);
   bool get_next_ready_agnocast_executable(AgnocastExecutable & agnocast_executable);
+
+  // Return false iff this Executor is SingleThreadedAgnocastExecutor
+  // and used for internal implementation of CallbackIsolatedAgnocastExecutor
+  // and `group` is "not" the callback group dedicated by this Executor.
   virtual bool validate_callback_group(const rclcpp::CallbackGroup::SharedPtr & group) const = 0;
 
 protected:
--- a/src/agnocastlib/include/agnocast/agnocast_executor.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_executor.hpp
@@ -26,7 +26,7 @@ class AgnocastExecutor : public rclcpp::Executor
 
   void wait_and_handle_epoll_event(const int timeout_ms);
   bool get_next_ready_agnocast_executable(AgnocastExecutable & agnocast_executable);
-  virtual void validate_callback_group(const rclcpp::CallbackGroup::SharedPtr & group) const = 0;
+  virtual bool validate_callback_group(const rclcpp::CallbackGroup::SharedPtr & group) const = 0;
 
 protected:
   int epoll_fd_;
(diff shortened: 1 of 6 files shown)


src/agnocastlib/include/agnocast/agnocast_executor.hpp
```diff
@@ -26,7 +26,11 @@ class AgnocastExecutor : public rclcpp::Executor
 
   void wait_and_handle_epoll_event(const int timeout_ms);
   bool get_next_ready_agnocast_executable(AgnocastExecutable & agnocast_executable);
-  virtual void validate_callback_group(const rclcpp::CallbackGroup::SharedPtr & group) const = 0;
+
+  // Return false iff this Executor is SingleThreadedAgnocastExecutor
```
atsushi421, line 30:
```suggestion
  // Return false if this Executor is SingleThreadedAgnocastExecutor
```

atsushi421:
@sykwer 
Could you address my comment above and [this](https://github.com/tier4/agnocast/actions/runs/16330766004/job/46132501224?pr=660)?

src/agnocastlib/include/agnocast/agnocast_executor.hpp
```diff
@@ -26,7 +26,11 @@ class AgnocastExecutor : public rclcpp::Executor
...
+  // and used for internal implementation of CallbackIsolatedAgnocastExecutor
+  // and `group` is "not" the callback group dedicated by this Executor.
```
Koichi98, line 32: >   // and `group` is "not" the callback group dedicated by this Executor.

I still couldn't understand what this sentence implies. 

>   // and used for internal implementation of CallbackIsolatedAgnocastExecutor

Also, I thought CallbackIsolatedAgnocastExecutor will have its own validate_callback_group implementation, but I couldn't find it in PR #656. Could you briefly explain?

src/agnocastlib/include/agnocast/agnocast_executor.hpp
sykwer: @atsushi421 if and only if

src/agnocastlib/include/agnocast/agnocast_executor.hpp
sykwer: > I still couldn't understand what this sentence implies.

As described in our private design doc, a SingleThreadedAgnocastExecutor used for the implementation of CallbackIsolatedAgnocastExecutor dedicates to one CallbackGroup.

>  Also, I thought CallbackIsolatedAgnocastExecutor will have its own validate_callback_group implementation, but I couldn't find it in PR https://github.com/tier4/agnocast/pull/656. Could you briefly explain?

CallbackIsolatedAgnocastExecutor extends rclcpp:Executor not AgnocastExecutor.

Issue #658, opened 2025-07-11
Support Runtime Fallback to ROS 2 Behavior When ENABLE_AGNOCAST=0 Despite Agnocast-Enabled Build


sykwer:
Autoware Universe currently provides a package called [autoware_agnocast_wrapper](https://github.com/autowarefoundation/autoware_universe/tree/main/common/autoware_agnocast_wrapper), which serves as a wrapper around the Agnocast API. When Autoware uses Agnocast, it does not interact directly with the Agnocast API; instead, it uses macros provided by this wrapper package. The main reason for this indirection is to allow easy switching between enabling and disabling Agnocast via the `ENABLE_AGNOCAST` environment variable. The current build and runtime behavior controlled by this variable is as follows:

1. If `ENABLE_AGNOCAST!=1` at build-time and `ENABLE_AGNOCAST!=1` at runtime, the system runs using the default ROS 2 behavior.
2. If `ENABLE_AGNOCAST=1` at build-time and `ENABLE_AGNOCAST=1` at runtime, the system runs with Agnocast enabled.
3. If `ENABLE_AGNOCAST=1` at build-time but `ENABLE_AGNOCAST!=1` at runtime, the system crashes due to the absence of LD_PRELOAD.

Until now, behavior (3) was considered by design. However, this issue proposes a change to that design:
> Even if the system is built with `ENABLE_AGNOCAST=1`, it should run normally without error when `ENABLE_AGNOCAST=0` is set at runtime, behaving identically to a build where `ENABLE_AGNOCAST=0`.

In other words, we want to gracefully fall back to default ROS 2 behavior when LD_PRELOAD is not set, instead of crashing. It’s important to note that the `ENABLE_AGNOCAST` environment variable is only known to `autoware_agnocast_wrapper` and not to the Agnocast library itself.

Currently, we rely on conditional compilation based on `ENABLE_AGNOCAST`. In the proposed change, we would always compile with Agnocast enabled and defer the decision to use Agnocast or fall back to ROS 2 behavior until runtime.

Koichi98:
Closing this issue since the above PR has been merged.

PR #665, opened 2025-07-18
refactor: pass rvalue reference to the callback

bdm-k:
## Description

Before this PR, the callback functions received a `const` lvalue reference to `ipc_shared_ptr`. If a callback wanted to own an `ipc_shared_ptr`, it had no choice but to copy-construct one, which incurs a system call.

The aim of this PR is to allow the callbacks to move-construct an `ipc_shared_ptr`. It changes the argument passed to the callbacks from a `const` lvalue reference to an rvalue reference. This ability to move-construct is required in #658 for performance reasons.

Currently, `std::function<void(const ipc_shared_ptr<T> &)>` is the only option for the callbacks. So, this PR is simply a refactoring for now. In the subsequent PR, I'm going to relax this constraint so that any callable that is invocable with `ipc_shared_ptr<T> &&` can be used as a callback. Then, the changes in this PR will take effect.

## Related links

- The subsequent PR: #670 

## How was this PR tested?

- [x] Autoware (required)
- [x] sample application

## Notes for reviewers

The motivation for move construction comes from #658. You can see that move construction is actually used in autowarefoundation/autoware_universe#11060.


sykwer:
LGTM. Great work!

Koichi98 (approved):

PR #675, opened 2025-07-29
test(integration): Add integration test investigate malloc region differences after fork

Enigamict:
## Description
Add integration tests to verify that fork memory allocation works correctly in the agnocast heaphook system.

Mainly handled within heaphook
```rust
extern "C" fn post_fork_handler_in_child() {
    IS_FORKED_CHILD.store(true, Ordering::Relaxed);
}
```
```rust
fn should_use_original_func() -> bool {
    if IS_FORKED_CHILD.load(Ordering::Relaxed) {
        return true;
    }

    unsafe {
        if agnocast_get_borrowed_publisher_num() == 0 {
            return true;
        }
    }

    false
}
```
```rust
match (is_shared, is_forked_child) {
        (true, true) => (), // In the child processes, ignore the free operation to the shared memory
        (true, false) => tlsf_deallocate_wrapped(non_null_ptr),
        (false, _) => (*ORIGINAL_FREE.get_or_init(init_original_free))(ptr),
}
```

So The tests verify that these three functions work correctly:

1. IS_FORKED_CHILD flag is properly set in child processes after fork
2. Child processes always use original malloc regardless of publisher state
3. Child process free on shared memory are safely ignored

This ensures memory safety and proper process isolation in multi-process scenarios.

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

### DIFF
--- a/src/agnocastlib/test/integration/test_agnocast_heaphook.cpp
+++ b/src/agnocastlib/test/integration/test_agnocast_heaphook.cpp
@@ -97,3 +97,49 @@ TEST_F(HeaphookIntegrationTest, malloc_publisher_num_switching)
   std::free(ptr2);
   std::free(ptr3);
 }
+TEST_F(HeaphookIntegrationTest, fork_memory_behavior)
+{
+  agnocast::validate_ld_preload();
+
+  agnocast::increment_borrowed_publisher_num();
+
+  void * parent_ptr = std::malloc(100);
+  EXPECT_TRUE(is_address_in_shared_mapping(parent_ptr));
+
+  pid_t pid = fork();
+
+  if (pid == 0) {
+    // malloc in the child process
+    void * child_ptr = std::malloc(100);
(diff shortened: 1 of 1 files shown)


src/agnocastlib/test/integration/test_agnocast_heaphook.cpp
```diff
@@ -1,7 +1,121 @@
...
+    EXPECT_TRUE(is_address_in_shared_mapping(next_parent_ptr));
+
+    int status;
+    waitpid(pid, &status, 0);
+
+    EXPECT_EQ(WEXITSTATUS(status), 0);
+    EXPECT_TRUE(is_address_in_shared_mapping(next_parent_ptr));
+
+    // Confirms that it is still allocated in the parent process
+    EXPECT_TRUE(is_address_in_shared_mapping(parent_ptr));
```
veqcc, line 111: Does it really fail if you inject a bug such that a child process frees `parent_ptr` ?

veqcc (approved):

PR #676, opened 2025-07-31
Remove broken source.crates-io config

jerry73204:
## Description

The `agnocast_heaphook/.cargo/config.toml` configuration file instructs the Cargo to search for packages in the `vendor` directory. However, the directory is not available for public users. The patch removes this file to fix the issue.

## How was this PR tested?

- [x] Autoware (required)
- [x] sample application

### DIFF
--- a/agnocast_heaphook/.cargo/config.toml
+++ b/agnocast_heaphook/.cargo/config.toml
@@ -1,5 +1,0 @@
-[source.crates-io]
-replace-with = "vendored-sources"
-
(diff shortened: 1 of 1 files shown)


atsushi421:
@jerry73204 
Thank you for your contribution!

> However, the directory is not available for public users. The patch removes this file to fix the issue.

Is the `vendor` directory still unavailable even when running `scripts/build_all`?

PR #705, opened 2025-10-10
fix(heaphook) Fix `realloc` to call `malloc` when the pointer is `NULL`

r1ru:
## Description
In the current implementation of `realloc`, when the pointer is `NULL`, `is_shared` becomes `false`, and therefore glibc realloc is always used. However, when the pointer is `NULL`, the operation is equivalent to `malloc(size)`, Therefore, there are situations where allocation should be made from shared memory. Furthermore, since `is_shared` implies `ptr != NULL`, the following code contains unnecessary branches.

Therefore, I modified the implementation so that when the pointer is `NULL`, it calls `malloc`.

```rust
 match (is_shared, should_use_original) {
       ...
        (true, false) => {
            ...
            match NonNull::new(ptr.cast()) {
                ...
                // This code is never actually executed.
                None => {
                    // If `ptr` is NULL, then the call is equivalent to `malloc(size)`.
                    match AGNOCAST_SHARED_MEMORY_ALLOCATOR
                        .get()
                        .unwrap()
                        .inner
                        .allocate(new_layout)
                    {
                        Some(non_null_ptr) => non_null_ptr.as_ptr().cast(),
                        None => ptr::null_mut(),
                    }
                }
            }
        }
        (false, _) => (*ORIGINAL_REALLOC.get_or_init(init_original_realloc))(ptr, new_size),
}
```

Additionally, since the logic for deciding whether to use shared memory or the heap differs for each function (for example, `free` or `realloc` determine this based on the pointer value), the name `should_use_original_func` is misleading. I have renamed this to `should_use_heap`.

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

## Notes for reviewers
I’m not confident about how to handle the case when the size is 0. According to [C23](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3088.pdf), `malloc(0)` is implementation-defined:

> If the size of the space requested is zero, the behavior is implementation-defined.

Furthermore, it is unclear whether `rlsf::Tlsf::{allocate, reallocate}`  permits a size of zero — at least, there is no mention of this in the [documentation](https://docs.rs/rlsf/latest/rlsf/struct.Tlsf.html).

### DIFF
--- a/src/agnocastlib/CMakeLists.txt
+++ b/src/agnocastlib/CMakeLists.txt
@@ -27,7 +27,8 @@ add_library(agnocast SHARED
   src/agnocast.cpp src/agnocast_utils.cpp src/agnocast_publisher.cpp src/agnocast_subscription.cpp
   src/agnocast_smart_pointer.cpp src/agnocast_callback_info.cpp src/agnocast_executor.cpp
   src/agnocast_single_threaded_executor.cpp src/agnocast_multi_threaded_executor.cpp
-  src/agnocast_callback_isolated_executor.cpp src/agnocast_tracepoint_wrapper.c)
+  src/agnocast_callback_isolated_executor.cpp src/agnocast_tracepoint_wrapper.c
+  src/agnocast_client.cpp)
 # ament_target_dependencies(agnocast rclcpp)
 
 target_link_libraries(agnocast
--- a/src/agnocastlib/include/agnocast/agnocast.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast.hpp
@@ -2,8 +2,10 @@
 
 #include "agnocast/agnocast_callback_info.hpp"
 #include "agnocast/agnocast_callback_isolated_executor.hpp"
+#include "agnocast/agnocast_client.hpp"
 #include "agnocast/agnocast_multi_threaded_executor.hpp"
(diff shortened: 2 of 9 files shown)


agnocast_heaphook/src/lib.rs
```diff
@@ -462,65 +463,43 @@ pub extern "C" fn calloc(num: usize, size: usize) -> *mut c_void {
 ///
 #[no_mangle]
 pub unsafe extern "C" fn realloc(ptr: *mut c_void, new_size: usize) -> *mut c_void {
-    let is_shared = is_shared(ptr.cast());
-    let should_use_original = should_use_original_func();
+    // If `ptr` is NULL, then the call is equivalent to `malloc(size)`.
+    if ptr.is_null() {
+        return malloc(new_size);
+    }
```
veqcc: In my understanding, inserting null-check before `is_shared` check is the core part in this PR.
Why are other changes needed?

Koichi98: Likely because of this in the PR description.

> Furthermore, since is_shared implies ptr != NULL, the following code contains unnecessary branches.

Koichi98 (approved):
LGTM when the above issues are resolved.

veqcc (approved):

PR #726, opened 2025-11-14
test(heaphook): Add unit tests for the fundamental alignment

r1ru:
## Description
According to [C23](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3220.pdf), when allocation succeeds, memory management functions such as `aligned_alloc` and `malloc` must return a pointer that is suitably aligned for storing **any** obuject with a *fundamental alignment* requirement and size less than or equal to the size requested. The *fundamental alignment* is a nonnegative integral power of two less than or equal to `alignof(max_align_t)`.

This PR explicitly documents the above alignment constraints and adds some unit tests. Since the behavior of `posix_memalign` and `memalign` is not defined by the C standard, there are no corresponding unit tests for these two functions.

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

## Notes for reviewers
There is a subtle point with `calloc`. For example, for `calloc(16, 1)`, it is not clear from the C standard whether the size should be treated as 1 byte, so no alignment is needed, or as 16 bytes, requiring 16-byte alignment. The current implementation conservatively uses the latter.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,13 +1,13 @@
-* @veqcc @atsushi421 @Koichi98
+* @veqcc @atsushi421 @Koichi98 @sykwer
 
 # agnocast_heaphook
-agnocast_heaphook/ @Koichi98 @veqcc
+agnocast_heaphook/ @Koichi98 @veqcc @sykwer
 
 # agnocastlib executor and component container
-src/agnocastlib/src/*executor* @Koichi98 @atsushi421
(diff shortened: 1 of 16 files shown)


veqcc (approved):

veqcc:
Thank you!!

> There is a subtle point with calloc. For example, for calloc(16, 1), it is not clear from the C standard whether the size should be treated as 1 byte, so no alignment is needed, or as 16 bytes, requiring 16-byte alignment. The current implementation conservatively uses the latter.

I think we can accept it.

PR #729, opened 2025-11-15
fix(agnocastlib): consistent locking semantics

sykwer:
## Description
std::scoped_lock is available from C++17 and provides a mechanism for locking multiple mutexes at once without worrying about acquisition order.
Also, while std::lock_guard can omit its template argument thanks to class template argument deduction (CTAD) in C++17 and later, it requires explicit template parameters in earlier standards.

This PR unifies the style to the C++14-compatible form.
There was inconsistency in how these constructs were written within agnocastlib, and this change standardizes them.

## How was this PR tested?

- [ ] Autoware (required)
- [x] sample application


Koichi98:
@sykwer 
Could you try e2e tests just to make sure? I always want to make sure that e2e tests work even if it is a small fix.

PR #744, opened 2025-11-25
feat(kmod) [needs minor version update]: add filtered pub/sub count ioctls

kobayu858:
## Description
Introduced new ioctl commands: `AGNOCAST_GET_FILTERED_SUBSCRIBER_NUM_CMD` and `AGNOCAST_GET_FILTERED_PUBLISHER_NUM_CMD`.

These commands allow retrieving the number of subscribers or publishers while excluding a specific PID (`exclude_pid`).

**Motivation:**
In the bridge functionality, the bridge node itself maintains its own publishers and subscribers. To determine if a topic is still required by external processes, it need to calculate the number of participants excluding the bridge's own process.

Without this exclusion, the count includes the bridge itself, making it impossible to determine when to safely shut down the bridge. This feature will be utilized in the future for the bridge's shutdown determination logic.

## How was this PR tested?

- [ ] Autoware (required)
- [x] sample application

**Important notes:**

### DIFF
--- a/agnocast_kmod/agnocast_main.c
+++ b/agnocast_kmod/agnocast_main.c
@@ -1383,13 +1383,12 @@ static int get_topic_publisher_info(
 }
 
 int get_filtered_subscriber_num(
-  const char * topic_name, const struct ipc_namespace * ipc_ns,
+  const char * topic_name, const struct ipc_namespace * ipc_ns, pid_t exclude_pid,
   union ioctl_get_filtered_subscriber_num_args * ioctl_ret)
 {
   struct topic_wrapper * wrapper = find_topic(topic_name, ipc_ns);
   if (wrapper) {
-    pid_t target_pid = ioctl_ret->exclude_pid;
-    ioctl_ret->ret_ext_subscriber_num = get_size_sub_info_htable(wrapper, target_pid);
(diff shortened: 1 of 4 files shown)


agnocast_kmod/agnocast_main.c
```diff
@@ -1376,6 +1382,36 @@ static int get_topic_publisher_info(
   return 0;
 }
 
+int get_filtered_subscriber_num(
+  const char * topic_name, const struct ipc_namespace * ipc_ns,
+  union ioctl_get_filtered_subscriber_num_args * ioctl_ret)
+{
+  struct topic_wrapper * wrapper = find_topic(topic_name, ipc_ns);
+  if (wrapper) {
+    pid_t target_pid = ioctl_ret->exclude_pid;
```
Koichi98: I thought ioctl_ret was usually used for the kmod return value in agnocast. Regarding exclude_pid, could you create a dedicated variable for it and pass it directly instead of reusing ioctl_ret?

kobayu858: Thank you for the advice. I've changed it to take the pid as an argument.

kobayu858:
I'm closing this since it's no longer needed.
[goal](https://github.com/tier4/agnocast/compare/bridge_normal_goal)

PR #755, opened 2025-11-26
feat(agnocastlib): add safe_strncpy

kobayu858:
## Description
Introduced `safe_strncpy`, a wrapper around `std::strncpy` that guarantees null-termination.
`agnocast_bridge_utils.cpp/hpp` will implement common functions to be used in the bridge.

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

**Important notes:**


Koichi98:
@kobayu858 
Why don't we use snprintf or some other existing functions? We can't apply those here?

PR #764, opened 2025-11-29
feat(agnocastlib): add parse param rule

Koichi98:
## Description
Parse and store `-p`/`--param` command-line arguments in Context.
This PR only implements parsing and storage. Actual parameter retrieval API and tests for this functionality will be added in a follow-up PR.

## Limitations compared to rcl

  - **No yaml parser**: Only scalar values are supported. Arrays and complex YAML types are not parsed.
  - **No node name prefix**: Node-specific parameter syntax is not supported.

### Examples of differences

  | Input | rcl | agnocast |
  |-------|-----|----------|
  | `-p foo:=123` | int `123` | int `123` |
  | `-p foo:=true` | bool `true` | bool `true` |
  | `-p foo:=[1,2,3]` | int array `{1,2,3}` | string `"[1,2,3]"` |
  | `-p /my_node:foo:=123` | `foo=123` for `/my_node` only | param name becomes `/my_node:foo` |

## Related links
rcl_parse_arguments - parsing '-p/--param'.
https://github.com/ros2/rcl/blob/b7d6d69e670aa97bf69a6b92d12321ed31e68a4c/rcl/src/rcl/arguments.c#L341

## How was this PR tested?

- [ ] Autoware (required)
- [x] sample application

**Important notes:**

### DIFF
--- a/src/agnocastlib/src/agnocast_context.cpp
+++ b/src/agnocastlib/src/agnocast_context.cpp
@@ -108,8 +108,9 @@ Context::ParameterValue Context::parse_parameter_value(const std::string & value
 {
   // Convert to lowercase for case-insensitive comparison
   std::string lower_value = value_str;
-  std::transform(lower_value.begin(), lower_value.end(), lower_value.begin(),
-                 [](unsigned char c) { return std::tolower(c); });
+  std::transform(lower_value.begin(), lower_value.end(), lower_value.begin(), [](unsigned char c) {
+    return std::tolower(c);
+  });
 
   if (lower_value == "true") {
     return true;


src/agnocastlib/include/agnocast/agnocast_context.hpp
```diff
@@ -14,7 +16,13 @@ namespace agnocast
...
 #define AGNOCAST_ROS_ARGS_EXPLICIT_END_TOKEN "--"
 
-class Context
+/// The ROS flag that precedes the setting of a ROS parameter.
+#define AGNOCAST_PARAM_FLAG "--param"
+
+/// The short version of the ROS flag that precedes the setting of a ROS parameter.
+#define AGNOCAST_SHORT_PARAM_FLAG "-p"
+
+main class Context
```
sykwer: typo?

Koichi98: Fixed! Thank you

PR #799, opened 2025-12-09
feat(agnocastlib): add start_ros_execution

kobayu858:
## Description
Implemented `BridgeManager::start_ros_execution` to initialize the ROS 2 node and executor, and run them in a dedicated thread.

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

**Important notes:**

### DIFF
--- a/src/agnocast_sample_application/launch/no_rclcpp_listener.launch.xml
+++ b/src/agnocast_sample_application/launch/no_rclcpp_listener.launch.xml
@@ -1,5 +1,13 @@
 <launch>
-    <node pkg="agnocast_sample_application" exec="no_rclcpp_listener" name="no_rclcpp_listener_node" output="screen">
+    <arg name="use_remap" default="false" />
+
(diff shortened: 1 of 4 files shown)


src/agnocastlib/src/agnocast_bridge_manager.cpp
```diff
@@ -45,7 +52,27 @@ void BridgeManager::run()  // NOLINT(readability-convert-member-functions-to-sta
...
   // TODO(yutarokobayashi): event_loop_;
 }
 
+void BridgeManager::start_ros_execution()
+{
+  std::string node_name = "agnocast_bridge_node_" + std::to_string(getpid());
+  container_node_ = std::make_shared<rclcpp::Node>(node_name);
+
+  executor_ =
+    std::make_shared<agnocast::MultiThreadedAgnocastExecutor>(rclcpp::ExecutorOptions(), 0, 0);
```
Koichi98: Seems like these arguments are default value. Is there specific reason that you explicitly pass only these arguments here?

PR #802, opened 2025-12-10
feat(agnocastlib): implement skeleton of BridgeManager main loop and lifecycle checks

kobayu858:
## Description
This PR implements the main execution loop (`run()`) for the `BridgeManager`. It establishes the basic structure for process lifecycle management, including parent process monitoring and exit conditions.

## Changes
- Implemented `BridgeManager::run()` to handle the main loop.
- Added `check_parent_alive()` to monitor the parent process using `kill(pid, 0)`.
- Added `check_should_exit()` to handle graceful shutdown when the parent is gone and no bridges are active.
- Added `check_active_bridges()` as a placeholder for dynamic bridge management.

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

**Important notes:**

### DIFF
--- a/src/agnocastlib/src/agnocast_bridge_manager.cpp
+++ b/src/agnocastlib/src/agnocast_bridge_manager.cpp
@@ -118,7 +118,9 @@ void BridgeManager::check_should_exit()
 
 void BridgeManager::remove_active_bridges(const std::string & topic_name_with_dirction)
 {
-  if (!active_bridges_.count(topic_name_with_dirction)) return;
+  if (active_bridges_.count(topic_name_with_dirction) == 0) {
+    return;
+  }
 
   active_bridges_.erase(topic_name_with_dirction);
   // TODO(yutarokobayashi): Unregister from the kernel only if the paired bridge in the reverse


src/agnocastlib/src/agnocast_bridge_manager.cpp
```diff
@@ -74,4 +86,30 @@ void BridgeManager::start_ros_execution()
...
+}
+
+void BridgeManager::check_active_bridges()
+{
+  // TODO(yutarokobayashi): Verifying the number of connections and dynamic bridge removal
+}
+
+void BridgeManager::check_should_exit()
+{
+  if (!is_parent_alive_ && active_bridges_.empty()) {
```
Koichi98: When will the number of active_bridges_ decrease?

kobayu858: The BridgeManager holds function pointers that are utilized by the parent process.

Even if active_bridges_ drops to zero temporarily, the parent process may request to create new Pub/Sub connections in the future. If the BridgeManager terminates solely because there are no active participants, the parent process would lose the ability to establish new bridges. Therefore, the BridgeManager must remain active to handle potential future requests as long as the parent process is alive.

PR #805, opened 2025-12-10
feat(agnocastlib): implement safe epoll event loop with timeout

kobayu858:
## Description
This PR implements the `BridgeIpcEventLoop::spin_once` method to handle IPC events efficiently using epoll.
Specific event registration (e.g., message queues, signals) and the actual event processing logic inside the loop are not yet implemented and are marked as TODO. These will be added in a subsequent PR.

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

**Important notes:**

### DIFF
--- a/agnocast_kmod/Makefile
+++ b/agnocast_kmod/Makefile
@@ -17,6 +17,7 @@ else
 	  agnocast_kunit/agnocast_kunit_receive_msg.o \
 	  agnocast_kunit/agnocast_kunit_take_msg.o \
 	  agnocast_kunit/agnocast_kunit_get_subscriber_num.o \
+	  agnocast_kunit/agnocast_kunit_get_subscriber_qos.o \
 	  agnocast_kunit/agnocast_kunit_do_exit.o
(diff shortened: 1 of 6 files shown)


src/agnocastlib/src/agnocast_bridge_ipc_event_loop.cpp
```diff
@@ -1,15 +1,74 @@
...
 namespace agnocast
 {
 
 BridgeIpcEventLoop::BridgeIpcEventLoop(pid_t /*target_pid*/, const rclcpp::Logger & logger)
 : logger_(logger)
 {
+  try {
+    setup_epoll();
+  } catch (...) {
+    if (epoll_fd_ != -1) {
```
Koichi98: When will this be true?

Koichi98: Ah, I understand. It's after when you add epol_ctl function in setup_epoll().

Koichi98 (approved):

PR #809, opened 2025-12-10
feat(agnocastlib): add get_subscriber/publisher_qos

kobayu858:
## Description
This PR implements the C++ side logic to retrieve QoS settings dynamically from the agnocast kernel module via ioctl. It updates the bridge node factory functions to apply these QoS settings instead of using default hardcoded values.

If ioctl fails, communication guarantees are lost. Throwing an exception here ensures that the bridge node initialization fails fast.
 In the future, This exception will be caught by BridgeLoader::create_bridge_instance, logged as an error, and prevent the creation of the bridge resource.

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

**Important notes:**

### DIFF
--- a/src/agnocastlib/include/agnocast/agnocast_bridge_node.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_bridge_node.hpp
@@ -216,7 +216,7 @@ void send_bridge_request(
   (void)fn_reverse;  // TODO(yutarokobayashi): Remove
 
   // TODO(yutarokobayashi): Implement the actual message queue communication to request a bridge.
-  //
+  // std::string mq_name = create_mq_name_for_bridge_parent(getppid());
   // Note: This implementation depends on AgnocastPublisher and AgnocastSubscription.
 }
 
--- a/src/agnocastlib/include/agnocast/agnocast_bridge_ipc_event_loop.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_bridge_ipc_event_loop.hpp
@@ -2,7 +2,10 @@
 
 #include <rclcpp/logger.hpp>
 
+#include <mqueue.h>
 #include <sys/types.h>
+
+#include <string>
 
 namespace agnocast
 {
@@ -23,7 +26,17 @@ private:
 
   int epoll_fd_ = -1;
 
+  mqd_t mq_parent_fd_ = (mqd_t)-1;
+
+  std::string mq_parent_name_;
+
+  void setup_mq(pid_t target_pid);
   void setup_epoll();
+
+  mqd_t create_and_open_mq(const std::string & name, const std::string & label);
+  void add_fd_to_epoll(int fd, const std::string & label) const;
+
+  void cleanup_resources();
 };
 
 }  // namespace agnocast
--- a/src/agnocastlib/src/agnocast_bridge_ipc_event_loop.cpp
+++ b/src/agnocastlib/src/agnocast_bridge_ipc_event_loop.cpp
@@ -1,8 +1,13 @@
 #include "agnocast/agnocast_bridge_ipc_event_loop.hpp"
 
+#include "agnocast/agnocast_mq.hpp"
+#include "agnocast/agnocast_utils.hpp"
+
 #include <rclcpp/logging.hpp>
 
+#include <fcntl.h>
 #include <sys/epoll.h>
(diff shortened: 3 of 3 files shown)


src/agnocastlib/include/agnocast/agnocast_bridge_node.hpp
```diff
@@ -140,20 +140,59 @@ class AgnocastToRosBridge
...
 
+inline rclcpp::QoS get_subscriber_qos(
+  const std::string & topic_name, topic_local_id_t subscriber_id)
+{
+  struct ioctl_get_subscriber_qos_args get_subscriber_qos_args = {};
+  get_subscriber_qos_args.topic_name = {topic_name.c_str(), topic_name.size()};
+  get_subscriber_qos_args.subscriber_id = subscriber_id;
+
+  if (ioctl(agnocast_fd, AGNOCAST_GET_SUBSCRIBER_QOS_CMD, &get_subscriber_qos_args) < 0) {
+    //  In the future, This exception will be caught by BridgeLoader::create_bridge_instance
```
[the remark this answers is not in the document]
kobayu858: I clarified the sentence.

src/agnocastlib/include/agnocast/agnocast_bridge_node.hpp
```diff
@@ -140,20 +140,59 @@ class AgnocastToRosBridge
...
+                                                    : rclcpp::DurabilityPolicy::Volatile);
+}
+
 template <typename MessageT>
 std::shared_ptr<void> start_ros_to_agno_node(
   rclcpp::Node::SharedPtr node, const BridgeTargetInfo & info)
 {
-  // TODO(yutarokobayashi): Implementation of get_subscriber_qos
-  return std::make_shared<RosToAgnocastBridge<MessageT>>(node, info.topic_name, rclcpp::QoS(10));
+  std::string topic_name(static_cast<const char *>(info.topic_name));
```
Koichi98: Can be easily
```
std::string topic_name(info.topic_name); 
```
?

src/agnocastlib/include/agnocast/agnocast_bridge_node.hpp
```diff
@@ -140,20 +140,59 @@ class AgnocastToRosBridge
...
+
+inline rclcpp::QoS get_publisher_qos(const std::string & topic_name, topic_local_id_t publisher_id)
+{
+  struct ioctl_get_publisher_qos_args get_publisher_qos_args = {};
+  get_publisher_qos_args.topic_name = {topic_name.c_str(), topic_name.size()};
+  get_publisher_qos_args.publisher_id = publisher_id;
+
+  if (ioctl(agnocast_fd, AGNOCAST_GET_PUBLISHER_QOS_CMD, &get_publisher_qos_args) < 0) {
+    // This exception is intended to be caught by the factory function that instantiates the bridge.
+    throw std::runtime_error("Failed to fetch publisher QoS from agnocast kernel module");
```
Koichi98: Might be better to include topic_name and publisher_id for this error message.

kobayu858: Thank you. I have corrected the error message.

src/agnocastlib/include/agnocast/agnocast_bridge_node.hpp
kobayu858: The reason is that it caused an error with the local clang-tidy.

```
clang-tidy src/agnocastlib/include/agnocast/agnocast_bridge_node.hpperror: do not implicitly decay an array into a pointer; consider using gsl::array_view or an explicit cast instead [cppcoreguidelines-pro-bounds-array-to-pointer-decay,hicpp-no-array-decay,-warnings-as-errors]
  std::string topic_name(info.topic_name);
                         ^
```

However, since similar patterns in the existing code don't seem to trigger this, I applied this fix tentatively. If it fails in CI, I will revise it.

Koichi98: Might be because clang-tidy check has been introduced after those codes are implemented... not sure. 

> If it fails in CI, I will revise it.

Ok.

kobayu858: It seems to be flagged by CI after all, so I'll revert it back.

PR #810, opened 2025-12-11
feat(agnocastlib): support handling parent IPC requests in BridgeManager

kobayu858:
## Description
This PR implements the handling of IPC messages received from the parent process within the BridgeManager.

Changes:

- Added logic to listen for events on mq_parent_fd_.
- Implemented set_parent_mq_handler to register a callback for parent process events.
- Added on_mq_event to process incoming MqMsgBridge requests from the message queue.
- Registered the handler to the event loop. The allow_delegation flag is set to true here, as requests originating from the parent process are allowed to be delegated to other child processes if necessary.

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

**Important notes:**

### DIFF
--- a/src/agnocastlib/src/agnocast_bridge_manager.cpp
+++ b/src/agnocastlib/src/agnocast_bridge_manager.cpp
@@ -104,6 +104,9 @@ void BridgeManager::handle_create_request(const MqMsgBridge & req, bool /*allow_
   // - true: Initial request from the parent process. Delegation to an existing owner is allowed.
   // - false: Request received from a peer child process (already delegated). Further delegation
   //          is disabled to prevent infinite loops.
+  // Note: This check (if false) acts as a safety guard against unexpected infinite recursion loops
+  // due to race conditions or timing issues. It is not expected to be triggered in normal
+  // operation.
 
   // Locally, unique keys include the direction. However, we register the raw topic name (without
   // direction) to the kernel to enforce single-process ownership for the entire topic.


src/agnocastlib/src/agnocast_bridge_manager.cpp
```diff
@@ -90,6 +90,44 @@ void BridgeManager::start_ros_execution()
...
+    handle_create_request(req, allow_delegation);
+  }
+}
+
+void BridgeManager::handle_create_request(const MqMsgBridge & req, bool /*allow_delegation*/)
+{
+  // The 'allow_delegation' flag indicates the source of the request:
+  // - true: Initial request from the parent process. Delegation to an existing owner is allowed.
+  // - false: Request received from a peer child process (already delegated). Further delegation
+  //          is disabled to prevent infinite loops.
```
Koichi98: When will delegation fail two times in a row? This allow_delegation:false is used when the request is already delegated, right?

Koichi98: Have you considered how the process that delegates the bridge creation actually verifies whether the bridge has been successfully created? I believe the processing of handle_create_request should be considered complete only after this verification succeed.

Koichi98: > Logically, delegation failing two times in a row should not happen. However, I included this check as a safety guard to ensure we never fall into an infinite recursion loop, even if the kernel state or IPC timing behaves unexpectedly. Given that the overhead of this check is minimal, I preferred safety here.

Could you add in a comment that this is just a safety guard and not intended to in an expected behavior?

src/agnocastlib/include/agnocast/agnocast_bridge_ipc_event_loop.hpp
```diff
@@ -5,14 +5,16 @@
 #include <mqueue.h>
 #include <sys/types.h>
 
+#include <functional>
 #include <string>
 
 namespace agnocast
 {
```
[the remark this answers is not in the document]
kobayu858: False positive.

src/agnocastlib/src/agnocast_bridge_manager.cpp
kobayu858: > Could you add in a comment that this is just a safety guard and not intended to in an expected behavior?

Sure, I've added a comment clarifying that this check is a safety guard and not part of the standard execution flow.

Koichi98 (approved):

PR #843, opened 2025-12-16
feat(agnocast): add SubscriptionBase destructor

kobayu858:
## Description
Previously, subscriber cleanup depended solely on process termination. However, the introduction of the Bridge feature requires subscribers to be dynamically created and removed during the process's lifetime, making this explicit destructor necessary.

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

**Important notes:**

### DIFF
--- a/src/agnocastlib/src/agnocast_bridge_manager.cpp
+++ b/src/agnocastlib/src/agnocast_bridge_manager.cpp
@@ -98,6 +98,9 @@ void BridgeManager::on_mq_create_request(mqd_t fd)
 {
   MqMsgBridge req{};
   while (mq_receive(fd, reinterpret_cast<char *>(&req), sizeof(req), nullptr) > 0) {
+    if (shutdown_requested_) {
+      break;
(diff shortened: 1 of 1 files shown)


src/agnocastlib/include/agnocast/agnocast_subscription.hpp
```diff
@@ -81,6 +81,18 @@ class SubscriptionBase
 public:
   SubscriptionBase(rclcpp::Node * node, const std::string & topic_name);
   SubscriptionBase(agnocast::Node * node, const std::string & topic_name);
+
+  virtual ~SubscriptionBase()
```
Koichi98:
```
NOTE: Unmapping memory when a subscriber is destroyed is not implemented. Multiple subscribers may share the same mmap region, requiring reference counting in kmod. Since leaving the memory mapped should not cause any functional issues, this is left as future work.
```

Could you add the above sentences as a comment?

PR #844, opened 2025-12-16
feat(kmod)[needs minor version update]: remove publisher cmd

kobayu858:
## Description
This PR adds support for explicitly removing a publisher through the AGNOCAST_REMOVE_PUBLISHER_CMD ioctl command. This allows user-space applications to properly stop publishing and ensures that kernel resources are cleaned up immediately when a publisher is destroyed.

Previously, publisher cleanup depended solely on process termination. However, the introduction of the Bridge feature requires publishers to be dynamically created and removed during the process's lifetime, making this explicit removal command necessary.

Changes:

- Added struct ioctl_remove_publisher_args for publisher removal parameters.
- Added AGNOCAST_REMOVE_PUBLISHER_CMD ioctl command definition.
- Implemented remove_publisher() function:
- Removes the publisher from the hash table.
- Iterates through message entries to clear ownership references held by this publisher.

Test Coverage:

- Basic removal: Verify publisher is removed and topic is deleted if empty.
- Persistence: Verify topic remains if a subscriber still exists after publisher removal.
- Orphaned Messages: Verify that messages currently referenced by subscribers remain available even after the publisher is removed, while unreferenced messages are immediately cleaned up.

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

**Important notes:**

### DIFF
--- a/src/agnocastlib/include/agnocast/agnocast_subscription.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_subscription.hpp
@@ -81,6 +81,22 @@ protected:
 public:
   SubscriptionBase(rclcpp::Node * node, const std::string & topic_name);
   SubscriptionBase(agnocast::Node * node, const std::string & topic_name);
+
+  virtual ~SubscriptionBase()
+  {
+    // NOTE: Unmapping memory when a subscriber is destroyed is not implemented. Multiple
+    // subscribers
+    // may share the same mmap region, requiring reference counting in kmod. Since leaving the
+    // memory mapped should not cause any functional issues, this is left as future work.
+    struct ioctl_remove_subscriber_args remove_subscriber_args
+    {
+    };
+    remove_subscriber_args.topic_name = {topic_name_.c_str(), topic_name_.size()};
+    remove_subscriber_args.subscriber_id = id_;
+    if (ioctl(agnocast_fd, AGNOCAST_REMOVE_SUBSCRIBER_CMD, &remove_subscriber_args) < 0) {
+      RCLCPP_WARN(logger, "Failed to remove subscriber (id=%d) from kernel.", id_);
+    }
+  }
 };
 
 template <typename MessageT, typename BridgeRequestPolicy>


agnocast_kmod/agnocast_main.c
```diff
@@ -1543,6 +1543,67 @@ int remove_subscriber(
...
+      }
+    }
+
+    if (!is_referenced(en)) {
+      pub_info->entries_num--;
+      remove_entry_node(wrapper, en);
+    }
+  }
+
+  hash_del(&pub_info->node);
```
Koichi98: We shouldn't delete the pub_info until all the entry_node are freed. The responsibility of freeing the pub_info would be the last referencing subscriber in such case.

https://github.com/tier4/agnocast/blob/a6dd68fe5506b166ec8074b056cd8e31f2296e44/agnocast_kmod/agnocast_main.c#L2359

Also, could we extract the common logic between pre_handler_publisher_exit and this function into a separate utility? It looks like a duplication of code with the exact same purpose and meaning, so I think it’s worth consolidating. I probably should have pointed this out for the subscriber side as well.

agnocast_kmod/agnocast_main.c
```diff
@@ -1543,6 +1543,73 @@ int remove_subscriber(
...
+    }
+  }
+
+  if (pub_info->entries_num == 0) {
+    hash_del(&pub_info->node);
+    kfree(pub_info->node_name);
+    kfree(pub_info);
+  } else {
+    dev_info(
+      agnocast_device, "Publisher (id=%d) removed but pub_info kept for %d orphaned entries.\n",
```
Koichi98: Do we need this?

agnocast_kmod/agnocast_main.c
```diff
@@ -1543,6 +1543,73 @@ int remove_subscriber(
...
+      remove_entry_node(wrapper, en);
+    }
+
+    hash_del(&wrapper->node);
+    kfree(wrapper->key);
+    kfree(wrapper);
+    dev_info(agnocast_device, "Topic %s removed (empty).\n", topic_name);
+  } else {
+    dev_info(
+      agnocast_device, "Publisher (id=%d) removed from topic %s.\n", publisher_id, topic_name);
```
Koichi98: Why do we print this only in this case: when we don't remove the topic?
(Same for subscriber)

kobayu858: I had made it possible to tell which branch was entered, but since outputting in both cases would make it clear, I removed the else statement.

PR #850, opened 2025-12-17
feat(agnocastlib): add remove cmd in pub destructor

kobayu858:
## Description
Previously, publisher cleanup depended solely on process termination. However, the introduction of the Bridge feature requires publishers to be dynamically created and removed during the process's lifetime, making this explicit deletion operation
 necessary.

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

**Important notes:**

### DIFF
--- a/src/agnocastlib/include/agnocast/agnocast_bridge_ipc_event_loop.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_bridge_ipc_event_loop.hpp
@@ -53,7 +53,7 @@ private:
   void setup_signals();
   void setup_epoll();
 
-  mqd_t create_and_open_mq(const std::string & name, const std::string & label);
+  static mqd_t create_and_open_mq(const std::string & name, const std::string & label);
   void add_fd_to_epoll(int fd, const std::string & label) const;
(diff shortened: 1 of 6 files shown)


src/agnocastlib/include/agnocast/agnocast_publisher.hpp
```diff
@@ -135,6 +135,19 @@ class BasicPublisher
         RCLCPP_ERROR(logger, "mq_close failed: %s", strerror(errno));
       }
     }
+
+    // NOTE: Unmapping memory when a publisher is destroyed is not implemented. Multiple
+    // publishers
+    // may share the same mmap region, requiring reference counting in kmod. Since leaving the
+    // memory mapped should not cause any functional issues, this is left as future work.
```
Koichi98: Please replace this sentences with the following.

```
When a publisher is destroyed, subscribers should unmap its memory, but this is not yet implemented. Since multiple publishers in the same process share a mempool, process-level reference counting in kmod is needed. Leaving memory mapped causes no functional issues, so this is left as future work.
```

PR #893, opened 2026-01-06
refactor(agnocastlib): unify mq naming and refactor unlink mq

kobayu858:
## Description
Decoupling mq_unlink from BridgeIpcEventLoop:

- Removed mq_unlink calls from the BridgeIpcEventLoop destructor/cleanup logic. The class now only handles mq_close.
- Moved the responsibility of unlinking MQs to the poll_for_unlink daemon. This ensures that resources are cleaned up asynchronously based on process exit events from the kernel module.

Unifying MQ Naming Functions:

- Merged two separate create_mq_name implementations into a single, unified helper function. This eliminates code duplication and ensures consistent naming conventions across the codebase.

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

**Important notes:**

### DIFF
--- a/src/agnocastlib/include/agnocast/bridge/agnocast_bridge_ipc_event_loop.hpp
+++ b/src/agnocastlib/include/agnocast/bridge/agnocast_bridge_ipc_event_loop.hpp
@@ -16,7 +16,7 @@ public:
   using EventCallback = std::function<void(int fd)>;
   using SignalCallback = std::function<void()>;
 
-  BridgeIpcEventLoop(const rclcpp::Logger & logger);
+  explicit BridgeIpcEventLoop(const rclcpp::Logger & logger);
   ~BridgeIpcEventLoop();
 
   BridgeIpcEventLoop(const BridgeIpcEventLoop &) = delete;


src/agnocastlib/src/bridge/agnocast_bridge_ipc_event_loop.cpp
```diff
@@ -150,9 +142,9 @@ sigset_t BridgeIpcEventLoop::block_signals(std::initializer_list<int> signals)
 
 void BridgeIpcEventLoop::setup_mq(pid_t target_pid)
 {
-  mq_parent_name_ = create_mq_name_for_bridge_parent(target_pid);
+  mq_parent_name_ = create_mq_name_for_bridge(target_pid);
   mq_parent_fd_ = create_and_open_mq(mq_parent_name_, "Parent");
-  mq_self_name_ = create_mq_name_for_bridge_daemon(getpid());
+  mq_self_name_ = create_mq_name_for_bridge(getpid());
```
Koichi98: Is there a reason we need two separate mqueues here? Both use identical handlers, and a single mqueue seems sufficient for both parent and peer messages.

Koichi98: Thanks for the explanation. However:
  1. Why should Parent messages have higher priority? Both are bridge creation requests processed identically by register_request. Also, the current code doesn't implement any priority handling - both MQs are processed equally in the event loop.
  2. Having two MQs uses more resources than one. (But this is negligible anyway)
  3. Is that level of debugging granularity really necessary? Although both MQs use identical handlers..

Koichi98: I know that, but that's a design choice, not a fundamental constraint. 
Since the Daemon is forked from Parent, fork() returns the Daemon's PID to the Parent. So Parent can send to a single MQ named by Daemon's PID, just like Peers do. Alternatively, we could register with kmod using the Parent's PID instead. Either way, there are options to unify them.

src/agnocastlib/src/agnocast.cpp
```diff
@@ -171,6 +171,9 @@ void poll_for_unlink()
       if (get_exit_process_args.ret_pid > 0) {
         const std::string shm_name = create_shm_name(get_exit_process_args.ret_pid);
         shm_unlink(shm_name.c_str());
+
+        const std::string mq_name = create_mq_name_for_bridge(get_exit_process_args.ret_pid);
```
Koichi98: Does the BridgeManager pid also get returned with `get_exit_process_args.ret_pid`?

Koichi98: Could you add a comment to indicate that we actually don't need to call unlink for non-BridgeManager processes but we actually do to avoid complexity?

kobayu858: Added a comment indicating the above.

src/agnocastlib/src/bridge/agnocast_bridge_ipc_event_loop.cpp
kobayu858: We discussed how to handle the PID for parent-to-child access without modifying kmod and considered the following two approaches:

Set the parent process PID in bridge_info

- Cons: This field is originally intended to hold the PID of the process managing the bridge. Using the parent process PID here would create a semantic mismatch and could lead to confusion regarding the message destination.

Store the bridged child process PID in a global variable within agnocastlib

- Pros: This maintains semantic consistency in the code.
- Implementation: No mutex is required as the variable is written once during initialization and read-only thereafter, keeping the complexity low.

Conclusion: To prioritize semantic consistency, we decided to adopt storing in a global variable.

[TIER IV Internal Slack](https://star4.slack.com/archives/C07FL8616EM/p1767685980552689)

Koichi98 (dismissed):

PR #908, opened 2026-01-09
docs(agnocastlib): add clarification comments to Executor API

atsushi421:
## Description

Add clarification comments to the `AgnocastOnlyExecutor` class to document that the `add_node()` method is implemented solely to unify the API with `rclcpp::Executor`. 

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application


src/agnocastlib/src/node/agnocast_only_executor.cpp
```diff
@@ -71,6 +71,7 @@ void AgnocastOnlyExecutor::execute_agnocast_executable(AgnocastExecutable & agno
   }
 }
 
+// Implemented only to unify the API with rclcpp::Executor
```
Koichi98:
```suggestion
// Implemented only to align the API with rclcpp::Executor
```

Koichi98 (approved):

PR #914, opened 2026-01-10
feat(agnocastlib): implement cancel() for CallbackIsolatedAgnocastExecutor

atsushi421:
## Description

Add cancel() public method to CallbackIsolatedAgnocastExecutor that enables shutdown of the executor and all child executors managing isolated callback groups.

## Related links

https://github.com/ros2/rclcpp/blob/41182a972055bdcfd9c19888c2f302ef8882b0ca/rclcpp/src/rclcpp/executor.cpp#L487

## How was this PR tested?

- [x] Unit test
- [x] sample application

### DIFF
--- a/src/agnocastlib/src/agnocast_callback_isolated_executor.cpp
+++ b/src/agnocastlib/src/agnocast_callback_isolated_executor.cpp
@@ -311,6 +311,7 @@ void CallbackIsolatedAgnocastExecutor::remove_node(rclcpp::Node::SharedPtr node_
 
 void CallbackIsolatedAgnocastExecutor::cancel()
 {
+  spinning.store(false);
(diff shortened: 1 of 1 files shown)


src/agnocastlib/src/agnocast_callback_isolated_executor.cpp
```diff
@@ -305,4 +309,15 @@ void CallbackIsolatedAgnocastExecutor::remove_node(rclcpp::Node::SharedPtr node_
   remove_node(node_ptr->get_node_base_interface(), notify);
 }
 
+void CallbackIsolatedAgnocastExecutor::cancel()
+{
```
Koichi98: No need to add `spinning.store(false)`?

PR #1001, opened 2026-02-05
test(agnocastlib): add unit tests for node base

Koichi98:
## Description
Reorganize the NodeBase unit tests into specification tests and fill in gaps in API coverage. All tests follow Arrange / Act / Assert and are grouped by the contract they document.

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

**Important notes:**

### DIFF
--- a/src/agnocastlib/test/unit/test_node_base.cpp
+++ b/src/agnocastlib/test/unit/test_node_base.cpp
@@ -675,6 +675,36 @@ TEST_F(TestNodeBase, resolve_topic_name_throws_for_unknown_substitution)
     node_base->resolve_topic_or_service_name("{unknown}", false, true), std::runtime_error);
 }
 
+TEST_F(TestNodeBase, resolve_topic_name_applies_remap_rule_when_only_expand_false)
+{
+  // Arrange
+  auto options = node_options_without_parameter_services();
+  options.arguments({"--ros-args", "-r", "/foo:=/bar"});
+  node_ = std::make_shared<agnocast::Node>("my_node", "/my_ns", options);
+  auto node_base = node_->get_node_base_interface();
+
+  // Act
+  auto resolved = node_base->resolve_topic_or_service_name("/foo", false, false);
+
+  // Assert
+  EXPECT_EQ("/bar", resolved);
+}
+
+TEST_F(TestNodeBase, resolve_service_name_applies_remap_rule_when_only_expand_false)
+{
+  // Arrange
+  auto options = node_options_without_parameter_services();
+  options.arguments({"--ros-args", "-r", "/foo_srv:=/bar_srv"});
+  node_ = std::make_shared<agnocast::Node>("my_node", "/my_ns", options);
+  auto node_base = node_->get_node_base_interface();
+
+  // Act
+  auto resolved = node_base->resolve_topic_or_service_name("/foo_srv", true, false);
+
+  // Assert
+  EXPECT_EQ("/bar_srv", resolved);
+}
+
 // =============================================================================
 // Category 9: on_callback_group_created hook
 //
--- a/src/agnocastlib/test/unit/test_node_base.cpp
+++ b/src/agnocastlib/test/unit/test_node_base.cpp
@@ -803,7 +803,10 @@ TEST_F(TestNodeBase, on_callback_group_created_fires_after_group_is_registered)
 //   - get_local_args() returns the rcl_arguments_t parsed from
 //     NodeOptions::arguments() (non-null even with no arguments).
 //   - get_global_args() returns nullptr when NodeOptions::use_global_arguments
-//     is false; otherwise it reflects the global agnocast context.
+//     is false, or when it is true but agnocast::init() has not been called.
+//     Only when use_global_arguments is true AND agnocast::init() has been
+//     called does it return a non-null pointer reflecting the global agnocast
+//     context.
 // =============================================================================
 
 TEST_F(TestNodeBase, get_local_args_is_non_null_for_default_options)
@@ -833,4 +836,24 @@ TEST_F(TestNodeBase, get_global_args_is_null_when_use_global_arguments_is_false)
 
   // Assert
   EXPECT_EQ(nullptr, global_args);
+}
+
+TEST_F(TestNodeBase, get_global_args_is_non_null_when_g_context_is_initialized)
+{
+  // Arrange: agnocast::init() populates g_context with parsed arguments,
+  // which the constructor reads when use_global_arguments=true (default).
+  agnocast::init(0, nullptr);
+  {
+    auto options = node_options_without_parameter_services();
+    auto node = std::make_shared<agnocast::Node>("my_node", "/my_ns", options);
+    auto node_base = std::static_pointer_cast<agnocast::node_interfaces::NodeBase>(
+      node->get_node_base_interface());
+
+    // Act
+    const rcl_arguments_t * global_args = node_base->get_global_args();
+
+    // Assert
+    EXPECT_NE(nullptr, global_args);
+  }
+  agnocast::shutdown();
 }
--- a/src/agnocastlib/test/unit/test_node_base.cpp
+++ b/src/agnocastlib/test/unit/test_node_base.cpp
@@ -144,6 +144,33 @@ TEST_F(TestNodeBase, get_fully_qualified_name_supports_nested_namespaces)
   EXPECT_STREQ("/ns1/ns2/ns3/my_node", fqn);
 }
 
+TEST_F(TestNodeBase, constructor_applies_node_and_ns_remap_rules)
+{
+  // Arrange
+  auto options = node_options_without_parameter_services();
+  options.arguments({"--ros-args", "-r", "__node:=remapped_node", "-r", "__ns:=/remapped_ns"});
+  node_ = std::make_shared<agnocast::Node>("my_node", "/my_ns", options);
+  auto node_base = node_->get_node_base_interface();
+
+  // Act / Assert
+  EXPECT_STREQ("remapped_node", node_base->get_name());
+  EXPECT_STREQ("/remapped_ns", node_base->get_namespace());
+  EXPECT_STREQ("/remapped_ns/remapped_node", node_base->get_fully_qualified_name());
+}
+
+TEST_F(TestNodeBase, constructor_throws_when_no_args_available_for_remap)
+{
+  // Arrange: with both local and global args NULL, rcl_remap_name returns
+  // RCL_RET_INVALID_ARGUMENT.
+  auto context = rclcpp::contexts::get_global_default_context();
+
+  // Act / Assert
+  EXPECT_THROW(
+    agnocast::node_interfaces::NodeBase(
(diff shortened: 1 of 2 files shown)


src/agnocastlib/test/unit/test_node_base.cpp
```diff
@@ -0,0 +1,774 @@
...
+// =============================================================================
+
+TEST_F(TestNodeBase, get_notify_guard_condition_throws_runtime_error)
+{
+  // Arrange
+  auto node_base = create_node_base("my_node", "/my_ns");
+
+  // Act / Assert
+  EXPECT_THROW(node_base->get_notify_guard_condition(), std::runtime_error);
+}
```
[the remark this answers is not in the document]
Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1001/commits/c6353ae26bf96983558cbe634c62e0e05ad8dc0b

src/agnocastlib/test/unit/test_node_base.cpp
```diff
@@ -0,0 +1,774 @@
...
+
+// =============================================================================
+// Category 6: Notify guard condition (unsupported by design)
+//
+// Specification:
+//   - agnocast::Node uses an epoll-based dispatcher rather than rclcpp's
+//     guard-condition notification, so guard-condition accessors throw
+//     std::runtime_error. On rclcpp 28+ (Jazzy), the additional
+//     get_shared_notify_guard_condition() and trigger_notify_guard_condition()
+//     overloads share the same contract.
```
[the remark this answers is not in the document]
Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1001/commits/c6353ae26bf96983558cbe634c62e0e05ad8dc0b.

atsushi421:
Audited this PR against `rclcpp::node_interfaces::NodeBaseInterface` for both `humble` and `jazzy`:

- ✅ Every public method of `NodeBaseInterface` is covered (18/18 on humble, 20/20 on jazzy — the jazzy-only `get_shared_notify_guard_condition` / `trigger_notify_guard_condition` are correctly guarded by `#if RCLCPP_VERSION_MAJOR >= 28`).
- ⚠️ One branch-level gap: all nine `resolve_topic_or_service_name` tests pass `only_expand = true`, so the remap path (the production default `only_expand = false`, which calls `rcl_remap_topic_name` / `rcl_remap_service_name`) is never exercised. Could you add tests for that branch — e.g. construct a node with `--ros-args -r /foo:=/bar` and assert remapping is applied for both `is_service=true` and `is_service=false`?

Koichi98:
@atsushi421 
> Could you add tests for that branch — e.g. construct a node with --ros-args -r /foo:=/bar and assert remapping is applied for both is_service=true and is_service=false?

Thanks! Added the tests as such.

src/agnocastlib/test/unit/test_node_base.cpp
```diff
@@ -0,0 +1,816 @@
...
+
+  // Act / Assert
+  EXPECT_THROW(node_base->get_notify_guard_condition(), std::runtime_error);
+}
+
+TEST_F(TestNodeBase, get_notify_guard_condition_returns_guard_when_context_is_valid)
+{
+  // Arrange: rclcpp::init() makes the default context valid. Scope the node
+  // so it is destroyed before rclcpp::shutdown() to keep cleanup well-defined.
+  rclcpp::init(0, nullptr);
```
atsushi421: Calling rclcpp::init/shutdown on the default rclcpp context creates fragile coupling with other tests in the same binary that may also touch the default context (e.g., via agnocast::init). If the default context is already initialized when this test runs, rclcpp::init will throw ContextAlreadyInitialized; once shutdown runs, the default context cannot be cleanly re-initialized later in the process. Use a local rclcpp::Context (via NodeOptions().context(ctx)) and shut down only that local context to isolate from global state.

src/agnocastlib/test/unit/test_node_base.cpp
```diff
@@ -0,0 +1,816 @@
...
+#include "agnocast/node/node_interfaces/node_base.hpp"
+#include "rclcpp/rclcpp.hpp"
+#include "rclcpp/version.h"
+
+#include <gtest/gtest.h>
+
+class TestNodeBase : public ::testing::Test
+{
+protected:
+  void SetUp() override {}
```
atsushi421: Empty SetUp() and TearDown() overrides duplicate the base class defaults and add boilerplate without functionality. Remove both.

src/agnocastlib/test/unit/test_node_base.cpp
```diff
@@ -0,0 +1,816 @@
...
+    return node_->get_node_base_interface();
+  }
+
+  // Down-cast to the concrete NodeBase to access methods that are not part of
+  // rclcpp::NodeBaseInterface (e.g. set_on_callback_group_created, get_local_args).
+  agnocast::node_interfaces::NodeBase::SharedPtr create_concrete_node_base(
+    const std::string & node_name, const std::string & node_namespace)
+  {
+    auto base = create_node_base(node_name, node_namespace);
+    return std::dynamic_pointer_cast<agnocast::node_interfaces::NodeBase>(base);
```
atsushi421: create_concrete_node_base just constructed an agnocast::Node, so the downcast can never fail. Use std::static_pointer_cast<agnocast::node_interfaces::NodeBase>(base) to express the invariant directly and drop the redundant ASSERT_NE(nullptr, ...) checks at each call site.

src/agnocastlib/test/unit/test_node_base.cpp
```diff
@@ -0,0 +1,816 @@
...
+  auto node_base = create_node_base("my_node", "/my_ns");
+
+  // Act
+  bool value = node_base->get_use_intra_process_default();
+
+  // Assert
+  EXPECT_FALSE(value);
+}
+
+TEST_F(TestNodeBase, get_use_intra_process_default_reflects_node_options_true)
```
atsushi421, line 517: The four ..._reflects_node_options_true/false tests in Category 7 differ only in a boolean flag. Convert them to TEST_P with INSTANTIATE_TEST_SUITE_P(::testing::Bool()) to halve the source size and remove copy-paste drift risk.

src/agnocastlib/test/unit/test_node_base.cpp
```diff
@@ -0,0 +1,816 @@
...
+  auto options = node_options_without_parameter_services();
+  options.use_global_arguments(false);
+  auto node = std::make_shared<agnocast::Node>("my_node", "/my_ns", options);
+  auto node_base =
+    std::dynamic_pointer_cast<agnocast::node_interfaces::NodeBase>(node->get_node_base_interface());
+  ASSERT_NE(nullptr, node_base);
+
+  // Act
+  const rcl_arguments_t * global_args = node_base->get_global_args();
+
```
atsushi421, line 836: Only the use_global_arguments=false branch is tested. The complementary case (use_global_arguments=true with an initialized agnocast g_context returning the parsed global arguments) — i.e., the production path — is uncovered. Add a test that initializes the global context, builds a node with use_global_arguments(true), and asserts get_global_args() returns non-null.

src/agnocastlib/test/unit/test_node_base.cpp
```diff
@@ -0,0 +1,816 @@
...
+  // Assert
+  EXPECT_EQ(created, captured_last_group);
+  EXPECT_TRUE(last_group_known_to_node);
+}
+
+// =============================================================================
+// Category 10: Constructor-argument accessors (get_local_args / get_global_args)
+//
+// Specification:
```
atsushi421, line 802: This spec says get_global_args() returns nullptr when use_global_arguments is false, otherwise it reflects the global agnocast context. The implementation also returns nullptr when use_global_arguments is true but g_context.is_initialized() is false (node_base.cpp:39-44). Update the spec to: returns nullptr when use_global_arguments is false, OR when use_global_arguments is true but agnocast::init() has not been called; only when both conditions hold does it return a non-null pointer.

atsushi421 (commented):
Constructor coverage gap: NodeBase's constructor calls rcl_remap_node_name and rcl_remap_node_namespace and throws std::runtime_error on failure (node_base.cpp:48-82), but no test exercises the remap success path or the throw path. Tests for get_name/get_namespace/get_fully_qualified_name only cover pass-through. Add tests passing NodeOptions::arguments({"--ros-args", "-r", "__node:=remapped"}) and -r __ns:=/remapped_ns, asserting the accessors return the remapped values.

src/agnocastlib/test/unit/test_node_base.cpp
Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1001/commits/4eaca28a31a33050cfd8bb4f119420fcd83a5bf3.

src/agnocastlib/test/unit/test_node_base.cpp
Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1001/commits/414121a4c0a7e1354a2e0123dacae65121eea4db (and also in https://github.com/autowarefoundation/agnocast/pull/1001/commits/4eaca28a31a33050cfd8bb4f119420fcd83a5bf3).

src/agnocastlib/test/unit/test_node_base.cpp
Koichi98: I considered the same refactor, but `TEST_P` hurts a bit of readability and the savings are marginal after fixture/`INSTANTIATE_TEST_SUITE_P` boilerplate. For just two boolean variants, I'd prefer to keep the explicit TEST_F form.

src/agnocastlib/test/unit/test_node_base.cpp
Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1001/commits/71287ba7bf62963233b944c3da5b9f1c903bd98e.

src/agnocastlib/test/unit/test_node_base.cpp
Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1001/commits/af8369955ca14f5f0f761e41542f123cd7c542be.

src/agnocastlib/test/unit/test_node_base.cpp
Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1001/commits/af8369955ca14f5f0f761e41542f123cd7c542be.

Koichi98:
@atsushi421 
> Add tests passing NodeOptions::arguments({"--ros-args", "-r", "__node:=remapped"}) and -r __ns:=/remapped_ns, asserting the accessors return the remapped values.

Added tests in https://github.com/autowarefoundation/agnocast/pull/1001/commits/570d13b3b6cce1ecee4f58eab934691f359a10ae.

atsushi421 (approved):

PR #1008, opened 2026-02-06
fix(agnocast_e2e_test): 1to1 e2e test failure when Bridge Mode is off

kobayu858:
## Description
The e2e_test_1to1_with_ros2_sub was previously failing when Bridge Mode was set to OFF. This PR corrects the test behavior for this scenario.

Changes:

- Modified the test to handle the "Bridge OFF" state correctly.
- When the bridge is disabled, the test now focuses on verifying pure Agnocast-to-Agnocast communication.

In addition, renamed e2e_test_1to1_with_ros2_sub to e2e_test_1to1. The test no longer strictly requires a ROS 2 subscriber. It now serves as a generic 1-to-1 communication test that adapts to the environment (Bridge ON or OFF), so the new name better reflects its actual scope.

## How was this PR tested?

- [ ] Autoware (required)
- [x] `bash scripts/e2e_test_1to1` (required)
- [ ] sample application

**Important notes:**


scripts/e2e_test_1to1
```diff
@@ -42,30 +42,44 @@ while true; do
     esac
 done
 
+# Check bridge mode
+LOWER_BRIDGE_MODE=$(echo "$AGNOCAST_BRIDGE_MODE" | tr '[:upper:]' '[:lower:]')
+[[ "$LOWER_BRIDGE_MODE" =~ ^(0|off)$ ]] && BRIDGE_OFF=true || BRIDGE_OFF=false
```
Koichi98: Could you rewrite this as follow?
```
  if [[ "$LOWER_BRIDGE_MODE" =~ ^(0|off)$ ]]; then
      BRIDGE_OFF=true
  else
      BRIDGE_OFF=false
  fi
```
This would be more explicit.

PR #1085, opened 2026-02-19
fix(agnocast_kmod): fix string handling bugs in node queries and get_version

k1832:
## Description

- Replace `strncmp` with `strcmp` for exact node name matching in `ioctl_get_node_subscriber_topics` and `ioctl_get_node_publisher_topics`. The old `strncmp` would incorrectly match any node whose name started with the query prefix.
- Replace `strncpy` with `strscpy` in `ioctl_get_topic_subscriber_info` and `ioctl_get_topic_publisher_info` to guarantee NUL-termination.
- Replace `memcpy` with `strscpy` in `ioctl_get_version` to respect buffer size.
- Add KUnit tests for node name exact matching and get_version.

## How was this PR tested?

- [ ] Autoware (required)
- [x] `bash scripts/e2e_test_1to1` (required)
- [ ] sample application

`need-patch-update`


agnocast_kmod/agnocast_main.c
```diff
@@ -2865,6 +2865,62 @@ pid_t get_bridge_owner_pid(const char * topic_name, const struct ipc_namespace *
...
+        count++;
+        break;
+      }
+    }
+  }
+  return count;
+}
+
+// ioctl_get_version is static, so expose it for KUnit.
+int get_version_for_test(struct ioctl_get_version_args * ioctl_ret)
```
Koichi98: We don't need this wrapper function for test, since all other `ioctl_` prefixed functions which directly corresponds to `IOCTL_AGNOCAST_*_CMD` are called directly from kunit tests. You can just convert `ioctl_get_version` to non-static function for test.

agnocast_kmod/agnocast_main.c
```diff
@@ -2865,6 +2865,62 @@ pid_t get_bridge_owner_pid(const char * topic_name, const struct ipc_namespace *
...
+      if (strcmp(sub_info->node_name, node_name) == 0) {
+        count++;
+        break;
+      }
+    }
+  }
+  return count;
+}
+
+int count_node_publisher_topics(const struct ipc_namespace * ipc_ns, const char * node_name)
```
Koichi98: These helper functions duplicate the core logic of `ioctl_get_node_subscriber_topics` / `ioctl_get_node_publisher_topics` rather than testing the actual functions. Following the same convention as above, please make `ioctl_get_node_subscriber_topics` and `ioctl_get_node_publisher_topics` non-static and test them directly.

agnocast_kmod/agnocast_main.c
Koichi98: > As a quick note regarding this: since making a function non-static exposes it to the global namespace, Linux module convention usually adds a module-specific prefix to prevent namespace pollution (e.g. agnocast_ioctl_get_version).

Nice share! I will fix this in a few days. Thank you for pointing out.

Koichi98 (approved):

PR #1089, opened 2026-02-20
feat(agnocast_kmod): replace do_exit kprobe with sched_process_exit tracepoint

k1832:
## Description

Replace the `do_exit` kprobe with a `sched_process_exit` tracepoint for process exit detection.

**Why:**

| Aspect | kprobe on `do_exit` | `sched_process_exit` tracepoint |
|--------|---------------------|--------------------------------|
| **Stability** | `do_exit` is an internal symbol — no stability guarantee across kernel versions | Stable kernel ABI since Linux 2.6.35 (2010) |
| **Hook mechanism** | `int3` breakpoint trap (x86) or `BRK` (arm64) | Static key (`nop` when disabled, `jmp` when enabled) |
| **Cost per invocation** | ~0.5-1.0 us (trap + single-step) | ~20-50 ns (direct call) |
| **Preemption** | Handler runs with preemption disabled | Callback runs in normal preemptible context |
| **Unload safety** | `unregister_kprobe()` — no explicit "all callbacks done" guarantee | `tracepoint_synchronize_unregister()` — formal guarantee no callbacks in flight |
| **Task context** | Uses `current->pid` | Receives `task_struct *` directly — correct by construction |
| **Version compat** | Needs `#if LINUX_VERSION_CODE` guards (6.2+ `__noreturn`, 6.7+ inlining) | Identical API across all target kernels (5.x, 6.x) |

The tracepoint symbol (`__tracepoint_sched_process_exit`) is not exported to modules, so we use dynamic lookup via `for_each_kernel_tracepoint()` + `tracepoint_probe_register()`, both of which are `EXPORT_SYMBOL_GPL` and compatible with the module's `MODULE_LICENSE("Dual BSD/GPL")`.

**Behavioral equivalence:** The kprobe handler and tracepoint callback both call `enqueue_exit_pid(pid)`. The cleanup logic (`process_exit_cleanup`, `exit_worker_thread`, ring buffer) is completely unchanged. `task->pid` is valid at both hook points — the PID is not freed until `exit_notify()` → `release_task()`, which runs after the tracepoint.

**Benchmark (50000 × `/bin/true` on same machine):**

| Branch | real | user | sys |
|--------|------|------|-----|
| main (kprobe) | 3m10.299s | 0m29.098s | 2m40.464s |
| tracepoint | 3m10.931s | 0m28.792s | 2m40.739s |

No measurable throughput difference. This is expected — the ~1 us per-call saving is invisible in a test dominated by fork+exec+exit cost (~3.8 ms per iteration). The real value is portability (stable API), unload safety (`tracepoint_synchronize_unregister`), and eliminating preemption-disabled windows.

## Related links

- #601

## How was this PR tested?

- [ ] Autoware (required)
- [x] `bash scripts/test/e2e_test_1to1` (required)
- [x] `bash scripts/test/e2e_test_2to2` (required)
- [ ] sample application

KUnit tests do not need changes — they call `enqueue_exit_pid()` / `process_exit_cleanup()` directly and do not exercise the hook mechanism.

## Notes for reviewers

- `for_each_kernel_tracepoint()` is used for dynamic tracepoint lookup because the static tracepoint symbol is not exported to modules. This is a standard pattern used by other out-of-tree modules.
- `tracepoint_synchronize_unregister()` is called during module exit to guarantee no callbacks are in flight before the module is unloaded.

### DIFF
--- a/scripts/test/test_rmmod_refcount
+++ b/scripts/test/test_rmmod_refcount
@@ -3,7 +3,7 @@
 set -u
 
 MODULE_NAME="agnocast"
-KMOD_DIR="$(cd "$(dirname "$0")/../agnocast_kmod" && pwd)"
+KMOD_DIR="$(cd "$(dirname "$0")/../../agnocast_kmod" && pwd)"
 
 # Check sudo is cached (non-interactive)
 if ! sudo -n true 2>/dev/null; then
--- a/src/agnocastlib/include/agnocast/agnocast_callback_isolated_executor.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_callback_isolated_executor.hpp
@@ -10,6 +10,7 @@ class CallbackIsolatedAgnocastExecutor : public rclcpp::Executor
   RCLCPP_DISABLE_COPY(CallbackIsolatedAgnocastExecutor)
 
   const int next_exec_timeout_ms_;
+  const int monitor_polling_interval_ms_;
 
   // Nodes associated with this AgnocastCallbackIsolatedExecutor, appended by add_node() and removed
   // by remove_node()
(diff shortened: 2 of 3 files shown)


Koichi98:
@k1832 
Looks good! I will approve once you resolve the conflicts.

k1832:
@Koichi98

Thank you for the review! I resolved the conflicts and Copilot's comments too. Please take a look again!

Koichi98:
@k1832 
You should've added the sign-off to each commit to fix `DCO` in the CI.

https://github.com/autowarefoundation/agnocast/pull/1089/checks?check_run_id=64777556508
Could you follow the instruction written in this link?

k1832:
@Koichi98 
Oops, sorry about that. Done!

Koichi98 (approved):

PR #1150, opened 2026-03-05
fix(agnocastlib): unlink message queue in IpcEventLoopBase cleanup

sykwer:
## Description
See https://github.com/autowarefoundation/agnocast/pull/1150#issuecomment-4004503169

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

**Important notes:**

### DIFF
--- a/src/agnocastlib/include/agnocast/bridge/agnocast_bridge_ipc_event_loop_base.hpp
+++ b/src/agnocastlib/include/agnocast/bridge/agnocast_bridge_ipc_event_loop_base.hpp
@@ -263,16 +263,12 @@ inline void IpcEventLoopBase::cleanup_resources()
         logger_, "Failed to close mq_fd for mq_name='" << mq_name_ << "': " << strerror(errno));
     }
     mq_fd_ = -1;
-  }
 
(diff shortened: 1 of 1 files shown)


sykwer:
## Problem

`IpcEventLoopBase::cleanup_resources()` closes the message queue file descriptor (`mq_close`) but never calls `mq_unlink()`. This leaves `/dev/mqueue/agnocast@*` entries behind when a bridge process crashes or exits abnormally. In the normal shutdown path, the daemon or subscription destructors happen to unlink first, masking the issue.

## Change

Add `mq_unlink()` after `mq_close()` in `IpcEventLoopBase::cleanup_resources()`, with `ENOENT` tolerance to handle the case where another path already removed it.

Crucially, the `mq_unlink()` call is placed **inside** the `if (mq_fd_ != -1)` guard, so it only executes when this instance successfully opened the queue. Without this guard, if `mq_open()` had failed (e.g., `EMFILE`/`ENFILE`), `cleanup_resources()` would unlink a queue belonging to another still-running process that uses the same name (e.g., the performance bridge manager), causing disruption.

PR #1172, opened 2026-03-10
feat(agnocastlib): add --enable-rosout-logs full support with test

PeterWrighten:
## Description
add `--enable-rosout-logs` full support.

Since agnocast cannot use `rcl_logging_configure()` (it requires `rcl_node_t` and spdlog), the flag is detected by manually scanning `argv` within the `--ros-args` scope during `Context::init()`.
                                                                                                              
  When enabled, the first constructed Node calls `setup_rosout_handler()`, which creates a `/rosout` publisher    
  (reliable, transient_local, depth 100) and installs a custom rcutils log output handler. The handler chains
  the original console handler, then formats and publishes each log message as `rcl_interfaces/msg/Log` via     
  loaned messages, with a recursion guard to prevent infinite loops during publishing.

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

**Important notes:**

### DIFF
--- a/src/agnocastlib/src/node/agnocast_rosout.cpp
+++ b/src/agnocastlib/src/node/agnocast_rosout.cpp
@@ -1,0 +1,131 @@
+#include "agnocast/node/agnocast_rosout.hpp"
+
+#include "agnocast/agnocast_publisher.hpp"
+#include "agnocast/node/agnocast_node.hpp"
+
+#include <rcl_interfaces/msg/log.hpp>
+
+#include <rcutils/logging.h>
+
+#include <atomic>
+#include <cstdarg>
+#include <cstdio>
+#include <mutex>
+
+namespace agnocast
+{
+
+static Publisher<rcl_interfaces::msg::Log>::SharedPtr g_rosout_pub;
+static std::mutex g_rosout_mtx;
+static std::atomic<bool> g_in_handler{false};
+static std::atomic<bool> g_rosout_initialized{false};
+static rcutils_logging_output_handler_t g_original_handler = nullptr;
+
+static uint8_t severity_to_log_level(int severity)
+{
+  switch (severity) {
+    case RCUTILS_LOG_SEVERITY_DEBUG:
+      return rcl_interfaces::msg::Log::DEBUG;
+    case RCUTILS_LOG_SEVERITY_INFO:
+      return rcl_interfaces::msg::Log::INFO;
+    case RCUTILS_LOG_SEVERITY_WARN:
+      return rcl_interfaces::msg::Log::WARN;
+    case RCUTILS_LOG_SEVERITY_ERROR:
+      return rcl_interfaces::msg::Log::ERROR;
+    case RCUTILS_LOG_SEVERITY_FATAL:
+      return rcl_interfaces::msg::Log::FATAL;
+    default:
+      return rcl_interfaces::msg::Log::INFO;
+  }
+}
+
+static void rosout_output_handler(
+  const rcutils_log_location_t * location, int severity, const char * name,
+  rcutils_time_point_value_t timestamp, const char * format, va_list * args) noexcept
+{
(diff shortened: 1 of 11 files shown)


src/agnocastlib/src/node/agnocast_rosout.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+#include <atomic>
+#include <cstdarg>
+#include <cstdio>
+#include <mutex>
+
+namespace agnocast
+{
+
+static Publisher<rcl_interfaces::msg::Log>::SharedPtr g_rosout_pub;
```
Koichi98: rclcpp gives each Node its own `/rosout` publisher with per-node QoS, and `rcl_logging_rosout` dispatches log messages by logger name to the matching publisher. On the other hand, this implementation uses a single process-global publisher with a hard-coded QoS instead. Could we align with rclcpp — per-Node publisher keyed by logger name, with QoS taken from the node?

PeterWrighten: After my research, I consider we could refactor into the rclcpp style implementation. I would show the research result in the end and try to make a plan to implement.

PeterWrighten:
## [WIP] `rclcpp` style refactor

#### What rclcpp actually does
  
- rclcpp's approach:
  - Each `rclcpp::Node calls rcl_logging_rosout_add_publisher_to_node(rcl_node_handle)` during construction,
  which creates a per-node /rosout publisher inside rcl and registers it in a hash map keyed by logger name
  - The rcl output handler (`rcl_logging_rosout_output_handler`) looks up that map by the name argument (logger
  name) and publishes through the matching publisher
  - QoS comes from `NodeOptions::rosout_qos()` — defaults to `rclcpp::RosoutQoS() = depth 1000`, reliable,
  transient_local
  - On node destruction, `rcl_logging_rosout_remove_publisher_for_node` removes the entry

- Current agnocast:
  - One `static Publisher<...>::SharedPtr g_rosout_pub` for the whole process
  - Hard-coded QoS depth 100 (vs. rclcpp's 1000)
  - The `g_rosout_initialized atomic` means only the first node ever creates a publisher; subsequent nodes are
  no-ops
  - All log messages go through that one publisher regardless of which node logged them
  - No cleanup on node destruction

  ---
  What needs to change

  Since agnocast uses its own `agnocast::Publisher` (not `rcl_publisher_t`), it can't call
  `rcl_logging_rosout_add_publisher_to_node`. It needs its own equivalent map.

1. `agnocast_rosout.cpp` — replace the single publisher with a logger-name-keyed map:

```cpp
  // before
  static Publisher<rcl_interfaces::msg::Log>::SharedPtr g_rosout_pub;
  static std::atomic<bool> g_rosout_initialized{false};

  // after
  static std::unordered_map<std::string, Publisher<rcl_interfaces::msg::Log>::SharedPtr> g_rosout_pubs;
  static std::atomic<bool> g_handler_installed{false};
```

  The output handler looks up name in the map instead of using the single global. setup_rosout_handler
  registers the calling node's publisher under `node->get_logger().get_name()`, and installs the rcutils handler
   only on the first call. Add `teardown_rosout_handler(Node * node)` to erase the entry.

  2. `agnocast_rosout.hpp` — update signature and expose teardown:

```cpp
  void setup_rosout_handler(Node * node, const rclcpp::QoS & qos);
  void teardown_rosout_handler(Node * node);
```

  3. `agnocast_node.cpp` — pass QoS from options and add destructor call:
  
```cpp
// constructor
  if (g_context.is_rosout_enabled()) {
    setup_rosout_handler(this, options.rosout_qos());  // uses per-node QoS
  }

  // destructor (needs to be added)
  Node::~Node() {
    if (g_context.is_rosout_enabled()) {
      teardown_rosout_handler(this);
    }
  }
```

  4. `agnocast_node.hpp` — declare `~Node()`.

PR #1173, opened 2026-03-10
feat(agnocastlib, test): add --disable-stdout-logs support with test

PeterWrighten:
## Description
add `--disable-stdout-logs` support with test

When agnocast::init() is called, Context::init() in agnocast_context.cpp does two things:                   
                                                                                                              
  1. Always: calls `rcl_logging_configure_with_output_handler(…, rcl_logging_multiple_output_handler)` to route 
  logs to file via spdlog (the normal path inherited from main).                                              
  2. Conditionally: scans `argv` directly for `--disable-stdout-logs` within the `--ros-args … -- `scope. If found, 
  it immediately overrides the handler with a local `noop_log_output_handler` — a no-op function that silently  
  drops every log call.                                                                                       
                                                                                                              
  The argv scan is manual because `rcl_arguments_t` has no public API to query this flag, and using             
  `rcl_logging_configure()` in agnocast is avoided since it would try to set up a `rosout` publisher, which       
  requires an `rcl_node_t` that agnocast doesn't have. 

## Related links

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

## Notes for reviewers

**Important notes:**

### DIFF
--- a/src/agnocastlib/test/unit/test_agnocast_context.cpp
+++ b/src/agnocastlib/test/unit/test_agnocast_context.cpp
@@ -1,0 +1,91 @@
+#include "agnocast/node/agnocast_context.hpp"
+
+#include <gtest/gtest.h>
+#include <rcl/logging.h>
+#include <rcutils/logging.h>
+#include <rcutils/logging_macros.h>
+
+// =========================================
+// agnocast::Context --disable-stdout-logs tests
+// =========================================
+
+class AgnocastContextStdoutLogsTest : public ::testing::Test
+{
+protected:
+  void SetUp() override { original_handler_ = rcutils_logging_get_output_handler(); }
+
+  void TearDown() override
+  {
+    // Finalize rcl logging so the next test can call rcl_logging_configure again.
+    rcl_ret_t ret = rcl_logging_fini();
+    (void)ret;
+    // Restore the original handler so other tests are not affected.
+    rcutils_logging_set_output_handler(original_handler_);
+  }
+
+  rcutils_logging_output_handler_t original_handler_;
+};
+
+TEST_F(AgnocastContextStdoutLogsTest, disable_stdout_logs_suppresses_console_output)
+{
+  const char * argv[] = {"program", "--ros-args", "--disable-stdout-logs", "--log-level", "debug"};
+  int argc = 5;
+  agnocast::Context ctx;
+  ctx.init(argc, argv);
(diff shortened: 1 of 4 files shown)


src/agnocastlib/test/unit/test_agnocast_context.cpp
```diff
@@ -0,0 +1,93 @@
...
+  agnocast::Context ctx;
+
+  // Act
+  ctx.init(argc, argv);
+
+  // Assert: flag after -- must be ignored; both stdout and file handlers active.
+  const auto current_handler = rcutils_logging_get_output_handler();
+  EXPECT_EQ(current_handler, rcl_logging_multiple_output_handler);
+  EXPECT_NE(current_handler, original_handler_);
+}
```
Koichi98, line 93: These tests only check that the top-level handler is `rcl_logging_multiple_output_handler`, which is always true regardless of the flag.  Could we capture stderr around a `RCUTILS_LOG_INFO_NAMED` call and assert no output is produced when the flag is set?

src/agnocastlib/test/unit/test_agnocast_context.cpp
```diff
@@ -0,0 +1,91 @@
...
+TEST_F(AgnocastContextStdoutLogsTest, disable_stdout_logs_suppresses_console_output)
+{
+  const char * argv[] = {"program", "--ros-args", "--disable-stdout-logs", "--log-level", "debug"};
+  int argc = 5;
+  agnocast::Context ctx;
+  ctx.init(argc, argv);
+
+  testing::internal::CaptureStderr();
+  RCUTILS_LOG_INFO_NAMED("agnocast_test", "test message");
+  fflush(stdout);
```
Koichi98: The flag is named `--disable-stdout-logs` (per ROS 2 convention), but the rcutils/rcl default console handler actually writes to **stderr** — which is why this test uses `CaptureStderr()`. A short comment noting this discrepancy would help future readers avoid confusion.

Also, `fflush(stdout)` here is a no-op for this test: stderr is unbuffered, and we're not capturing stdout anyway. It can be removed.

PeterWrighten: I have added corresponding comments and removed `fflush(stdout)`. Please review.

Koichi98 (approved):

PR #1207, opened 2026-03-30
refactor(agnocastlib): refine bridge cleanup

bdm-k:
## Description

This PR optimizes cleanup of standard bridges by switching from deferred, key-based removal to iterator-based, in-place removal. The function `check_active_bridges()` now inspects `active_bridges_` and removes bridges during a single traversal. The removal criteria are encapsulated in the lambda `should_remove()` for better readability.

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

## Notes for reviewers

- @Koichi98 Currently, `check_active_bridges()` is updating `managed_bridges_` when the Agnocast entity no longer exists. But, can we defer that to `check_managed_bridges()` (in the next spin cycle)? If so, I'll include the patch in this PR.

### DIFF
--- a/src/agnocast_cie_thread_configurator/src/prerun_node.cpp
+++ b/src/agnocast_cie_thread_configurator/src/prerun_node.cpp
@@ -15,9 +15,14 @@ PrerunNode::PrerunNode(const std::set<size_t> & domain_ids) : Node("prerun_node"
 {
   size_t default_domain_id = agnocast_cie_thread_configurator::get_default_domain_id();
 
+  auto cbg_qos = rclcpp::QoS(rclcpp::KeepAll()).reliable().transient_local();
+  // volatile: publisher context in spawn_non_ros2_thread is destroyed after publish,
+  // so transient_local is ineffective.
+  auto non_ros_thread_qos = rclcpp::QoS(rclcpp::KeepAll()).reliable();
+
   // Create subscription for non-ROS thread info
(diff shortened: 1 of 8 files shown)


src/agnocastlib/src/bridge/standard/agnocast_standard_bridge_manager.cpp
```diff
@@ -299,11 +299,49 @@ void StandardBridgeManager::check_parent_alive()
...
+    } else {
+      count = get_agnocast_publisher_count(topic_name_str).count;
+      is_demanded_by_ros2 =
+        has_external_ros2_subscriber(this->container_node_.get(), topic_name_str);
+      ros2_count_ok = update_ros2_subscriber_num(this->container_node_.get(), topic_name_str);
+    }
 
-  for (const auto & [key, bridge] : active_bridges_) {
+    bool remove = false;
+    bool keep_managed = false;
```
kobayu858: `remove` and `keep_managed` point in opposite directions when true — one means "delete" and the other means "keep" — which makes the pair a bit confusing to read together. 

I appreciate that you matched the parameter name from the old `remove_active_bridge(key, keep_managed)` interface, but since that function is gone now, would you mind aligning them to something like `remove_managed` so both flags point the same way?

bdm-k: fixed in [7e4d928](https://github.com/autowarefoundation/agnocast/pull/1207/commits/7e4d928aed7323f4a75c16a0eac4952384dfe8d3)

PR #1208, opened 2026-03-30
fix: topic list_agnocast warning

Koichi98:
## Description
Include the following two changes:                                               
  - Fix false bridge warnings in ros2 topic list by aligning with the bridge manager's actual conditions (Agnocast pub + ROS2 sub, or Agnocast sub + ROS2 pub)                                                                                                                                                                                                                                                                           
  - Fix resource leaks: missing `close(fd)` in `topic_info.cpp` and missing `free()`/`close()` in topic_list.cpp  

## Related links

## How was this PR tested?
I've confirmed that it now doesn't warn when both agnocast and ros2 publisher exist but the subscriber doesn't exist.

- [x] Autoware (required)
- [ ] sample application

## Notes for reviewers

**Important notes:**

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -14,4 +14,4 @@ src/agnocastlib/src/node @Koichi98 @sykwer @kobayu858
 src/agnocastlib/include/agnocast/node @Koichi98 @sykwer @kobayu858
 
 # ros2agnocast
-src/ros2agnocast/ @TetsuKawa @veqcc @atsushi421 @Koichi98 @sykwer
+src/ros2agnocast/ @TetsuKawa @veqcc @atsushi421 @Koichi98 @sykwer @kobayu858


src/ros2agnocast/ros2agnocast/verb/list_agnocast.py
```diff
@@ -120,19 +130,25 @@ def remove_service_topic(topic_names):
...
                 else:
-                    match get_bridge_status(topic):
+                    bridge_status, has_agnocast_pub, has_agnocast_sub = get_bridge_status(topic)
+                    needs_r2a = has_agnocast_sub and topic in ros2_pub_topics
+                    needs_a2r = has_agnocast_pub and topic in ros2_sub_topics
+                    match bridge_status:
                         case BridgeStatus.BIDIRECTION:
                             suffix = " (Agnocast enabled, bridged)"
                         case BridgeStatus.ROS2_TO_AGNOCAST:
-                            if topic in ros2_pub_topics:
```
TetsuKawa: memo
This if-branch was originally designed to account for the difference between the static initialization of the standard bridge and the dynamic initialization of the performance version, verifying whether a topic is truly being communicated via the bridge even when it is active.

PR #1234, opened 2026-04-07
refactor(thread_configurator): split into separate executables and use ROS parameters

atsushi421:
## Description

Refactor `agnocast_cie_thread_configurator` to split `thread_configurator_node` and `prerun_node` into independent executables, and use ROS parameters (`config_file`, `domains`) instead of CLI arguments. This enables canonical launch file usage and aligns with ROS 2 best practices.

Key changes:
- **Separate executables**: `prerun_node` is now its own executable (previously a `--prerun` mode of `thread_configurator_node`)
- **ROS parameters**: `config_file` and `domains` are now ROS parameters, enabling declarative launch file configuration
- **Launch file**: Added `thread_configurator.launch.xml` as the canonical way to launch either node
- **Direct `ros2 run` support**: `prerun_node` can also be run directly with `ros2 run ... --ros-args -p domains:=[0,1]` to avoid the extra launcher process
- **CLI backward compatibility**: `main.cpp` still accepts `--prerun`, `--config-file`, `--domains` with a deprecation warning. This is because breaking changes to the public API are not expected within the same ROS distribution.
- **Validation moved into nodes**: `validate_hardware_info` and YAML loading moved from `main.cpp` into `ThreadConfiguratorNode` constructor
- **Error handling**: Node construction/spin wrapped in try/catch for clean shutdown on failure; YAML load errors include the config file path for easier diagnosis
- **Portability**: Fixed non-portable `%ld` format specifier for `int64_t` in `prerun_node.cpp`

## Related links

[JIRA ticket (TIER IV internal)](https://tier4.atlassian.net/browse/T4DEV-51207)

## How was this PR tested?

- [ ] Autoware (required)
- [x] sample application

**Important notes:**

### DIFF
--- a/docs/callback_isolated_executor_for_agnocast.md
+++ b/docs/callback_isolated_executor_for_agnocast.md
@@ -34,9 +34,13 @@ agnocast_components_register_node(
 
 #### Multiple ROS domain support
 
-`agnocast_cie_thread_configurator` can handle callback groups from multiple ROS domains. Use launch arguments to specify domain IDs:
+`agnocast_cie_thread_configurator` can handle callback groups from multiple ROS domains. Use ROS parameters to specify domain IDs:
 
 ```bash
+# Option 1: Run prerun_node directly with ROS parameters
+ros2 run agnocast_cie_thread_configurator prerun_node --ros-args -p domains:=[0,1]
+
+# Option 2: Use the launch file
 ros2 launch agnocast_cie_thread_configurator thread_configurator.launch.xml prerun:=true domains:=[0,1]
 ```
 
--- a/docs/callback_isolated_executor_for_agnocast.md
+++ b/docs/callback_isolated_executor_for_agnocast.md
@@ -6,11 +6,11 @@
 
 ### Package, Class, and Executable Names
 
-| | Original | Agnocast Version |
-|--|----------|------------------|
-| Package | `callback_isolated_executor` | `agnocast_components` (recommended) or `agnocastlib` (deprecated) |
-| Executor Class | `CallbackIsolatedExecutor` | `agnocast::CallbackIsolatedAgnocastExecutor` |
-| Container Executable | `component_container_callback_isolated` | `agnocast_component_container_cie` |
+|                      | Original                                | Agnocast Version                                                  |
+| -------------------- | --------------------------------------- | ----------------------------------------------------------------- |
+| Package              | `callback_isolated_executor`            | `agnocast_components` (recommended) or `agnocastlib` (deprecated) |
+| Executor Class       | `CallbackIsolatedExecutor`              | `agnocast::CallbackIsolatedAgnocastExecutor`                      |
+| Container Executable | `component_container_callback_isolated` | `agnocast_component_container_cie`                                |
 
 ### Features
 
@@ -34,10 +34,10 @@ agnocast_components_register_node(
 
 #### Multiple ROS domain support
 
-`agnocast_cie_thread_configurator` can handle callback groups from multiple ROS domains. Use the `--domains` option to specify domain IDs you use:
+`agnocast_cie_thread_configurator` can handle callback groups from multiple ROS domains. Use launch arguments to specify domain IDs:
 
 ```bash
-ros2 run agnocast_cie_thread_configurator thread_configurator_node --prerun --domains 0,1
+ros2 launch agnocast_cie_thread_configurator thread_configurator.launch.xml prerun:=true domains:=[0,1]
 ```
 
 #### RT Throttling
--- a/src/agnocast_cie_thread_configurator/src/main.cpp
+++ b/src/agnocast_cie_thread_configurator/src/main.cpp
@@ -48,10 +48,11 @@ int main(int argc, char * argv[])
 
   if (prerun_mode || !domain_ids.empty() || !config_filename.empty()) {
     std::cerr << "[DEPRECATED] CLI arguments (--prerun, --domains, --config-file) are deprecated. "
-              << "Use the launch file (thread_configurator.launch.xml) with ROS parameters instead."
-              << std::endl;
+              << "Run prerun_node directly with ROS parameters or use the launch file "
+              << "(thread_configurator.launch.xml) instead." << std::endl;
   }
 
+  // Backward-compatible CLI path: translates deprecated CLI args into ROS parameter overrides.
   try {
     if (prerun_mode) {
       std::cout << "prerun mode" << std::endl;
--- a/src/agnocast_cie_thread_configurator/src/main.cpp
+++ b/src/agnocast_cie_thread_configurator/src/main.cpp
@@ -3,7 +3,9 @@
 #include "rclcpp/rclcpp.hpp"
 
 #include <filesystem>
+#include <iostream>
 #include <memory>
+#include <sstream>
 #include <string>
 
 static std::vector<int64_t> parse_domain_ids(const std::string & domains_str)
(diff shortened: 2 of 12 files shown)


src/agnocast_cie_thread_configurator/src/main.cpp
```diff
@@ -156,16 +46,50 @@ int main(int argc, char * argv[])
...
-      return 1;
+    rclcpp::NodeOptions options;
+    if (!config_filename.empty()) {
+      options.append_parameter_override("config_file", config_filename);
+    }
+
+    auto node = std::make_shared<ThreadConfiguratorNode>(options);
+    auto executor = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
+
+    executor->add_node(node);
```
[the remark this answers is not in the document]
atsushi421: Fixed. Wrapped both branches in main.cpp and prerun_node_main.cpp with try/catch that logs the error, calls rclcpp::shutdown(), and returns 1.

src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
```diff
@@ -18,11 +18,28 @@
...
+    throw std::runtime_error(
+      "'config_file' parameter must be provided with a valid YAML file path.");
+  }
+
+  YAML::Node yaml;
+  try {
+    yaml = YAML::LoadFile(config_file);
+  } catch (const std::exception & e) {
+    RCLCPP_ERROR(this->get_logger(), "Error reading the YAML file: %s", e.what());
+    throw;
```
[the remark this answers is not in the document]
atsushi421: Fixed. The error log and rethrown exception now include the config_file path.

src/agnocast_cie_thread_configurator/src/prerun_node.cpp
```diff
@@ -9,10 +9,35 @@
...
+  constexpr size_t max_domain_id = 101;
+
+  const auto domains =
+    this->declare_parameter<std::vector<int64_t>>("domains", std::vector<int64_t>{});
+  std::set<size_t> domain_ids;
+  for (const auto raw_domain_id : domains) {
+    if (raw_domain_id < 0) {
+      RCLCPP_WARN(
+        this->get_logger(), "Negative domain ID %ld is invalid. Skipping.", raw_domain_id);
+      continue;
```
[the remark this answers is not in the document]
atsushi421: Fixed. Changed to cast to long long with %lld.

docs/callback_isolated_executor_for_agnocast.md
```diff
@@ -34,10 +34,10 @@ agnocast_components_register_node(
 
 #### Multiple ROS domain support
 
-`agnocast_cie_thread_configurator` can handle callback groups from multiple ROS domains. Use the `--domains` option to specify domain IDs you use:
+`agnocast_cie_thread_configurator` can handle callback groups from multiple ROS domains. Use launch arguments to specify domain IDs:
 
 ```bash
-ros2 run agnocast_cie_thread_configurator thread_configurator_node --prerun --domains 0,1
+ros2 launch agnocast_cie_thread_configurator thread_configurator.launch.xml prerun:=true domains:=[0,1]
```
paulsohn, line 44: On the second thought, launcher will get its own process so it might not be always optimal to use launcher. We can also add a guide to use `ros2 run agnocast_cie_thread_configurator prerun_node --ros-args -p domains:=[0,1]` if needed.

src/agnocast_cie_thread_configurator/src/main.cpp
```diff
@@ -156,16 +46,56 @@ int main(int argc, char * argv[])
...
-      return 1;
+  if (prerun_mode || !domain_ids.empty() || !config_filename.empty()) {
+    std::cerr << "[DEPRECATED] CLI arguments (--prerun, --domains, --config-file) are deprecated. "
+              << "Use the launch file (thread_configurator.launch.xml) with ROS parameters instead."
+              << std::endl;
+  }
+
+  try {
+    if (prerun_mode) {
+      std::cout << "prerun mode" << std::endl;
```
paulsohn, line 58: It would be helpful to have a reminder comment indicating this branch is for backward compatibility purpose.

src/agnocast_cie_thread_configurator/src/main.cpp
```diff
@@ -156,16 +46,56 @@ int main(int argc, char * argv[])
...
-    spin_prerun_node(domain_ids);
-  } else {
-    if (config_filename.empty()) {
-      std::cerr << "[ERROR] --config-file must be provided when not running in --prerun mode."
-                << std::endl;
-      rclcpp::shutdown();
-      return 1;
+  if (prerun_mode || !domain_ids.empty() || !config_filename.empty()) {
+    std::cerr << "[DEPRECATED] CLI arguments (--prerun, --domains, --config-file) are deprecated. "
+              << "Use the launch file (thread_configurator.launch.xml) with ROS parameters instead."
```
paulsohn: *"run `prerun_node` directly with ros params or use the launch file" (indicating there is a non-launcher option)

src/agnocast_cie_thread_configurator/src/prerun_node.cpp
```diff
@@ -9,10 +9,36 @@
...
+  constexpr size_t max_domain_id = 101;
+
+  const auto domains =
+    this->declare_parameter<std::vector<int64_t>>("domains", std::vector<int64_t>{});
+  std::set<size_t> domain_ids;
+  for (const auto raw_domain_id : domains) {
+    if (raw_domain_id < 0) {
+      RCLCPP_WARN(
+        this->get_logger(), "Negative domain ID %lld is invalid. Skipping.",
+        static_cast<long long>(raw_domain_id));
```
paulsohn, line 27: nit: why are we casting to long long? Is `size_t` unusable here?

docs/callback_isolated_executor_for_agnocast.md
atsushi421: Added the ros2 run alternative alongside the launch file example in the docs, so users can choose to avoid the extra launcher process. See 8a080713.

src/agnocast_cie_thread_configurator/src/main.cpp
atsushi421: Added a comment above the try block indicating this is the backward-compatible CLI path. See 8a080713.

src/agnocast_cie_thread_configurator/src/main.cpp
atsushi421: Updated the deprecation message to mention both running prerun_node directly with ROS parameters and using the launch file. See 8a080713.

src/agnocast_cie_thread_configurator/src/prerun_node.cpp
atsushi421: raw_domain_id is int64_t and this log is inside the `if (raw_domain_id < 0)` branch, so the value is necessarily negative. size_t is unsigned and would wrap negative values into large positive numbers, producing misleading log output. The long long cast with %lld is the portable way to printf a signed int64_t value.

paulsohn (approved):
No objections for interface.

src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
```diff
@@ -18,11 +18,29 @@
...
+      "'config_file' parameter must be provided with a valid YAML file path.");
+  }
+
+  YAML::Node yaml;
+  try {
+    yaml = YAML::LoadFile(config_file);
+  } catch (const std::exception & e) {
+    RCLCPP_ERROR(
+      this->get_logger(), "Error reading the YAML file '%s': %s", config_file.c_str(), e.what());
+    throw std::runtime_error("Error reading the YAML file '" + config_file + "': " + e.what());
```
kobayu858: The thrown exception is caught in both `main.cpp` and `prerun_node_main.cpp` and printed via `std::cerr << "[ERROR] " << e.what()`, so the user sees the same message twice. Removing the `RCLCPP_ERROR` and keeping only the `throw` would be simpler.

atsushi421: Removed the RCLCPP_ERROR call. The exception message already contains the file path and error details, and the catch blocks in main.cpp / prerun_node_main.cpp will print it via stderr. See e6c66aa1.

kobayu858 (approved):

PR #1239, opened 2026-04-08
refactor(agnocast_kmod): extract kunit test object list into Makefile.inc

k1832:
## Description

Extract the kunit test entries from the root `agnocast_kmod/Makefile` into `agnocast_kunit/Makefile.inc` using `addprefix` to avoid repeating the `agnocast_kunit/` directory prefix. The root Makefile now includes this file with `include $(src)/agnocast_kunit/Makefile.inc`.

This keeps the kunit test list co-located with the test sources, making it easier to maintain as tests are added or removed.

Also:
- Fixes tab/space indentation inconsistency in the `ifeq` block (tabs -> spaces to match the rest of the file).
- Reorders the top-level `ifneq`/`else` to `ifeq`/`else` so the positive condition (`KUNIT_BUILD=y`) comes first. The original used `ifneq` to place the normal build path first, but now that the kunit block is shorter (3 lines + include), having the kunit path first reads more naturally and avoids the double-negative.

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

## Notes for reviewers

This is a Makefile-only refactor. No source code changes, no build behavior should change.

**Important notes:**


Koichi98:
@k1832 
I've checked that the kunit tests build&work correctly.

Koichi98 (approved):

PR #1250, opened 2026-04-12
chore: add sparse check for agnocast-kmod

Koichi98:
## Description
Add `make sparse` target to `agnocast_kmod/Makefile` for running the Linux kernel sparse static analysis tool                                                                                                                                                                                                                                                                                                                        
  - Fix sparse warnings in `agnocast_ioctl.c`: add missing `__user` annotation to casts at lines 1604 and 1690                                                                                                                                                                                                                                                                                                                           
  - Add `sparse-check.yaml` CI workflow that runs automatically on PRs with `agnocast_kmod/` changes (required)                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                                                                                                         
### Background                                                                                                                                                                                                                                                                                                                                                                                                                          
sparse (`C=2`) detects address space annotation mismatches (e.g., mixing `__user` and kernel pointers) that the compiler does not catch. The two warnings were caused by casting to `struct topic_info_ret *` instead of `struct topic_info_ret __user *`.

## Related links
https://docs.kernel.org/dev-tools/sparse.html

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

## Notes for reviewers

**Important notes:**

### DIFF
--- a/.github/workflows/sparse-check.yaml
+++ b/.github/workflows/sparse-check.yaml
@@ -9,13 +9,10 @@ jobs:
     runs-on: ubuntu-22.04
 
     steps:
-      - name: Set PR fetch depth
-        run: echo "PR_FETCH_DEPTH=$(( ${{ github.event.pull_request.commits }} + 1 ))" >> "${GITHUB_ENV}"
-
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
-          fetch-depth: ${{ env.PR_FETCH_DEPTH }}
+          fetch-depth: 0
 
       - name: Check for agnocast_kmod changes
         id: check_changes
--- a/.github/workflows/sparse-check.yaml
+++ b/.github/workflows/sparse-check.yaml
@@ -1,0 +1,24 @@
+name: sparse-check
+
+on:
+  pull_request:
+    paths:
+      - 'agnocast_kmod/**'
+
+jobs:
+  sparse-check:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install sparse
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y sparse
+
+      - name: Run sparse
+        run: |
+          cd agnocast_kmod
+          make sparse
--- a/.github/workflows/sparse-check.yaml
+++ b/.github/workflows/sparse-check.yaml
@@ -2,23 +2,39 @@ name: sparse-check
 
 on:
   pull_request:
-    paths:
-      - 'agnocast_kmod/**'
+    types: [opened, synchronize]
 
 jobs:
   sparse-check:
     runs-on: ubuntu-22.04
 
     steps:
+      - name: Set PR fetch depth
+        run: echo "PR_FETCH_DEPTH=$(( ${{ github.event.pull_request.commits }} + 1 ))" >> "${GITHUB_ENV}"
+
       - name: Checkout repository
         uses: actions/checkout@v4
+        with:
+          fetch-depth: ${{ env.PR_FETCH_DEPTH }}
+
+      - name: Check for agnocast_kmod changes
+        id: check_changes
+        run: |
+          set -e
+          if git diff --name-only ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha }} | grep '^agnocast_kmod/'; then
+            echo "kmod_changed=true" >> $GITHUB_OUTPUT
+          else
+            echo "kmod_changed=false" >> $GITHUB_OUTPUT
+          fi
 
       - name: Install sparse
+        if: steps.check_changes.outputs.kmod_changed == 'true'
         run: |
           sudo apt-get update
           sudo apt-get install -y sparse
 
       - name: Run sparse
+        if: steps.check_changes.outputs.kmod_changed == 'true'
         run: |
           cd agnocast_kmod
           make sparse
--- a/agnocast_kmod/Makefile
+++ b/agnocast_kmod/Makefile
@@ -54,7 +54,8 @@ test:
 	make -C /lib/modules/$(shell uname -r)/build M=$(PWD) KUNIT_BUILD=y modules
 
 sparse:
-	! make -C /lib/modules/$(shell uname -r)/build M=$(PWD) C=2 modules 2>&1 | grep 'agnocast_kmod/.*warning'
+	set -o pipefail; make -C /lib/modules/$(shell uname -r)/build M=$(PWD) C=2 modules 2>&1 | tee sparse.log
+	! grep 'agnocast_kmod/.*warning' sparse.log
 
 clean:
 	make -C /lib/modules/$(shell uname -r)/build M=$(PWD) clean
--- a/agnocast_kmod/Makefile
+++ b/agnocast_kmod/Makefile
@@ -54,8 +54,8 @@ test:
 	make -C /lib/modules/$(shell uname -r)/build M=$(PWD) KUNIT_BUILD=y modules
 
 sparse:
-	set -o pipefail; make -C /lib/modules/$(shell uname -r)/build M=$(PWD) C=2 modules 2>&1 | tee sparse.log
-	! grep 'agnocast_kmod/.*warning' sparse.log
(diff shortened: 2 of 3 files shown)


agnocast_kmod/Makefile
```diff
@@ -53,7 +53,10 @@ all:
 test:
 	make -C /lib/modules/$(shell uname -r)/build M=$(PWD) KUNIT_BUILD=y modules
 
+sparse:
+	! make -C /lib/modules/$(shell uname -r)/build M=$(PWD) C=2 modules 2>&1 | grep 'agnocast_kmod/.*warning'
```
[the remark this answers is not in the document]
Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1250/commits/ff20f8105031ab443647fa04e3af9b57e457b846.

.github/workflows/sparse-check.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+      - name: Check for agnocast_kmod changes
+        id: check_changes
+        run: |
+          set -e
+          if git diff --name-only ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha }} | grep '^agnocast_kmod/'; then
+            echo "kmod_changed=true" >> $GITHUB_OUTPUT
+          else
+            echo "kmod_changed=false" >> $GITHUB_OUTPUT
+          fi
```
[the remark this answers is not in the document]
Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1250/commits/ff20f8105031ab443647fa04e3af9b57e457b846.

agnocast_kmod/Makefile
```diff
@@ -53,7 +53,11 @@ all:
 test:
 	make -C /lib/modules/$(shell uname -r)/build M=$(PWD) KUNIT_BUILD=y modules
 
+sparse:
+	make -C /lib/modules/$(shell uname -r)/build M=$(PWD) C=2 modules > sparse.log 2>&1 || { cat sparse.log; exit 1; }
+	! grep 'agnocast_kmod/.*warning' sparse.log && rm -f sparse.log
+
```
kobayu858, line 59: When grep matches (i.e. warnings exist), `! grep `returns non-zero and the `&&` short-circuits, skipping rm. However, Make doesn't necessarily fail the recipe in this case — it depends on the exit code of the last command in the line, and the short-circuited && chain can end up returning 0 in some shells. This could silently pass CI even when there are warnings.

Nit — happy to defer if you've verified the current behavior works as expected in CI.

agnocast_kmod/Makefile
```diff
@@ -53,7 +53,11 @@ all:
 test:
 	make -C /lib/modules/$(shell uname -r)/build M=$(PWD) KUNIT_BUILD=y modules
 
+sparse:
+	make -C /lib/modules/$(shell uname -r)/build M=$(PWD) C=2 modules > sparse.log 2>&1 || { cat sparse.log; exit 1; }
+	! grep 'agnocast_kmod/.*warning' sparse.log && rm -f sparse.log
```
k1832, line 58: It seems that there is a flag that handles this? -> `CF="-Werror"`

So these lines can be replaced with:
```
make -C /lib/modules/$(shell uname -r)/build M=$(PWD) C=2 CF="-Werror" modules
```

.github/workflows/sparse-check.yaml
```diff
@@ -0,0 +1,37 @@
+name: sparse-check
+
+on:
+  pull_request:
+    types: [opened, synchronize]
```
k1832, line 5: There seems to be an option to run the workflow only when the PR contains `agnocast_kmod/` change(s) in the first place?
https://docs.github.com/ja/actions/reference/workflows-and-actions/events-that-trigger-workflows#running-your-workflow-only-when-a-push-affects-specific-files

```
on:
  pull_request:
    types: [opened, synchronize]
    paths:
      - 'agnocast_kmod/**'
```

This could remove the `Check for agnocast_kmod changes` and the `if` checks

k1832 (changes requested):
Only small comments. 

This is a great and proactive change. Thanks!

agnocast_kmod/Makefile
Koichi98: I've verified the current behavior on GNU Make with both bash and dash:                                                                                                                                                                                                                                                                                                                                                                   

- When grep matches (warning exists): recipe exits 1, Make fails the target (Error 1), sparse.log is preserved for debugging. Tested with simulated warning lines.

- When grep doesn't match: recipe exits 0, sparse.log is removed.

agnocast_kmod/Makefile
Koichi98: I got the following with `CF="-Werror"` option.

```
 make -C /lib/modules/$(uname -r)/build M=$(pwd) C=2 CF="-Werror" modules
make: Entering directory '/usr/src/linux-headers-6.8.1+'
warning: the compiler differs from the one used to build the kernel
  The kernel was built by: gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
  You are using:           gcc (Ubuntu 11.4.0-1ubuntu1~22.04.3) 11.4.0
  CHECK   /home/koichiimai/Agnocast/agnocast/agnocast_kmod/agnocast_main.c
/home/koichiimai/Agnocast/agnocast/agnocast_kmod/agnocast_main.c: note: in included file (through include/linux/bitops.h, include/linux/radix-tree.h, include/linux/idr.h, include/linux/ipc_namespace.h, /home/koichiimai/Agnocast/agnocast/agnocast_kmod/agnocast.h, ...):
./arch/x86/include/asm/bitops.h:66:1: warning: unreplaced symbol 'return'
/home/koichiimai/Agnocast/agnocast/agnocast_kmod/agnocast_main.c: note: in included file (through include/linux/bitops.h, include/linux/radix-tree.h, include/linux/idr.h, include/linux/ipc_namespace.h, /home/koichiimai/Agnocast/agnocast/agnocast_kmod/agnocast.h, ...):
./include/asm-generic/bitops/generic-non-atomic.h:30:9: warning: unreplaced symbol 'mask'
./include/asm-generic/bitops/generic-non-atomic.h:31:9: warning: unreplaced symbol 'p'

```

There might be two problems.
1.  `C=2 CF="-Werror"` still exits 0 when sparse warns
2. even if it did, sparse emits unavoidable `unreplaced symbol` warnings from kernel headers (`include/asm-generic/bitops/*.h`) whenever we include `<linux/bitops.h>`, which would fail CI unconditionally. The `grep               
  'agnocast_kmod/.*warning'` filter is there specifically to skip that kernel-header noise.

.github/workflows/sparse-check.yaml
Koichi98: Thanks for the request!
I actually considered the same approach, but ran into this: if we later make sparse-check a `required` status check, this fix causes workflow to be skipped for PRs without `agnocast_kmod/` changes, which leaves the check stuck in "Pending" and blocks merging.

Reference: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks
> If a workflow is skipped due to [path filtering](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore), [branch filtering](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore) or a [commit message](https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs), then checks associated with that workflow will remain in a "Pending" state. A pull request that requires those checks to be successful will be blocked from merging.

kobayu858 (approved):

agnocast_kmod/Makefile
k1832: Ah that makes sense. Then I think the current change is fine as-is!

.github/workflows/sparse-check.yaml
k1832: Thanks for pointing that out! Understood

k1832 (approved):
LGTM

Issue #969, opened 2026-01-28
Refactor epoll event handling and flag management


Koichi98:
wait_and_handle_epoll_event currently handles multiple concerns (timer/subscription/shutdown). Refactor to improve separation of concerns: try to delete timer/subscription/shutdown specific operations from agnocast_epoll.hpp,cpp as much as possible. Also, event flags (SHUTDOWN_EVENT_FLAG, TIMER_EVENT_FLAG, etc.) are defined as separate constants. Consider organizing them into a more structured design (e.g., enum, dedicated header, or flag management class).

ruth561:
To resolve this issue, the following changes are required:
1. Remove `need_epoll_updates` and break the dependency between `id2_timer_info` and `id2_callback_info`.
2. Separate the handling for each event managed by epoll and move the implementations into separate files.

PR #1271, opened 2026-04-20
refactor(agnocastlib): remove the global variable `need_epoll_updates`

ruth561:
## Description

This PR removes the `need_epoll_updates` global variable and introduces an alternative mechanism that sends notifications to each Executor individually.

### Motivation
The global variable `need_epoll_updates` was a boolean that evaluated to true when there were elements marked as `need_epoll_update` in either `id2_callback_info` or `id2_timer_info`, and false otherwise. This variable created an unnecessary lock dependency between `id2_callback_info` and `id2_timer_info`. I needed to refactor this because this lock dependency would make it difficult to add new types of epoll events in the future.

### Solution
I introduced a new mechanism called `EpollUpdateTracker` to send notifications to each Executor individually. It holds an internal boolean variable, and setting this variable to true allows us to send a notification to the respective Executor.

Previously, all Executors had to poll the global `need_epoll_updates` variable to check if update operations were necessary. With this change, each Executor now only needs to check its own state once when an update is actually required.

### Future Work
By leveraging this per-Executor notification network, it is now possible to implement wake-up processes for the Executors in the future. Furthermore, since the data structure dependencies between `id2_callback_info` and `id2_timer_info` have been eliminated, we can now proceed with the epoll refactoring work (#969). Please refer to the linked Issue description for more details on the upcoming workflow.

## Related links

#969 

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

**Important notes:**


src/agnocastlib/include/agnocast/agnocast_executor.hpp
```diff
@@ -41,6 +41,8 @@ class AgnocastExecutor : public rclcpp::Executor
   bool get_next_agnocast_executable(AgnocastExecutable & agnocast_executable, const int timeout_ms);
   static void execute_agnocast_executable(AgnocastExecutable & agnocast_executable);
 
+  EpollUpdateTracker epoll_update_tracker_;
```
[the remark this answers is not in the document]
ruth561: Fixed in [7469d09](https://github.com/autowarefoundation/agnocast/pull/1271/commits/7469d09603296978160d6e63171f43791b09fb09).

src/agnocastlib/include/agnocast/agnocast_epoll_update_dispatcher.hpp
```diff
@@ -0,0 +1,72 @@
+#pragma once
+
+#include "sys/epoll.h"
+
+#include <atomic>
```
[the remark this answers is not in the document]
ruth561: Fixed in [7469d09](https://github.com/autowarefoundation/agnocast/pull/1271/commits/7469d09603296978160d6e63171f43791b09fb09).

src/agnocastlib/test/integration/test_agnocast_epoll_update.cpp
```diff
@@ -0,0 +1,329 @@
...
+#include <agnocast_cie_config_msgs/msg/callback_group_info.hpp>
+
+#include <gtest/gtest.h>
+
+#include <atomic>
+#include <memory>
+#include <thread>
+
+using namespace std::chrono_literals;
+
```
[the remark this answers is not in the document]
ruth561: Fixed in [7469d09](https://github.com/autowarefoundation/agnocast/pull/1271/commits/7469d09603296978160d6e63171f43791b09fb09).

src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
```diff
@@ -33,6 +34,9 @@ class AgnocastOnlyExecutor
   int shutdown_event_fd_;
   pid_t my_pid_;
 
+  EpollUpdateTracker epoll_update_tracker_;
+  uint64_t epoll_update_tracker_id_;
```
Koichi98: IMO `epoll_update_tracker_id_` seems redundant — it only duplicates `epoll_update_tracker_.id()` and invites the question of whether the two could ever diverge. Just calling `epoll_update_tracker_.id()` at the use sites should be enough.

src/agnocastlib/include/agnocast/agnocast_epoll_update_dispatcher.hpp
```diff
@@ -0,0 +1,71 @@
...
+private:
+  friend class EpollUpdateDispatcher;
+
+  EpollUpdateTracker(uint64_t tracker_id, std::shared_ptr<TrackerContext> context)
+  : id_(tracker_id), context_(std::move(context))
+  {
+  }
+
+  uint64_t id_;
+  std::shared_ptr<TrackerContext> context_;
```
Koichi98, line 68: nit: since TrackerContext only holds a single atomic<bool>, can't this just be shared_ptr<std::atomic<bool>>? Unless we expect to add more per-tracker state later.

ruth561: Yes, that’s right. I will add more per-tracker state later. Specifically, an eventfd to wake up the Executor for epoll updates.

src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
ruth561: Good point. I agree, so I fixed it. Please check commit [efa7d07](https://github.com/autowarefoundation/agnocast/pull/1271/commits/efa7d07be7b9246d830f48ddc07ec41ea956b3bd).

src/agnocastlib/include/agnocast/agnocast_epoll_update_dispatcher.hpp
Koichi98: Ah, that makes sense

Koichi98 (approved):

Koichi98:
@ruth561 
The new epoll update tests caught a real bug on `MultiThreadedAgnocastExecutor`: `SpinAddCbgTimer` and `AddSpinCbgTimer` fail because a callback group created after `add_node()` never reaches rclcpp's `weak_groups_to_nodes_` map.

One direction that seems to work: call `rclcpp::Executor::add_callback_groups_from_nodes_associated_to_executor()` from `AgnocastExecutor::prepare_epoll()` under `mutex_`. Does this approach sound reasonable, or do you see a better one?                                                                                                                                                                      
                                                                                                                                                                                                               
I'd like to get this resolved before this PR lands — I'll leave it to you whether to fold the fix into this PR or split it into a follow-up.

Thank you for adding the tests! It helps us a lot.

```
6: [ RUN      ] EpollUpdateTest/1.SpinAddCbgTimer
[Processing: agnocastlib]                           
6: /home/koichiimai/Agnocast/agnocast_latest/src/agnocastlib/test/integration/test_agnocast_epoll_update.cpp:199: Failure
6: Value of: success
6:   Actual: false
6: Expected: true
6: Timer callback was not called, epoll update might have failed
6: [ERROR] [1776734105.432921098] [Agnocast]: Agnocast internal implementation error: timer info cannot be found
6: -- run_test.py: return code 1
6: -- run_test.py: generate result file '/home/koichiimai/Agnocast/agnocast_latest/build/agnocastlib/test_results/agnocastlib/test_integration_agnocast_epoll_update_agnocastlib.gtest.xml' with failed test
6: -- run_test.py: verify result file '/home/koichiimai/Agnocast/agnocast_latest/build/agnocastlib/test_results/agnocastlib/test_integration_agnocast_epoll_update_agnocastlib.gtest.xml'
6/7 Test #6: test_integration_agnocast_epoll_update_agnocastlib ......................***Failed   11.42 sec

```

Koichi98 (changes requested):
https://github.com/autowarefoundation/agnocast/pull/1271#issuecomment-4285313470

ruth561:
@Koichi98 
Thank you so much for reporting the failing test cases and identifying the root cause! I completely agree with your analysis of why the failure is occurring.

I have created a proposed fix in commit [63eb94f](https://github.com/autowarefoundation/agnocast/pull/1271/commits/63eb94f8c1e809e5f80e13a3ce9ec61c59d52e72), so please take a look when you have a chance. While I agree that your suggestion to call it from within `prepare_epoll()` would also successfully fix the issue, I decided to call it from within `get_next_ready_agnocast_executable()` instead. My intention here is to keep the execution flow as consistent as possible with the jazzy version.

If you spot any problems with these changes or have any concerns, please let me know!

Koichi98 (approved):

PR #1275, opened 2026-04-21
refactor(agnocastlib): remove event-specific logic from `agnocast_epoll.cpp` and `.hpp`

### LINKED ISSUE #969: Refactor epoll event handling and flag management

ruth561:
## Description

This PR refactors `agnocast_epoll.cpp` and `.hpp` to improve code maintainability and make it easier to add new events in the future.

### Problem

Previously, the event-specific logic for handling epoll events was written directly inside `agnocast_epoll.cpp` and `.hpp`. This caused high code dependency. Every time a new event was added, the dependency grew, making the code harder to manage.

### Changes

- Removed event-specific processing from `agnocast_epoll.cpp` and `.hpp`.
- Moved the logic for each event into its own separate source file.
- Introduced `EpollEventHandler`, an abstract base class. The logic for each event is now implemented in a derived class that inherits from this base class.
- Introduced `EpollManager`, a new class to manage epoll and call these event handlers.

## Related links

close #969 
#1271

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

**Important notes:**


src/agnocastlib/src/agnocast_callback_info.cpp
```diff
@@ -18,7 +22,7 @@ std::atomic<uint32_t> next_callback_info_id;
 uint32_t allocate_callback_info_id()
 {
   const uint32_t callback_info_id = next_callback_info_id.fetch_add(1);
-  if ((callback_info_id & EPOLL_EVENT_ID_RESERVED_MASK) != 0U) {
+  if (callback_info_id >= MAX_CALLBACK_INFO_ID) {
     throw std::runtime_error("Callback info ID overflow: too many callbacks registered");
   }
   return callback_info_id;
```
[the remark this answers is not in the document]
ruth561: The issue already exists in the original implementation. Since this PR focuses on refactoring, bug fixes are out of scope and should not be included in this diff.

src/agnocastlib/src/agnocast_timer_info.cpp
```diff
@@ -191,7 +196,7 @@ int create_timer_fd(uint32_t timer_id, std::chrono::nanoseconds period, rcl_cloc
 uint32_t allocate_timer_id()
 {
   const uint32_t timer_id = next_timer_id.fetch_add(1);
-  if ((timer_id & EPOLL_EVENT_ID_RESERVED_MASK) != 0U) {
+  if (timer_id >= MAX_TIMER_ID) {
     throw std::runtime_error("Timer ID overflow: too many timers created");
   }
   return timer_id;
```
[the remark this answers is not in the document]
ruth561: The issue already exists in the original implementation. Since this PR focuses on refactoring, bug fixes are out of scope and should not be included in this diff.

src/agnocastlib/test/integration/test_agnocast_epoll_update.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+  void stop_spinning()
+  {
+    if (executor_) {
+      executor_->cancel();
+    }
+    if (spin_thread_.joinable()) {
+      spin_thread_.join();
+    }
+  }
```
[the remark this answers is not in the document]
ruth561: These changes are not included in this PR.

src/agnocastlib/src/agnocast_epoll.cpp
```diff
@@ -1,23 +1,77 @@
...
 #include <unistd.h>
 
+#include <cstdint>
+#include <vector>
+
 namespace agnocast
 {
 
-std::atomic<bool> need_epoll_updates{false};
+struct AgnocastExecutable;
```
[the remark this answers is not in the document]
ruth561: Fixed in [17198f4](https://github.com/autowarefoundation/agnocast/pull/1275/commits/17198f4ce2ee461afead857022cccc9596b6ab9e).

src/agnocastlib/include/agnocast/agnocast_epoll.hpp
```diff
@@ -1,119 +1,95 @@
...
-    }
+    (void)epoll_fd;
+    (void)validate_callback_group;
   }
 
-  // Register timers to epoll
+  bool handle(EpollEventLocalID /*event_local_id*/) override { return true; }
+};
+
+class DummyEventHandler : public EpollEventHandler
```
Koichi98, line 81: nit: Could you add a brief comment on `DummyEventHandler` and `EpollEventType::Dummy` explaining their purpose?  AFAIK it's a placeholder to fill the Shutdown slot in `AgnocastExecutor`, since `EventHandlerArray` requires every slot to be filled.

src/agnocastlib/src/agnocast_epoll.cpp
```diff
@@ -1,21 +1,73 @@
...
+}
+
+EpollManager::~EpollManager()
+{
+  if (epoll_fd_ >= 0) {
+    close(epoll_fd_);
+  }
+}
+
+int EpollManager::add_event(int fd, EpollEventType type, EpollEventLocalID local_id) const
```
Koichi98: This is currently only used for shutdown event. Could this also be used by the other handlers as well? Since `EpollManager` owns `epoll_fd`, it seems natural for all `epoll_ctl` calls to go through it. There may be dependency issues that make this difficult, but worth considering.

src/agnocastlib/src/agnocast_epoll.cpp
```diff
@@ -1,21 +1,73 @@
...
+{
+  struct epoll_event ev = {};
+  ev.events = EPOLLIN;
+  ev.data.u64 = pack_epoll_data(type, local_id);
+  return epoll_ctl(epoll_fd_, EPOLL_CTL_ADD, fd, &ev);
```
Koichi98: nit: Returning the raw `int` from `epoll_ctl` leaks the underlying API through the wrapper. Might be better to consider returning `bool` (or handling the error inside `add_event`) to keep the abstraction layer consistent.

src/agnocastlib/test/unit/test_agnocast_epoll_event.cpp
```diff
@@ -0,0 +1,51 @@
...
+namespace agnocast
+{
+
+TEST(EpollEventDataTest, PackAndUnpackRoundTripForAllKnownTypes)
+{
+  constexpr EpollEventLocalID id = 0x12345678;
+
+  const std::array<EpollEventType, static_cast<size_t>(EpollEventType::NrEventType)> types = {
+    EpollEventType::Subscription, EpollEventType::Timer, EpollEventType::Clock,
+    EpollEventType::Shutdown};
```
Koichi98, line 17: Could you add the following assertion?

```
static_assert(types.size() == static_cast<size_t>(EpollEventType::NrEventType), "Update this test when adding a new EpollEventType");
```

src/agnocastlib/test/unit/test_agnocast_epoll_event.cpp
```diff
@@ -0,0 +1,51 @@
...
+  constexpr EpollEventLocalID id = MAX_CALLBACK_INFO_ID - 1;
+
+  const auto packed = pack_epoll_data(type, id);
+  const auto [unpacked_type, unpacked_id] = unpack_epoll_data(packed);
+
+  EXPECT_EQ(unpacked_type, type);
+  EXPECT_EQ(unpacked_id, id);
+}
+
+}  // namespace agnocast
```
Koichi98, line 54: Could you add more tests to check the behavior of `EpollManager`? Might be better to define in another file.

src/agnocastlib/src/agnocast_epoll.cpp
ruth561: Thanks for the suggestion. I've updated the code in [46b6054](https://github.com/autowarefoundation/agnocast/pull/1275/commits/46b6054eea4637dedcbdf2207eeee14a0b2dcf2f).

src/agnocastlib/include/agnocast/agnocast_epoll.hpp
ruth561: I've added comments for all new classes in [7893d16](https://github.com/autowarefoundation/agnocast/pull/1275/commits/7893d163a6515088d0cbc2f0791ea828830d5e14). Please take a look.

src/agnocastlib/test/unit/test_agnocast_epoll_event.cpp
ruth561: Thanks! I've added the assertion in [c18712d](https://github.com/autowarefoundation/agnocast/pull/1275/commits/c18712d902e40f77495ab37e6fa0199278af63e1).

src/agnocastlib/src/agnocast_epoll.cpp
ruth561: Thank you for the thoughtful review! I completely agree that it would be better to avoid invoking the raw `epoll_ctl` API everywhere.

However, I have some architectural concerns about accessing `EpollManager` directly from `EpollEventHandler`:

**Dependency Issues**: `EpollManager` is designed as a facade class used by the `Executor` to bundle various components and make them easier to use. Its role is to oversee and manage the `EpollEventHandler`s.

**Coupling**: If we pass a reference of `EpollManager` down to the managed `EpollEventHandler`, it would create an unnecessary bidirectional dependency.

### Future Architecture
Currently, we don't have an API to remove events from epoll, which is why there isn't a dedicated class to manage the set of registered events. When we eventually need a remove API, we will need a component to manage these events. The ideal design would be for `EpollManager` to own this new event-managing component, and then pass it as an argument to `EpollEventHandler::prepare_epoll()`.

### Current PR Scope
For this PR, my goal was to keep the changes to the original implementation as minimal as possible. Because of this, I intentionally avoided introducing a new epoll abstraction class and used `epoll_ctl` directly.

If you feel it is necessary to introduce this abstraction and address the dependency structure in this PR, I would be happy to propose a fix. I'd love to hear your thoughts on how you'd like to proceed!

Koichi98: I see. Future work is fine. Thanks!

src/agnocastlib/test/unit/test_agnocast_epoll_event.cpp
ruth561: I've added unit tests for `EpollManager` in [8df96fc](https://github.com/autowarefoundation/agnocast/pull/1275/commits/8df96fc69fdeef337c99e0587c857b7ec0a7eaad). Please take a look!

Koichi98 (approved):

src/agnocastlib/test/unit/test_agnocast_epoll_event.cpp
Koichi98: Great, thanks!

Issue #1019, opened 2026-02-09
introduce shutdown/ok API for agnocast::Node


Koichi98:
As in title. Also, the current atomic operations in signal handler might not be async signal safe so we also should fix this.

ruth561:
### Discussion on PR scope for `agnocast::shutdown()` / `ok()`  implementation and signal handling

To implement these APIs, I am planning to add an internal state to `Context`.

When a `SIGINT` or `SIGTERM` is received, we need to access this internal state. However, doing this directly from a traditional signal handler is impossible because it requires acquiring the lock for `g_context` (which is not async-signal-safe).

### Proposed Solution
To resolve this, I propose creating a dedicated thread specifically for signal handling. When a signal occurs, the actual processing and interaction with `g_context` will be safely delegated to this thread.

### Alternative Considered
I also considered making the state variable inside `Context` atomic so it could be accessed directly from the signal handler. I decided against this approach because it would require external access to `Context`'s private fields, breaking encapsulation and leading to poor design.

### Discussion: PR Scope
I am opening this issue to discuss the scope of the upcoming PR. If you feel that introducing a new signal-handling thread is too large of a change for this specific issue, please let me know your thoughts. I am open to feedback if you think the thread creation should be handled in a separate PR, or if there's an alternative approach you'd prefer.

PR #1283, opened 2026-04-27
refactor(agnocastlib): move signal handling to a dedicated thread

ruth561:
### Background

Signal handlers are highly restricted (e.g., they cannot acquire locks), which limits our ability to implement new features. Since we now need to lock global variables when a signal occurs, we are moving signal handling to a dedicated thread to allow for more complex operations.

### Changes

- [x] Created a dedicated thread for signal handling and moved the logic there.
- [x] Added an `uninstall()` function to `SignalHandler` for proper cleanup. This allows the handler to be installed and uninstalled multiple times.
- [x] Added unit tests.

## Related links

#1019 

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

### Thread Management and Intentional Leak

`SignalHandler::install()` and `SignalHandler::uninstall()` are intended to be called within agnocast::init() and `agnocast::shutdown()`. Although `uninstall()` is expected to be called explicitly before the process ends, the `std::thread` is managed with a raw pointer and intentionally leaked. This prevents the `std::thread` destructor from running at program exit, which could cause an abort if the thread is still active.

### Regarding `signalfd`

The use of `signalfd` to handle signals was considered but deemed inappropriate for this library's specific use case. Please refer to the comments in the test files for more technical details on this decision.

**Important notes:**


src/agnocastlib/src/node/agnocast_signal_handler.cpp
```diff
@@ -15,44 +20,148 @@ namespace
...
     fd.store(-1);
   }
+  eventfd_count_.store(0);
+
+  const int signal_eventfd = eventfd(0, EFD_CLOEXEC);
+  if (signal_eventfd < 0) {
+    RCLCPP_ERROR(logger, "Failed to create signal eventfd: %s", strerror(errno));
+    exit(EXIT_FAILURE);
+  }
+  signal_eventfd_.store(signal_eventfd);
```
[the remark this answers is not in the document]
ruth561: Regarding this point, a write() to an eventfd will only block if the internal 64-bit counter overflows (i.e., exceeds 0xfffffffffffffffe). Since it is practically impossible for the counter to reach this saturation point in our use case, write() will never block. Therefore, I believe no change is necessary here.

Please see the document: https://man7.org/linux/man-pages/man2/eventfd.2.html

src/agnocastlib/src/node/agnocast_signal_handler.cpp
```diff
@@ -15,44 +20,148 @@ namespace
...
+  int signal_eventfd = signal_eventfd_.exchange(-1);
+  if (signal_eventfd == -1) {
+    RCLCPP_ERROR(logger, "Signal eventfd was already closed");
+    exit(EXIT_FAILURE);
+  }
+
+  // Wait for handlers that were already in-flight between the first
+  // handler_inflight_count_ wait and the exchange above.  Those handlers may have
+  // already loaded the valid fd (before our exchange) and must complete their write
+  // before we close it.
```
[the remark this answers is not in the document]
ruth561: Fixed in [05fe147](https://github.com/autowarefoundation/agnocast/pull/1283/commits/05fe147fbfa73ea89b9b8d4b01bce34bd1f6f731).

src/agnocastlib/src/node/agnocast_signal_handler.cpp
```diff
@@ -80,10 +193,57 @@ void SignalHandler::unregister_shutdown_event(int eventfd)
...
+    }
+  }
+}
+
+void SignalHandler::wait_for_signal_eventfd()
+{
+  const int fd = signal_eventfd_.load();
+  if (fd != -1) {
+    uint64_t count = 0;
+    [[maybe_unused]] auto ret = read(fd, &count, sizeof(count));
```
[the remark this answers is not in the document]
ruth561: Fixed in [05fe147](https://github.com/autowarefoundation/agnocast/pull/1283/commits/05fe147fbfa73ea89b9b8d4b01bce34bd1f6f731).

src/agnocastlib/test/unit/test_agnocast_signal_handler.cpp
```diff
@@ -0,0 +1,301 @@
+#include "../../src/node/agnocast_signal_handler.hpp"
+
+#include <gtest/gtest.h>
+#include <sys/epoll.h>
+#include <sys/eventfd.h>
+#include <unistd.h>
+
+#include <algorithm>
```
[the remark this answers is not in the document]
ruth561: Fixed in [05fe147](https://github.com/autowarefoundation/agnocast/pull/1283/commits/05fe147fbfa73ea89b9b8d4b01bce34bd1f6f731).

src/agnocastlib/src/node/agnocast_signal_handler.cpp
```diff
@@ -1,11 +1,16 @@
 #include "agnocast_signal_handler.hpp"
 
+#include "agnocast/node/agnocast_context.hpp"
```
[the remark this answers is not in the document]
ruth561: Fixed in [05fe147](https://github.com/autowarefoundation/agnocast/pull/1283/commits/05fe147fbfa73ea89b9b8d4b01bce34bd1f6f731).

src/agnocastlib/src/node/agnocast_signal_handler.cpp
```diff
@@ -15,44 +20,148 @@ namespace
...
+  if (!installed_) {
+    return;
+  }
+
+  if (sigaction(SIGINT, &old_sigint_action_, nullptr) != 0) {
+    RCLCPP_ERROR(logger, "Failed to restore SIGINT handler: %s", strerror(errno));
+  }
+  if (sigaction(SIGTERM, &old_sigterm_action_, nullptr) != 0) {
+    RCLCPP_ERROR(logger, "Failed to restore SIGTERM handler: %s", strerror(errno));
+  }
```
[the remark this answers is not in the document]
ruth561: Fixed in [5fdbbb9](https://github.com/autowarefoundation/agnocast/pull/1283/commits/5fdbbb9b0a65ef409dae3661d33b3afbd56fd6c6).

src/agnocastlib/test/unit/test_agnocast_signal_handler.cpp
```diff
@@ -0,0 +1,301 @@
...
+
+class SignalHandlerTest : public ::testing::Test
+{
+protected:
+  void TearDown() override
+  {
+    for (int fd : owned_eventfds_) {
+      close(fd);
+    }
+  }
```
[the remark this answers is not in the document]
ruth561: Fixed in [05fe147](https://github.com/autowarefoundation/agnocast/pull/1283/commits/05fe147fbfa73ea89b9b8d4b01bce34bd1f6f731).

src/agnocastlib/src/node/agnocast_signal_handler.cpp
```diff
@@ -15,44 +19,157 @@ namespace
...
+    exit(EXIT_FAILURE);
+  }
+
+  // join signal_thread_
+  if (signal_thread_->joinable()) {
+    signal_thread_->join();
+  }
+  delete signal_thread_;  // NOLINT
+  signal_thread_ = nullptr;
+
```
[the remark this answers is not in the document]
ruth561: As suggested, I used finer-grained locking in `uninstall()` and moved `join()` outside the lock to reduce the risk of future deadlocks. Because of this change, I also added an internal state variable to `SignalHandler` to ensure `uninstall()` runs exactly once. Please see the commit [f17e778](https://github.com/autowarefoundation/agnocast/pull/1283/commits/f17e778d4a62c64d792ec039515768a3011de781).

src/agnocastlib/src/node/agnocast_signal_handler.cpp
```diff
@@ -80,10 +201,73 @@ void SignalHandler::unregister_shutdown_event(int eventfd)
   }
 }
 
+void SignalHandler::signal_processing_loop()
+{
+  while (true) {
+    if (signal_received_.exchange(false)) {
+      notify_all_executors();
```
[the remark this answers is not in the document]
ruth561: I have updated `notify_all_executors()` to properly handle `write()` errors. The code now retries the write operation if `EINTR` is returned to prevent missed notifications.

src/agnocastlib/src/node/agnocast_signal_handler.cpp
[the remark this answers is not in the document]
Koichi98: Agreed. Also, since `EFD_NONBLOCK` would make `read()` non-blocking too and force us to add an epoll/poll layer in the worker, I'd rather keep it as-is.

src/agnocastlib/src/node/agnocast_signal_handler.cpp
```diff
@@ -15,84 +19,283 @@ namespace
...
+  }
 }
 
 void SignalHandler::register_shutdown_event(int eventfd)
 {
-  std::lock_guard<std::mutex> lock(eventfds_mutex_);
-  size_t count = eventfd_count_.load();
+  std::lock_guard<std::mutex> lock(mutex_);
+  if (state_ != State::Installed) {
+    return;
```
Koichi98: Might be better to add WARNING here and also the corresponding part of `unregister_shutdown_event`?

ruth561: Thank you for the review and the suggestion!

I actually added WARNING logs there initially. However, I noticed it generated noisy warnings during a standard shutdown sequence. For example, in the following code:

```cpp
int main(int argc, char * argv[])
{
  agnocast::init(argc, argv);
  agnocast::SingleThreadedAgnocastOnlyExecutor executor;
  // ...
  agnocast::shutdown(); // SignalHandler::uninstall() is called here
  return 0;
} // The executor's destructor calls unregister_shutdown_event() here. 
  // Since the state is already NotInstalled, it outputs a warning.
```

Calling `unregister_shutdown_event()` when the state is already `Uninstalled` is safe because the event will never be used again and will be automatically cleared during the next `SignalHandler::install()`.
Therefore, to avoid this noise and keep things consistent, I decided to remove the warnings from both `register_shutdown_event()` and `unregister_shutdown_event()`.

### Proposal:

I agree that it is better to output a warning in `register_shutdown_event()`. On the other hand, to prevent noisy logs, I think we should avoid adding it to `unregister_shutdown_event()`.

So, in summary, how about we add the WARNING log only to register_shutdown_event()? This would keep the shutdown process quiet while still catching registration issues.

Please let me know what you think!
Thanks!

src/agnocastlib/test/unit/test_agnocast_signal_handler.cpp
```diff
@@ -0,0 +1,301 @@
...
+
+  epoll_event triggered_event{};
+  const int wait_result = epoll_wait(epoll_fd, &triggered_event, 1, timeout_ms);
+  close(epoll_fd);
+  if (wait_result <= 0) {
+    return false;
+  }
+
+  const ssize_t bytes_read = read(fd, &value, sizeof(value));
+  return bytes_read == static_cast<ssize_t>(sizeof(value));
```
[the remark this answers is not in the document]
ruth561: Fixed in [49523ba](https://github.com/autowarefoundation/agnocast/pull/1283/commits/49523ba6d1dae08160d2c49c00f90b8623ff26c4).

src/agnocastlib/src/node/agnocast_signal_handler.cpp
```diff
@@ -15,84 +19,283 @@ namespace
...
+  notify_signal_eventfd();
+  handler_inflight_count_.fetch_sub(1);
+
+  errno = saved_errno;
 }
 
 void SignalHandler::notify_all_executors()
 {
+  std::lock_guard<std::mutex> lock(mutex_);
   uint64_t val = 1;
```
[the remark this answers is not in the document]
ruth561: Updated the PR description.

src/agnocastlib/test/unit/test_agnocast_signal_handler.cpp
```diff
@@ -0,0 +1,301 @@
+#include "../../src/node/agnocast_signal_handler.hpp"
```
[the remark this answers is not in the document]
ruth561: Fixed in [3e3a965](https://github.com/autowarefoundation/agnocast/pull/1283/commits/3e3a9657d0f2ba706ebd6fd62c5edc3d4e890150).

src/agnocastlib/src/node/agnocast_signal_handler.cpp
Koichi98: That make sense. Please go ahead with it!

ruth561: Added warning in `register_shutdown_event()`. Please see [7ffe9d8](https://github.com/autowarefoundation/agnocast/pull/1283/commits/7ffe9d8bb52bc527c56f51cf6f66df21410bd68b)!

src/agnocastlib/test/unit/test_agnocast_signal_handler.cpp
```diff
@@ -0,0 +1,341 @@
...
+
+  send_sigint();
+  waiter.join();
+
+  EXPECT_TRUE(notified.load());
+
+  agnocast::SignalHandler::uninstall();
+}
+
+TEST_F(SignalHandlerTest, SigintAfterDelayNotifiesRegisteredEventfd)
```
Koichi98: This test looks nearly identical to `SigintNotifiesRegisteredEventfdViaWorkerThread` apart from the 300ms sleep. Is there a distinct scenario being verified here? If so, could you add a comment explaining the intent? If not, I'd suggest folding it into `SigintNotifiesRegisteredEventfdViaWorkerThread` and removing this test.

src/agnocastlib/test/unit/test_agnocast_signal_handler.cpp
```diff
@@ -0,0 +1,341 @@
...
+  uint64_t value = 0;
+  ASSERT_TRUE(read_event_fd_with_timeout(fd, TIMEOUT_MS_DEFAULT, value));
+
+  agnocast::SignalHandler::uninstall();
+
+  keep_running.store(false);
+  for (auto & t : pre_spawn_threads) {
+    t.join();
+  }
+}
```
Koichi98, line 343: I also want some tests to check the behavior of error/warning cases. Could you do that?

src/agnocastlib/test/unit/test_agnocast_signal_handler.cpp
```diff
@@ -0,0 +1,341 @@
...
+// This requires either calling pthread_sigmask(SIG_BLOCK, ...) in every thread, or calling it in
+// the main thread before spawning any threads (child threads inherit the signal mask).
+// A library cannot impose either requirement on its callers, so this test verifies that
+// the signal handler works correctly even when threads are spawned before install() is called.
+// This test is not strictly necessary for the current sigaction + eventfd implementation,
+// but is kept as a regression guard against future internal implementation changes.
+TEST_F(SignalHandlerTest, SigintNotifiesEventfdEvenWhenThreadsSpawnedBeforeInstall)
+{
+  // Spawn several threads before calling SignalHandler::install()
+  constexpr int pre_spawn_thread_count = 4;
```
Koichi98: We might have to increate this to a larger number so that Linux will likely choose these threads.

src/agnocastlib/test/unit/test_agnocast_signal_handler.cpp
Koichi98, line 343: Could you add boundary tests for `MAX_EXECUTORS_NUM`? Specifically two cases: (1) registering beyond the limit is rejected as expected, and (2) a slot freed by `unregister_shutdown_event` is correctly reused by a subsequent `register_shutdown_event`.

src/agnocastlib/src/node/agnocast_signal_handler.cpp
Koichi98: Great. Thanks!

src/agnocastlib/test/unit/test_agnocast_signal_handler.cpp
ruth561: Great catch, thank you! I've made the suggested change to fold this into the other test. Please check the commit [2795e55](https://github.com/autowarefoundation/agnocast/pull/1283/commits/2795e55e801ff3ef4a5a298630d1db6098c4a99a) when you have a moment.

src/agnocastlib/test/unit/test_agnocast_signal_handler.cpp
ruth561: Thanks! I think so too. I increased this number to 30. Please check the commit [59f27f8](https://github.com/autowarefoundation/agnocast/pull/1283/commits/59f27f87048a7c1ecc2b8d0e9bfce48a3302a689).

src/agnocastlib/test/unit/test_agnocast_signal_handler.cpp
ruth561: Thank you for your questions and suggestions!

I have a fundamental question: do we still need `MAX_EXECUTORS_NUM`? Since array access now occurs outside the signal handler, we could potentially use a data structure like `std::set<T>` for `eventfds_` and avoid the boundary check.

Does this macro serve any other purpose? If not, I would like to propose switching to a set implementation. This would remove the upper limit and eliminate the need for boundary check tests and slot reuse tests. I would love to hear your thoughts on this.

Koichi98: Ah, true. We just don't need `MAX_EXECUTORS_NUM`. Please just delete it by using some other data structures. This doesn't have to be in this PR (follow-up PR can be fine). I will leave it on you.

src/agnocastlib/src/node/agnocast_signal_handler.cpp
```diff
@@ -15,84 +19,285 @@ namespace
...
+  if (signal_eventfd < 0) {
+    RCLCPP_ERROR(logger, "Failed to create signal eventfd: %s", strerror(errno));
+    exit(EXIT_FAILURE);
   }
+  signal_eventfd_.store(signal_eventfd);
+
+  stop_requested_.store(false);
+  signal_received_.store(false);
+
+  signal_thread_ = new std::thread(&SignalHandler::signal_processing_loop);  // NOLINT
```
Koichi98, line 71: @atsushi421 

As part of the `agnocast::SignalHandler` refactor, a long-lived thread will be spawned (one per process) on the first `SignalHandler::install()` call. 

**What it does:** sleeps in `read()` on an internal eventfd; on `SIGINT`/`SIGTERM`, the signal handler wakes it, and it notifies each registered executor's `shutdown_event_fd_` so `spin()` exits.                                                                                                                                                                                                                                        
                             
  **Why:** signal handlers are restricted to async-signal-safe operations. For upcoming `agnocast::shutdown()` / `ok()`, the handler will need to touch internal state like `Context`, so we want to keep the handler minimal.                                                                                                                             
                             
 **Concern:** the thread is idle at runtime but wakes on signal. Since it's a plain `std::thread`, it's outside CIE's management — in MRM scenarios it could run in parallel with degraded-mode operation that requires controlled thread priorities.

I would like to have some comments from you about this.

Koichi98: @atsushi421 
Options I've considered:

- Might be possible to bring this thread under CIE's management one way or another — for example by having the signal-handler thread itself publish its tid to `/agnocast_cie_thread_configurator/callback_group_info`, or by adding an API on the CIE side so external threads can register themselves. 
-  Have the thread lower its own priority at startup.                                                                                                      
                      
We considered designs without a dedicated thread, but distributing post-signal `Context` handling across each executor's spin loop hurts maintainability. The dedicated-thread approach gives a much cleaner separation of responsibilities, so we'd like to go with it if possible.

src/agnocastlib/test/unit/test_agnocast_signal_handler.cpp
ruth561: Thanks! Since it only requires minor changes, I’ve included them in this PR.
I changed the implementation of `eventfds_` from `std::array` to `std::set` in [4cc2ffd](https://github.com/autowarefoundation/agnocast/pull/1283/commits/4cc2ffda7a6a61f968bb5475bce20e79639c0ca9).

Koichi98: Cool

src/agnocastlib/src/node/agnocast_signal_handler.cpp
atsushi421: Thanks for the thoughtful writeup. After thinking through the scheduling implications, I don't think this worker thread needs explicit scheduling-attribute management, and I'd lean toward leaving it outside CIE's management:

- **No explicit attribute setting is needed.** The thread is created via `std::thread` without attributes, so it inherits from the caller of `SignalHandler::install()`. In practice that's the main thread under `agnocast::init()`, which runs as `SCHED_OTHER`. The worker also spends almost all of its life blocked in `read()` on the eventfd and only wakes briefly on shutdown.
- **As long as the worker stays on `SCHED_OTHER`, it cannot impact MRM-class real-time tasks.** On Linux, `SCHED_FIFO` always preempts `SCHED_OTHER`, so the worker can only run when no MRM thread is runnable — it cannot delay degraded-mode operation, and at most it consumes a small slice of idle CPU cycles when a shutdown signal actually arrives.

Given that, I think the current design is fine as-is.

Koichi98: @atsushi421 
Thanks, that makes sense. If the worker is guaranteed to stay on `SCHED_OTHER`, I agree it won't interfere with MRM real-time tasks, so I'm fine leaving it as-is.

One thing I'd flag (more as a future consideration than a blocker): having a `SCHED_OTHER`-pinned thread that CIE doesn't know about, in theory, does constrain what scheduling policies CIE can apply later. In practice this is unlikely to matter, so I'm happy to keep the current design.

src/agnocastlib/test/unit/test_agnocast_signal_handler.cpp
ruth561: I updated the register_shutdown_event function signature to improve error handling and added several test cases. For more details, please see [0320453](https://github.com/autowarefoundation/agnocast/pull/1283/commits/0320453f0cf49b7e600c4482e9ca582e435e389a).

Koichi98 (approved):

src/agnocastlib/test/unit/test_agnocast_signal_handler.cpp
Koichi98: Thanks! I've checked that the tests are added correctly.

PR #1288, opened 2026-04-28
refactor(cie_thread_configurator): split callback groups and atomicize shared state

atsushi421:
## Description

Prepares `ThreadConfiguratorNode` and `PrerunNode` for a future switch to `MultiThreadedExecutor`. Behavior under the existing `SingleThreadedExecutor` is unchanged.

**Callback group split** — Both nodes now create one explicit `MutuallyExclusive` callback group on `this` for the `non_ros_thread_info` subscription. The default-domain `callback_group_info` subscription stays on the node's implicit default callback group (also `MutuallyExclusive`), mirroring the per-domain extra nodes (`nodes_for_each_domain_`), each of which hosts only one subscription on its default group. With this layout, a future `MultiThreadedExecutor` would parallelize: `non_ros_thread_callback` ⟷ `callback_group_callback`, and `callback_group_callback` between any two domains.

**`ThreadConfiguratorNode` synchronization** — Members reachable from both callback families are atomicized:
- `unapplied_num_` → `std::atomic<int>`
- `cgroup_num_` → `std::atomic<int>` (uses `fetch_add(relaxed)` for unique cgroup sequence numbers)
- `configured_at_least_once_` → `std::atomic<bool>` (used with `compare_exchange_strong` for one-shot success log)

The trailing block of both callbacks is rewritten as a single CAS that pairs the decrement and the one-shot trigger:
```cpp
if (!config->applied) {
  config->applied = true;
  if (unapplied_num_.fetch_sub(1, std::memory_order_acq_rel) == 1) {
    bool expected = false;
    if (configured_at_least_once_.compare_exchange_strong(expected, true)) {
      RCLCPP_INFO(this->get_logger(), "Success: All of the configurations are applied.");
    }
  }
}
```
The standalone `on_all_configured()` method is removed (folded inline so the bookkeeping and log fire atomically).

**`PrerunNode` synchronization** — `domain_and_cbg_ids_` (a `std::set` written from each domain's `topic_callback`) is guarded by a new `std::mutex`. The log line is moved outside the lock scope.

`non_ros_thread_names_`, `ThreadConfig::applied`, and `ThreadConfig::thread_id` stay non-atomic by design: each is reachable from exactly one `MutuallyExclusive` callback group, so no concurrent writers can exist. Inline comments on the relevant headers document this invariant.

**Out of scope** — `main.cpp` and `prerun_node_main.cpp` are intentionally unchanged. The executor remains `SingleThreadedExecutor`; this PR only positions the nodes so a future executor swap is safe without further refactoring.

## How was this PR tested?

- [ ] Autoware (required)
- [x] sample application

**Important notes:**


atsushi421:
Thanks for the review. The PR description was indeed stale — I dropped the dedicated callback group for the default-domain `callback_group_info` subscription in commit f3c73480 because the node's implicit default callback group is also `MutuallyExclusive`, so parallelization with `cbg_non_ros_thread_` under a future `MultiThreadedExecutor` is preserved either way. The PR description has now been updated to match the actual behavior.

kobayu858 (approved):

PR #1289, opened 2026-04-28
feat(bridge): add standard-mode R2A service bridge support

bdm-k:
## Description

This PR adds standard-mode support for R2A service bridges following #1274.

R2A service bridges in standard mode do not implement delegation logic that pubsub bridges use. An R2A service bridge is always created on the corresponding bridge manager of the target Agnocast service's process, so there is no need for additional mediation.

The changes in `agnocast_standard_bridge_manager.cpp` mirror the implementation of the performance bridge manager. In fact, I added two member functions, `create_service_bridge_if_needed()` and `check_and_remove_service_bridges()`, both of which are derived from member functions with the same names in the performance bridge manager.

The PR also renames pubsub bridge entities for clarity.

## How was this PR tested?

- [ ] Autoware (required)
- [x] sample application


src/agnocastlib/include/agnocast/bridge/agnocast_bridge_node.hpp
```diff
@@ -208,17 +208,66 @@ class AgnocastToRosBridge : public BridgeBase
...
+    ros_srv_ = parent_node->create_service<ServiceT>(
+      service_name,
+      [this](
+        const typename ServiceT::Request::SharedPtr ros_req,
+        typename ServiceT::Response::SharedPtr ros_res) {
+        auto agno_req = this->agno_client_->borrow_loaned_request();
+        *agno_req = *ros_req;
+
+        auto future = this->agno_client_->async_send_request(std::move(agno_req));
+        auto agno_res = future.get();
```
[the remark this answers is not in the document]
bdm-k: this is irrelevant

src/agnocastlib/src/bridge/agnocast_bridge_utils.cpp
```diff
@@ -212,4 +216,46 @@ rclcpp::QoS get_service_qos(const std::string & service_name)
...
+  }
+
+  snprintf(
+    static_cast<char *>(factory.shared_lib_path), SHARED_LIB_PATH_BUFFER_SIZE, "%s",
+    info.dli_fname);
+  snprintf(static_cast<char *>(factory.symbol_name), SYMBOL_NAME_BUFFER_SIZE, "%s", symbol_to_send);
+
+  auto base_addr = reinterpret_cast<uintptr_t>(info.dli_fbase);
+  factory.fn_offset = fn_current - base_addr;
+  factory.fn_offset_reverse = fn_reverse - base_addr;
```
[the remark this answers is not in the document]
bdm-k: fixed by setting `fn_reverse` to the same value as `fn_current`, which I think is safer than setting random addresses ([d41389a](https://github.com/autowarefoundation/agnocast/pull/1289/commits/d41389a3f805ea796731b777ad72cf6a9306ef3d)).

src/agnocastlib/include/agnocast/agnocast_client.hpp
```diff
@@ -25,6 +24,8 @@
 namespace agnocast
 {
 
+struct NoBridgeRequestPolicy;
```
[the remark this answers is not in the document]
bdm-k: It can't be helped

src/agnocastlib/src/bridge/standard/agnocast_standard_bridge_manager.cpp
```diff
@@ -383,22 +403,86 @@ bool StandardBridgeManager::should_remove_bridge(const std::string & topic_name,
...
+
+  if (req.direction != BridgeDirection::ROS2_TO_AGNOCAST) {
+    // A2R service bridge is not implemented yet.
+    return;
+  }
+
+  const std::string service_name = static_cast<const char *>(req.srv_target.service_name);
+  if (active_r2a_service_bridges_.count(service_name) != 0U) {
+    return;
+  }
```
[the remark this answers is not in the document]
bdm-k: This pattern is widely used across the codebase and the senders use `snprintf`, so this is fine.

src/agnocastlib/src/bridge/standard/agnocast_standard_bridge_manager.cpp
```diff
@@ -124,17 +130,26 @@ void StandardBridgeManager::on_signal()
...
+
   // Locally, unique keys include the direction. However, we register the raw topic name (without
   // direction) to the kernel to enforce single-process ownership for the entire topic.
-  const auto [topic_name, topic_name_with_direction] = extract_topic_info(req);
-  if (active_bridges_.count(topic_name_with_direction) != 0U) {
+
+  const std::string topic_name = static_cast<const char *>(req.pubsub_target.topic_name);
+  std::string_view suffix =
+    (req.direction == BridgeDirection::ROS2_TO_AGNOCAST) ? SUFFIX_PUBSUB_R2A : SUFFIX_PUBSUB_A2R;
+  const std::string topic_name_with_direction = topic_name + std::string(suffix);
```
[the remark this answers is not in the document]
bdm-k: This pattern is widely used across the codebase and the senders use `snprintf`, so this is fine.

src/agnocastlib/src/bridge/standard/agnocast_standard_bridge_loader.cpp
```diff
@@ -118,7 +151,7 @@ std::pair<BridgeFn, std::shared_ptr<void>> StandardBridgeLoader::resolve_factory
     if (raw_handle != nullptr) {
       dlclose(raw_handle);
     }
-    return {nullptr, nullptr};
+    return {0, nullptr};
   }
```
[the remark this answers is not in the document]
bdm-k: fixed in [6804533](https://github.com/autowarefoundation/agnocast/pull/1289/commits/68045338a000c90017bc2412cdca0112e1d44ebe)

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_node.hpp
```diff
@@ -208,17 +208,66 @@ class AgnocastToRosBridge : public BridgeBase
...
```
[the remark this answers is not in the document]
bdm-k: this is irrelevant; it actually compiles.

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_node.hpp
```diff
@@ -275,52 +324,47 @@ void send_standard_pubsub_bridge_request(
...
-  if (!is_self_executable && info.dli_sname != nullptr) {
-    symbol_to_send = info.dli_sname;
-  }
+template <typename ServiceT>
+void send_standard_service_bridge_request(
+  const std::string & service_name, BridgeDirection direction)
+{
+  static const auto logger = rclcpp::get_logger("agnocast_service_bridge_requester");
+
+  // Service bridges currently support only the ROS2 -> Agnocast direction.
```
kobayu858: Please leave a TODO comment.

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_utils.hpp
```diff
@@ -1,27 +1,39 @@
...
 {
 
-inline constexpr std::string_view SUFFIX_R2A = "_R2A";
-inline constexpr std::string_view SUFFIX_A2R = "_A2R";
-inline constexpr size_t SUFFIX_LEN = SUFFIX_R2A.length();
+inline constexpr std::string_view SUFFIX_PUBSUB_R2A = "_P_R2A";
+inline constexpr std::string_view SUFFIX_PUBSUB_A2R = "_P_A2R";
+inline constexpr std::string_view SUFFIX_SERVICE_R2A = "_S_R2A";
+inline constexpr std::string_view SUFFIX_SERVICE_A2R = "_S_A2R";
+inline constexpr size_t SUFFIX_LEN = SUFFIX_PUBSUB_R2A.length();
```
kobayu858: The service bridge suffix `_S_R2A` happens to be the same length as the pubsub suffix, so this is not a problem for now. However, if the length changes in the future, the suffix check in `check_active_pubsub_bridges()` will break. It would be safer to make it explicit via a comment or a name that `SUFFIX_LEN` is intended for pubsub only.

src/agnocastlib/src/bridge/standard/agnocast_standard_bridge_manager.cpp
```diff
@@ -435,13 +519,37 @@ void StandardBridgeManager::check_active_bridges()
...
 
-void StandardBridgeManager::check_managed_bridges()
+void StandardBridgeManager::check_and_remove_service_bridges()
 {
-  for (auto it = managed_bridges_.begin(); it != managed_bridges_.end();) {
+  for (auto it = active_r2a_service_bridges_.begin(); it != active_r2a_service_bridges_.end();) {
+    const std::string & service_name = it->first;
+
+    try {
+      get_service_qos(service_name);
```
kobayu858: Could you help me understand the intent behind the design decision to use `get_service_qos() `exception handling as the criterion for bridge removal?

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_node.hpp
bdm-k: I've added a TODO there, and removed A2R direction blocking in standard bridge loader ([1be0358](https://github.com/autowarefoundation/agnocast/pull/1289/commits/1be03588f3e05ef6bf325b119c1b67599051ae10)). The remaining places that block A2R direction are limited to bridge manager implementations, which are unlikely to be missed when adding A2R direction support.

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_utils.hpp
bdm-k: fixed in [c9134bb](https://github.com/autowarefoundation/agnocast/pull/1289/commits/c9134bbbe92ccedd7315f24d4e3d3becb7d8e6ed)
added `static_assert`s that ensure all suffixes are the same length

src/agnocastlib/src/bridge/standard/agnocast_standard_bridge_manager.cpp
bdm-k: The sole criterion for removing R2A service bridges is that the underlying Agnocast service has gone down. The name might be misleading, but `get_service_qos()` is actually a perfect fit for this.

I've added a brief comment about the usage of `get_service_qos()` ([2c860e4](https://github.com/autowarefoundation/agnocast/pull/1289/commits/2c860e4197952e3585bde36c76dc50fe41d14b38)).

kobayu858 (dismissed):

src/agnocastlib/src/bridge/standard/agnocast_standard_bridge_manager.cpp
```diff
@@ -124,17 +130,26 @@ void StandardBridgeManager::on_signal()
   }
 }
 
-void StandardBridgeManager::register_request(const MqMsgBridge & req)
+void StandardBridgeManager::register_pubsub_request(const MqMsgBridge & req)
 {
+  if (req.is_service) {
```
Koichi98: This early return (and its counterpart in `create_service_bridge_if_needed`) seems redundant — the function name states the precondition, and the only caller already dispatches on `is_service`. Having it inside each function also seems to hurt readability by suggesting this path could be reached for pubsub. If we want a guard, an `assert` would express "must never happen" more clearly than a silent return.

src/agnocastlib/src/bridge/standard/agnocast_standard_bridge_manager.cpp
Koichi98: Using `get_service_qos()` for its throw-on-missing side-effect is not really a good way — the name suggests we want the QoS, but we actually discard it. Could we wrap it in a `bool is_agnocast_service_alive(...)` helper so the intent is expliciit?
                                                                                                                                                                                                            
Longer term it'd be nicer to have a dedicated ioctl for this, mirroring pubsub. Could you leave a TODO?

src/agnocastlib/src/bridge/agnocast_bridge_utils.cpp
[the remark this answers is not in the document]
Koichi98: Could you add a bit more explicit TODO at `auto fn_reverse = fn_current;  // dummy value`? Noting that it must be replaced with `&start_a2r_service_node<ServiceT>` once A2R is implemented.

src/agnocastlib/src/bridge/standard/agnocast_standard_bridge_manager.cpp
```diff
@@ -383,22 +403,86 @@ bool StandardBridgeManager::should_remove_bridge(const std::string & topic_name,
...
+    factory_spec.shared_lib_path = std::nullopt;
+  } else {
+    factory_spec.shared_lib_path =
+      std::string(static_cast<const char *>(req.factory.shared_lib_path));
+  }
+  factory_spec.fn_offset_r2a = req.factory.fn_offset;
+  factory_spec.fn_offset_a2r = req.factory.fn_offset_reverse;
+
+  try {
+    // Check that the target service does not already exist in ROS 2.
```
Koichi98, line 438: Could you share the reasoning behind this duplication check? My take is that it might be cleaner to drop it and leave service-name uniqueness as the user's responsibility — the current check depends on ROS 2 discovery timing, so whether the bridge gets created or not can flip based on startup order, which feels more fragile than just not checking.

Koichi98: ROS2 seems like it is already non-deterministic with duplicate service names, so the same applies here when a user runs both an Agnocast service and a ROS 2 service under the same name.

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_node.hpp
Koichi98: Could `future.get()` block indefinitely here if no response ever arrives? My concern is that bridge teardown wouldn't be able to join the callback group while the callback is stuck here — is there a wake-up path I'm missing?

src/agnocastlib/src/bridge/agnocast_bridge_utils.cpp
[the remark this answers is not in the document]
bdm-k: Clarified `start_a2r_service_node` in the TODO comment (4af35f2).

src/agnocastlib/src/bridge/standard/agnocast_standard_bridge_manager.cpp
bdm-k: fixed in [34595fb](https://github.com/autowarefoundation/agnocast/pull/1289/commits/34595fb0e7e39d5c02bbc7e8abac6fed59bd9c38)

src/agnocastlib/src/bridge/standard/agnocast_standard_bridge_manager.cpp
bdm-k: I've removed early returns in `create_service_bridge_if_needed()` and `register_pubsub_request()` (7c02ba2).

src/agnocastlib/src/bridge/standard/agnocast_standard_bridge_manager.cpp
bdm-k: > the current check depends on ROS 2 discovery timing, so whether the bridge gets created or not can flip based on startup order, which feels more fragile than just not checking

Absolutely, but we previously agreed to keep a warning log when a ROS 2 service with the same name already exists, as a heuristic safeguard ([TIER IV internal link](https://tier4.atlassian.net/wiki/spaces/CRL/pages/4951212938/r2a+bridge?focusedCommentId=4950688915)). I would like to reconfirm your thoughts on this.

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_node.hpp
bdm-k: You've made an essential point. I'm trying to come up with a mitigation for this issue by the end of the day.

src/agnocastlib/src/bridge/standard/agnocast_standard_bridge_manager.cpp
Koichi98: I understand the prior agreement and have no objection to keeping the warning log itself. What I was pointing out is the return — the fact that whether the bridge gets created depends on ROS 2 discovery timing.
How about keeping the warning as a best-effort heuristic but not skipping bridge creation? Same-name collisions would then be the user's responsibility and we'd defer to ROS 2's handling.

bdm-k: Thanks, this makes sense and I agree.

I’ve updated both standard and performance bridge managers so the ROS 2 same-name check is now warning-only and no longer gates bridge creation ([042edd9](https://github.com/autowarefoundation/agnocast/pull/1289/commits/042edd98f5a5a2c2fa27736bb035c8183725f880)).

Koichi98: Great! Thanks

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_node.hpp
bdm-k: @Koichi98 
I believe the changes in commit 0ac4d34 do solve the issue, but I’m encountering another weird behavior: the bridge manager’s thread count decreases by only one, even though the logs indicate that two threads (one for the ROS 2 service and the other for the Agnocast client) have been joined.

Koichi98: I've reproduced the behavior, and I think I found out the main reason, which isn't caused by this PR. Let me address to this issue, and let me approve this PR once the fix is merged.
Thank you for finding out the bug.

Koichi98: I will share the link of the PR here in this conversation when I create it.

Koichi98: Since I've already checked that this issue isn't caused by this PR, I will approve this PR and merge. I am going to ask you to check whether the problem is solved, after the fix PR is merged.

Koichi98 (dismissed):

bdm-k:
@Koichi98 
We should check the return value of `snprintf()` across the codebase, as clang-tidy reports this issue. I think it would be better to terminate the program when `snprintf()` indicates the buffer is too small. To minimize the program's possible states, we should treat successful `snprintf()` calls as an invariant. What do you think?

Koichi98 (dismissed):

Koichi98:
@bdm-k 
> We should check the return value of snprintf() across the codebase, as clang-tidy reports this issue. I think it would be better to terminate the program when snprintf() indicates the buffer is too small. To minimize the program's possible states, we should treat successful snprintf() calls as an invariant. What do you think?

Good point. Totally agree with you, and thank you for pointing out. I've created the issue.
https://github.com/autowarefoundation/agnocast/issues/1317

bdm-k:
@Koichi98 
Regarding the `snprintf()` issue, I only updated the call sites in the `.cpp` files as a temporary fix to get clang-tidy to pass ([885e46a](https://github.com/autowarefoundation/agnocast/pull/1289/commits/885e46a7d19d695f74a8a4c05759948f1fbc570e)). I'm going to fix all `snprintf()` call sites and refactor the code in another PR.

Koichi98 (approved):

Koichi98:
@bdm-k 
> I'm going to fix all snprintf() call sites and refactor the code in another PR.

I would really appreciate it if you could do so. Thank you so much.

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_node.hpp
Koichi98: The main cause was that Agnocast doesn't currently support dynamic `add_callback_group` / `remove_callback_group`. This will be supported in the future, but since we don't expect the application to dynamically create the pub/sub/service/client, there is no problem for now.
https://github.com/autowarefoundation/agnocast/pull/1312

PR #1295, opened 2026-04-28
refactor(cie_thread_configurator): migrate non-ROS thread reporting from rclcpp pub/sub to abstract Unix domain socket

atsushi421:
## Description

`spawn_non_ros2_thread` previously created an isolated `rclcpp::Context`, `Node`, and `Publisher` on every call just to report a `(thread_id, thread_name)` pair to `thread_configurator_node`. That meant every non-ROS worker spawned a fresh DDS Participant, paid up to 5 s of DDS discovery latency, and then waited up to 500 ms more on `wait_for_all_acked`. This PR replaces the entire path with a single `sendto` on an abstract Unix domain `SOCK_DGRAM` socket — no DDS Participant, no discovery, no rclcpp on the publishing thread.

### What changed

- **New** `non_ros_thread_ipc.hpp` / `non_ros_thread_ipc.cpp` containing:
  - `struct NonRosThreadInfo { int64_t tid; std::string name; }` — the wire payload as a single named type, used uniformly by `encode` / `decode` / `send` / the listener `Callback`.
  - Wire format `[int64 tid][uint16 name_len][name bytes]`, host byte order. `static_assert(sizeof(NonRosThreadInfo::tid) == 8)` locks the size invariant at compile time.
  - `send_non_ros_thread_info(info, total_timeout=1s, retry_interval=50ms)` — encodes first, then opens the socket (so a `bad_alloc` in encode cannot leak the fd), and `sendto`s once. Retries on `ECONNREFUSED` / `EAGAIN` / `EWOULDBLOCK` / `ENOBUFS` / `EINTR` until the deadline.
  - `NonRosThreadInfoListener` RAII class owning the listener fd, a stop-`eventfd`, an `epoll` fd, and a private reader thread. `recv` is called with `MSG_TRUNC` so oversized datagrams are detected and dropped instead of silently truncated. Both fatal-exit paths in the reader thread (`epoll_wait` failure and `EPOLLERR/EPOLLHUP` on the listener fd) log at `RCLCPP_ERROR` with the same "Subsequent non-ROS thread announcements will not be received." note. The user-supplied callback is invoked inside a `try/catch` so that an exception from it cannot escape the `std::thread` entry point and call `std::terminate()`.
- `spawn_non_ros2_thread` collapses to constructing a `NonRosThreadInfo` and calling `send_non_ros_thread_info(info)` followed by the user function.
- `ThreadConfiguratorNode` and `PrerunNode` replace their ROS subscription on `/agnocast_cie_thread_configurator/non_ros_thread_info` with a `std::unique_ptr<NonRosThreadInfoListener>`. Both nodes gain a `stop()` method, called from `main.cpp` / `prerun_node_main.cpp` between `executor->spin()` returning and the post-spin reads (`print_all_unapplied()` / `dump_yaml_config()`), so the listener thread is joined before its mutated state is read.
- `agnocast_cie_config_msgs::msg::NonRosThreadInfo` is deleted (no remaining consumers in-tree).

### Design choices

- **Fixed abstract socket name** `\0agnocast_cie_thread_configurator/non_ros_thread_info`. Two configurators on the same host cannot both bind; the second throws `std::system_error` (EADDRINUSE). UID/ROS_DOMAIN_ID disambiguation was considered and dropped as YAGNI. The constraint is documented on the listener class.
- **Weak ack semantics** preserved: client returns once the datagram is in the kernel buffer, matching the previous DDS-level `wait_for_all_acked` (which also did not guarantee callback execution). The user thread starts as soon as the IPC call returns.
- **Host byte order** on the wire — single-host AUDS, no endianness concern.
- **`epoll_wait(-1)` + `eventfd`** for shutdown so that `stop()` wakes the reader thread cleanly without racing `close()` against a concurrent `recv`.
- **Stack-allocated reader buffer**: the listener's 64 KB+10 receive buffer is a `std::array` on the reader thread's stack rather than a heap-allocated `std::vector`.
- **Constructor exception safety**: every fd-acquiring step closes earlier-acquired fds before throwing, including the final `std::thread` construction. `stop()` is idempotent via atomic CAS and is the single owner of fd close + thread join.
- **Sender-side logging** uses `std::fprintf(stderr, "[agnocast_cie_thread_configurator][WARN] ...")` rather than `RCLCPP_*` because `send_non_ros_thread_info` runs on user threads where no rclcpp `Logger` is in scope.

### Behavioural notes

- The e2e log line `Received NonRosThreadInfo:` is preserved verbatim, so `test_agnocast_callback_isolated_executor_launch.py::test_thread_configurator_receives_non_ros_thread_info` continues to assert correctly.
- `spawn_non_ros2_thread`'s public signature is unchanged. Existing callers (e.g. `test_publisher_component.cpp`) recompile and run unmodified.
- If the configurator is not yet running when a thread is spawned, the client retries for 1 s then logs to stderr `No listener for non_ros_thread_info; thread '<name>' will not be configured.` and runs the user function anyway — same UX as before, shorter timeout.

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

Locally:
- `colcon build --packages-up-to agnocast_cie_thread_configurator --cmake-args -DBUILD_TESTING=ON` clean.
- `test_agnocast_cie_thread_configurator`: 17 / 17 passing (encode/decode round trips + edge cases, sockaddr setup, send happy path + timeout, listener bind/destruct, second-bind throws, dispatch, malformed-drop, stop idempotency, send-after-stop). Tests that need to bind the well-known abstract path `GTEST_SKIP()` (instead of failing) when another configurator is already running on the host.
- `colcon test --packages-select agnocast_cie_thread_configurator agnocast_cie_config_msgs`: clean.

- \`need-major-update\`: User API breaking changes
- \`need-minor-update\`: Internal API breaking changes (heaphook/kmod/agnocastlib compatibility)
- \`need-patch-update\`: Bug fixes and other changes

**Important notes:**

- If you need \`need-major-update\` or \`need-minor-update\`, please include this in the PR title as well.
  - Example: \`fix(foo)[needs major version update]: bar\` or \`feat(baz)[needs minor version update]: qux\`
- After receiving approval from reviewers, add the \`run-build-test\` label. The PR can only be merged after the build tests pass.


src/agnocast_cie_thread_configurator/include/agnocast_cie_thread_configurator/thread_configurator_node.hpp
```diff
@@ -1,20 +1,27 @@
...
 
 class ThreadConfiguratorNode : public rclcpp::Node
 {
-  // Each ThreadConfig instance is reachable from exactly one MutuallyExclusive
-  // callback group, so no two callbacks ever touch the same instance. Hence
-  // thread_id and applied stay non-atomic; print_all_unapplied reads post-spin.
+  // Each ThreadConfig instance is touched by exactly one writer:
+  // - callback_group_configs_ entries: written from the rclcpp executor's
+  //   non_ros_thread MutuallyExclusive callback group used for the
+  //   CallbackGroupInfo subscription.
```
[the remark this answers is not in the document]
atsushi421: Fixed in 80206668. Rewrote the comment block: callback_group_configs_ entries are now correctly described as written from CallbackGroupInfo subscription callbacks dispatched by the SingleThreadedExecutor on the node's default callback group (one subscription on the main node and one per per-domain node), and non_ros_thread_configs_ entries are written from the NonRosThreadInfoListener's private reader thread. The two vectors are disjoint, so there is no shared mutable state between the executor thread and the listener thread.

src/agnocast_cie_thread_configurator/src/non_ros_thread_ipc.cpp
```diff
@@ -0,0 +1,191 @@
...
+          buf.size());
+        continue;
+      }
+      NonRosThreadInfo info;
+      if (!decode_non_ros_thread_info(buf.data(), static_cast<size_t>(n), info)) {
+        RCLCPP_WARN(logger_, "Discarded malformed NonRosThreadInfo datagram (size=%zd)", n);
+        continue;
+      }
+      callback_(std::move(info));
+    }
```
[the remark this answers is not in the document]
atsushi421: Fixed in 80206668. Wrapped callback_(std::move(info)) in a try/catch (std::exception and catch-all). Exceptions from the user-supplied callback are now logged via RCLCPP_ERROR and the message is dropped, keeping the listener thread alive instead of letting the exception escape the std::thread entry point and call std::terminate().

src/agnocast_cie_thread_configurator/src/non_ros_thread_ipc.cpp
```diff
@@ -0,0 +1,191 @@
...
+      if (errno == EINTR) {
+        continue;
+      }
+      RCLCPP_ERROR(
+        logger_,
+        "epoll_wait() failed: errno=%d; reader thread exiting. "
+        "Subsequent non-ROS thread announcements will not be received.",
+        errno);
+      return;
+    }
```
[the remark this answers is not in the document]
atsushi421: Good catch on the inconsistency between the PR description and the code. We chose to fix the description rather than the code in this commit: introducing a separate cleanup_done_ atomic to decouple observability from stop()'s idempotency CAS adds complexity that no caller currently exploits (stopped_ is private and never queried externally), and the existing semantics where stop() drives all cleanup are simpler to reason about. The PR description will be updated separately to drop the claim that the worker sets stopped_=true on fatal exit.

src/agnocast_cie_thread_configurator/include/agnocast_cie_thread_configurator/non_ros_thread_ipc.hpp
```diff
@@ -0,0 +1,188 @@
...
+// Wire format: [int64 tid][uint16 name_len][name_len bytes of name].
+// Host byte order — single-host AUDS, no endianness concern.
+struct NonRosThreadInfo
+{
+  int64_t tid = 0;
+  std::string name;
+};
+
+constexpr size_t k_non_ros_thread_info_max_name_len = 65535u;
+constexpr size_t k_non_ros_thread_info_max_wire_size = 10u + k_non_ros_thread_info_max_name_len;
```
kobayu858: The header size `10u` is hardcoded, but it implicitly assumes `sizeof(int64_t) == 8` and `sizeof(uint16_t) == 2`. If the type of tid were ever changed (e.g. to int32_t), this constant would silently become wrong while encode/decode use the actual sizeof.

Consider deriving it explicitly:

```
cppconstexpr size_t k_wire_header_size =
    sizeof(NonRosThreadInfo::tid) + sizeof(uint16_t);
constexpr size_t k_non_ros_thread_info_max_wire_size =
    k_wire_header_size + k_non_ros_thread_info_max_name_len;
```

src/agnocast_cie_thread_configurator/include/agnocast_cie_thread_configurator/non_ros_thread_ipc.hpp
```diff
@@ -0,0 +1,188 @@
...
+  const uint16_t name_len = static_cast<uint16_t>(info.name.size());
+  out_len = 10u + name_len;
+  if (out_len > buf_cap) {
+    return false;
+  }
+  std::memcpy(buf, &info.tid, sizeof(info.tid));
+  std::memcpy(buf + 8, &name_len, sizeof(name_len));
+  if (name_len > 0) {
+    std::memcpy(buf + 10, info.name.data(), name_len);
+  }
```
kobayu858: `out_len = k_wire_header_size + name_len;`

```
std::memcpy(buf,                  &info.tid, k_wire_tid_size);
std::memcpy(buf + k_wire_tid_size, &name_len, k_wire_namelen_size);
std::memcpy(buf + k_wire_header_size, info.name.data(), name_len)
```

src/agnocast_cie_thread_configurator/include/agnocast_cie_thread_configurator/non_ros_thread_ipc.hpp
```diff
@@ -0,0 +1,188 @@
...
+  return true;
+}
+
+inline bool decode_non_ros_thread_info(const uint8_t * buf, size_t len, NonRosThreadInfo & info)
+{
+  if (len < 10u) {
```
kobayu858: k_wire_header_size

src/agnocast_cie_thread_configurator/include/agnocast_cie_thread_configurator/non_ros_thread_ipc.hpp
```diff
@@ -0,0 +1,188 @@
...
+    return false;
+  }
+  uint16_t name_len = 0;
+  std::memcpy(&info.tid, buf, sizeof(info.tid));
+  std::memcpy(&name_len, buf + 8, sizeof(name_len));
```
kobayu858: k_wire_tid_size

src/agnocast_cie_thread_configurator/include/agnocast_cie_thread_configurator/non_ros_thread_ipc.hpp
```diff
@@ -0,0 +1,188 @@
...
+  if (len != 10u + static_cast<size_t>(name_len)) {
```
kobayu858: k_wire_header_size

src/agnocast_cie_thread_configurator/include/agnocast_cie_thread_configurator/non_ros_thread_ipc.hpp
```diff
@@ -0,0 +1,188 @@
...
+    return false;
+  }
+  info.name.assign(reinterpret_cast<const char *>(buf + 10), name_len);
```
kobayu858: k_wire_header_size

src/agnocast_cie_thread_configurator/test/test_non_ros_thread_ipc.cpp
```diff
@@ -0,0 +1,276 @@
...
+  acie::NonRosThreadInfoListener listener(
+    [&](const acie::NonRosThreadInfo & info) {
+      std::lock_guard<std::mutex> lk(m);
+      received.push_back(info);
+      cv.notify_one();
+    },
+    test_logger());
+
+  acie::send_non_ros_thread_info({100, "alpha"});
+  acie::send_non_ros_thread_info({200, "beta"});
```
kobayu858, line 218: `SOCK_DGRAM` preserves datagram boundaries, but ordering is not guaranteed by the specification even for AUDS on the same host. While messages almost always arrive in order in practice, asserting a specific order in a test is strictly incorrect.

src/agnocast_cie_thread_configurator/src/non_ros_thread_ipc.cpp
```diff
@@ -0,0 +1,204 @@
...
+      if (!decode_non_ros_thread_info(buf.data(), static_cast<size_t>(n), info)) {
+        RCLCPP_WARN(logger_, "Discarded malformed NonRosThreadInfo datagram (size=%zd)", n);
+        continue;
+      }
+      // The user-supplied callback is invoked on this private reader thread,
+      // which is the std::thread entry point. Letting an exception escape it
+      // would call std::terminate() and take down the whole process, so we
+      // contain failures here, log them, and drop the message.
+      try {
+        callback_(std::move(info));
```
kobayu858, line 188: `stop()` calls `thread_.join()`, but if the user's callback (invoked on the reader thread inside `run()`) triggers `stop()` — e.g. by destroying the owning node — `join()` will return `EDEADLK`, which is silently swallowed by the `catch (const std::system_error &)` block. Execution then falls through to `::close(epfd_)` / `::close(listener_fd_)` while `run()` is still live on that thread, giving use-after-close UB.

src/agnocast_cie_thread_configurator/include/agnocast_cie_thread_configurator/non_ros_thread_ipc.hpp
atsushi421: Fixed in 1619ee52. Added named constants `k_wire_tid_size = sizeof(NonRosThreadInfo::tid)`, `k_wire_namelen_size = sizeof(uint16_t)`, and `k_wire_header_size = k_wire_tid_size + k_wire_namelen_size` so the layout is derived from the actual `sizeof` rather than hardcoded as `10u`. `k_non_ros_thread_info_max_wire_size` now composes from `k_wire_header_size`, and all of `encode`/`decode` use the named constants too.

src/agnocast_cie_thread_configurator/include/agnocast_cie_thread_configurator/non_ros_thread_ipc.hpp
atsushi421: Fixed in 1619ee52. `encode_non_ros_thread_info` now uses `out_len = k_wire_header_size + name_len`, copies tid with `k_wire_tid_size`, name_len at `buf + k_wire_tid_size` with `k_wire_namelen_size`, and the name at `buf + k_wire_header_size`.

src/agnocast_cie_thread_configurator/include/agnocast_cie_thread_configurator/non_ros_thread_ipc.hpp
atsushi421: Fixed in 1619ee52. The header-length guard is now `if (len < k_wire_header_size)`.

src/agnocast_cie_thread_configurator/include/agnocast_cie_thread_configurator/non_ros_thread_ipc.hpp
atsushi421: Fixed in 1619ee52. The decode tid copy is now `std::memcpy(&info.tid, buf, k_wire_tid_size)` and the name_len read uses `buf + k_wire_tid_size` / `k_wire_namelen_size`.

src/agnocast_cie_thread_configurator/include/agnocast_cie_thread_configurator/non_ros_thread_ipc.hpp
atsushi421: Fixed in 1619ee52. The size check now reads `if (len != k_wire_header_size + static_cast<size_t>(name_len))`.

src/agnocast_cie_thread_configurator/include/agnocast_cie_thread_configurator/non_ros_thread_ipc.hpp
atsushi421: Fixed in 1619ee52. The name copy is now `info.name.assign(reinterpret_cast<const char *>(buf + k_wire_header_size), name_len)`.

src/agnocast_cie_thread_configurator/test/test_non_ros_thread_ipc.cpp
atsushi421: Fixed in 1619ee52. The test now sorts the received vector by `tid` before asserting on contents, so it no longer relies on SOCK_DGRAM preserving order.

src/agnocast_cie_thread_configurator/src/non_ros_thread_ipc.cpp
atsushi421: Good catch — silently swallowing EDEADLK and then closing the fds while `run()` is still live is unsound. Even a runtime self-detect that detaches the reader thread does not fully fix the underlying problem: if the user destroys the listener from inside the callback, `*this` is torn down while the reader thread is still using `callback_`/`logger_`/the fds, which is UAF regardless of how `stop()` behaves. We addressed this in 1619ee52 by documenting on `NonRosThreadInfoListener` that `stop()` and the destructor must NOT be called from within the user callback (the reader thread cannot synchronously dispose of itself); the recommended pattern is to post the termination request to a different thread (e.g. an atomic flag observed by the executor thread, which then drives destruction). Happy to add a runtime detection + abort if you'd prefer a louder failure mode for misuse, but the core fix is the lifetime contract.

kobayu858 (approved):

PR #1300, opened 2026-05-01
feat(bridge): introduce shadow node to Agnocast bridge

bdm-k:
## Description

This PR introduces a new concept called *shadow nodes* to Agnocast bridge. Shadow nodes are lightweight `rclcpp::Node`s created in the bridge manager that share the same name as a specific `agnocast::Node`. In other words, they're *shadows* of that Agnocast node for other ROS 2 nodes.

### Motivation

The motivation behind shadow nodes is to use the `ros2 param` command with Agnocast nodes. Let's say we want to query an Agnocast node named `/agnocast_server` using the `ros2 param` command. If there is no shadow node of `/agnocast_server`, the `ros2 param` command thinks there is no such node named `/agnocast_server`, so the command fails. In fact, `/agnocast_server`'s parameter services are available on the ROS 2 side because an R2A service bridge exists. Therefore, we need to trick ROS 2 nodes into thinking that `/agnocast_service` exists so that they send service requests to it, which can then be received by the R2A service bridge or the bridge manager's node.

### Implementation details

The MQ struct for requesting bridges is extended to include information about whether the bridge manager needs to create a shadow node along with a bridge. Shadow nodes are only requested when Agnocast nodes request an R2A service bridge.

When requested, the bridge manager creates the shadow node. Shadow nodes are created without parameter services or parameter events, and they are not scheduled by the executor.

The bridge manager then packs each created shadow node with its corresponding R2A service bridge and stores them in the `active_r2a_service_bridges_` member:
```c++
struct R2AServiceBridgeItem
{
  std::shared_ptr<ServiceBridgeBase> bridge;
  rclcpp::Node::SharedPtr shadow_node;
};
std::map<std::string, R2AServiceBridgeItem> active_r2a_service_bridges_;
```
A mapping from a node name to a weak pointer of its shadow node is also added as a data member of the bridge manager:
```c++
std::map<std::string, rclcpp::Node::WeakPtr> shadow_nodes_;
```
This is useful when retrieving an existing shadow node in the bridge manager. But this might be overengineering, so just let me know if you think so.

## How was this PR tested?

- [ ] Autoware (required)
- [x] sample application

The `ros2 param` command now works with `agnocast::Node`.


src/agnocastlib/include/agnocast/bridge/performance/agnocast_performance_bridge_manager.hpp
```diff
@@ -36,9 +48,11 @@ class PerformanceBridgeManager
 
   std::unordered_map<std::string, PerformancePubsubBridgeResult> active_pubsub_r2a_bridges_;
   std::unordered_map<std::string, PerformancePubsubBridgeResult> active_pubsub_a2r_bridges_;
-  std::unordered_map<std::string, PerformanceServiceBridgeResult> active_r2a_service_bridges_;
   std::unordered_map<std::string, RequestMap> request_cache_;
 
+  std::unordered_map<std::string, R2AServiceBridgeItem> active_r2a_service_bridges_;
+  std::unordered_map<std::string, rclcpp::Node::WeakPtr> shadow_nodes_;
+
```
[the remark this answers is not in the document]
bdm-k: fixed

src/agnocastlib/include/agnocast/bridge/standard/agnocast_standard_bridge_manager.hpp
```diff
@@ -55,13 +55,25 @@ class StandardBridgeManager
...
+  {
+    std::shared_ptr<ServiceBridgeBase> bridge;
+    rclcpp::Node::SharedPtr shadow_node;
+
+    R2AServiceBridgeItem(
+      std::shared_ptr<ServiceBridgeBase> bridge, rclcpp::Node::SharedPtr shadow_node)
+    : bridge(std::move(bridge)), shadow_node(std::move(shadow_node))
+    {
+    }
+  };
```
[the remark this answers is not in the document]
bdm-k: fixed

src/agnocastlib/src/bridge/standard/agnocast_standard_bridge_manager.cpp
```diff
@@ -383,22 +403,113 @@ bool StandardBridgeManager::should_remove_bridge(const std::string & topic_name,
...
+  }
+  factory_spec.fn_offset_r2a = req.factory.fn_offset;
+  factory_spec.fn_offset_a2r = req.factory.fn_offset_reverse;
+
+  try {
+    // Check that the target service does not already exist in ROS 2.
+    const auto services = container_node_->get_service_names_and_types();
+    bool exists = std::any_of(services.begin(), services.end(), [&service_name](const auto & s) {
+      return s.first == service_name;
+    });
```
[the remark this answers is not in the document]
bdm-k: fixed

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_utils.hpp
```diff
@@ -1,27 +1,39 @@
...
 {
 
-inline constexpr std::string_view SUFFIX_R2A = "_R2A";
-inline constexpr std::string_view SUFFIX_A2R = "_A2R";
-inline constexpr size_t SUFFIX_LEN = SUFFIX_R2A.length();
+inline constexpr std::string_view SUFFIX_PUBSUB_R2A = "_P_R2A";
+inline constexpr std::string_view SUFFIX_PUBSUB_A2R = "_P_A2R";
+inline constexpr std::string_view SUFFIX_SERVICE_R2A = "_S_R2A";
+inline constexpr std::string_view SUFFIX_SERVICE_A2R = "_S_A2R";
+inline constexpr size_t SUFFIX_LEN = SUFFIX_PUBSUB_R2A.length();
```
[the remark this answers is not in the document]
bdm-k: fixed

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_node.hpp
```diff
@@ -208,17 +222,66 @@ class AgnocastToRosBridge : public BridgeBase
...
+    ros_srv_ = parent_node->create_service<ServiceT>(
+      service_name,
+      [this](
+        const typename ServiceT::Request::SharedPtr ros_req,
+        typename ServiceT::Response::SharedPtr ros_res) {
+        auto agno_req = this->agno_client_->borrow_loaned_request();
+        *agno_req = *ros_req;
+
+        auto future = this->agno_client_->async_send_request(std::move(agno_req));
+        auto agno_res = future.get();
```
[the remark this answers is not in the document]
bdm-k: this is outdated

src/agnocastlib/src/bridge/standard/agnocast_standard_bridge_manager.cpp
```diff
@@ -470,6 +480,26 @@ void StandardBridgeManager::create_service_bridge_if_needed(const MqMsgBridge &
...
+  auto it = shadow_nodes_.find(node_name);
+  if (it != shadow_nodes_.end()) {
+    if (auto node = it->second.lock()) {
+      return node;
+    }
+  }
+
+  rclcpp::NodeOptions options;
+  options.start_parameter_services(false);
+  options.start_parameter_event_publisher(false);
```
Koichi98: We might need to disable some other node options such as `/rosout` publisher, etc.

src/agnocastlib/include/agnocast/bridge/performance/agnocast_performance_bridge_manager.hpp
```diff
@@ -36,9 +51,11 @@ class PerformanceBridgeManager
 
   std::unordered_map<std::string, PerformancePubsubBridgeResult> active_pubsub_r2a_bridges_;
   std::unordered_map<std::string, PerformancePubsubBridgeResult> active_pubsub_a2r_bridges_;
-  std::unordered_map<std::string, PerformanceServiceBridgeResult> active_r2a_service_bridges_;
   std::unordered_map<std::string, RequestMap> request_cache_;
 
+  std::unordered_map<std::string, R2AServiceBridgeItem> active_r2a_service_bridges_;
+  std::unordered_map<std::string, rclcpp::Node::WeakPtr> shadow_nodes_;
```
Koichi98: > This is useful when retrieving an existing shadow node in the bridge manager. But this might be overengineering, so just let me know if you think so.

In my opinion, I don't think we need this specific data structure for shadow_nodes. We already have the same information in `active_r2a_service_bridges_`, so we can just go through it.

Adding data structures that hold redundant information tends to introduce complexity: the copies can drift out of sync, and forgetting to clean them up on teardown leads to memory leaks. I'd prefer to avoid this when possible. If there's a specific reason it's needed, I'm fine with keeping it.

src/agnocastlib/src/bridge/agnocast_bridge_utils.cpp
```diff
@@ -292,4 +292,16 @@ bool build_bridge_factory_info(
   return true;
 }
 
+std::pair<std::string, std::string> split_full_node_name(std::string_view fqn)
```
Koichi98: Do we actually need `split_full_node_name`? The sender has `get_name()` / `get_namespace()` available separately — sending them as two mqmsg fields would be cleaner than joining and re-splitting.

src/agnocastlib/src/bridge/performance/agnocast_performance_bridge_manager.cpp
```diff
@@ -431,4 +438,24 @@ void PerformanceBridgeManager::remove_invalid_requests(
...
+      return node;
+    }
+  }
+
+  rclcpp::NodeOptions options;
+  options.start_parameter_services(false);
+  options.start_parameter_event_publisher(false);
+
+  auto [ns, name] = split_full_node_name(node_name);
+  auto node = std::make_shared<rclcpp::Node>(name, ns, options);
```
Koichi98: Since we now have both `rclcpp::Node` and `agnocast::Node` with the same name on one system, we need to fix `ros2 node list_agnocast` command below.
https://github.com/autowarefoundation/agnocast/blame/21bb0f4d7a38dbd1de832aeafcec8a277d4d629a/src/ros2agnocast/ros2agnocast/verb/node_list_agnocast.py#L103

@TetsuKawa 
Quick question: why the extra `node_name not in ros2_node_name` here, rather than just checking `node_name in agnocast_node_name`? Want to make sure I understand the intent.

TetsuKawa: @Koichi98 

There is a discrepancy between the implementation and the documentation.
The [documentation](https://github.com/autowarefoundation/agnocast/blob/21bb0f4d7a38dbd1de832aeafcec8a277d4d629a/docs/ros2_command_extension.md?plain=1#L190) states that "(Agnocast enabled)" should only be appended to Agnocast Nodes. However, the current code also appends it to ROS 2 Nodes that contain Agnocast Topics. My original intent was for the implementation to be the correct behavior, meaning the documentation needs to be updated.

src/agnocastlib/include/agnocast/bridge/performance/agnocast_performance_bridge_manager.hpp
bdm-k: I removed `shadow_nodes_` and changed the code to go through `active_r2a_service_bridges_`.

Thanks for your feedback.

src/agnocastlib/src/bridge/agnocast_bridge_utils.cpp
bdm-k: fixed

src/agnocastlib/src/bridge/standard/agnocast_standard_bridge_manager.cpp
bdm-k: https://github.com/ros2/rclcpp/blob/52f97af45490cbbe2587fd2a668042cea9aced3d/rclcpp/include/rclcpp/node_options.hpp#L39-L65
I reviewed `NodeOptions`'s default values and considered `use_global_arguments` and `enable_rosout` should also be turned off to make shadow nodes zero-footprint. I would appreciate it if you could go through the node options.

src/agnocastlib/src/bridge/performance/agnocast_performance_bridge_manager.cpp
```diff
@@ -431,4 +439,24 @@ void PerformanceBridgeManager::remove_invalid_requests(
   }
 }
 
+rclcpp::Node::SharedPtr PerformanceBridgeManager::create_shadow_node_if_needed(
```
kobayu858: `create_shadow_node_if_needed` is duplicated between `StandardBridgeManager` and `PerformanceBridgeManager` with identical logic. Could we extract this into a shared utility function, e.g. in `agnocast_bridge_utils`?

src/agnocastlib/src/bridge/performance/agnocast_performance_bridge_manager.cpp
bdm-k: fixed in [f44e1f0](https://github.com/autowarefoundation/agnocast/pull/1300/commits/f44e1f00f21a14b94b0052437a37443d232e76a0)

I used parameter services to probe whether an rclcpp node is likely a shadow node. Likely shadow nodes are excluded in advance so that the corresponding Agnocast nodes can be listed with "(Agnocast enabled)".

src/agnocastlib/src/bridge/performance/agnocast_performance_bridge_manager.cpp
bdm-k: fixed in [260b2f6](https://github.com/autowarefoundation/agnocast/pull/1300/commits/260b2f6a4104c7c76173972703b5b3eadbf759f5)

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_utils.hpp
```diff
@@ -69,4 +71,38 @@ bool build_bridge_factory_info(
...
+template <typename MapT>
+std::shared_ptr<rcl_node_t> find_or_create_shadow_node(
+  const MapT & active_r2a_service_bridges, const std::string & ns, const std::string & name)
+{
+  for (const auto & [_, item] : active_r2a_service_bridges) {
+    const std::shared_ptr<rcl_node_t> & shadow_node = item.shadow_node;
+    if (
+      strcmp(rcl_node_get_name(shadow_node.get()), name.c_str()) == 0 &&
+      strcmp(rcl_node_get_namespace(shadow_node.get()), ns.c_str()) == 0) {
+      return shadow_node;
```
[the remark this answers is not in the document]
bdm-k: fixed

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_utils.hpp
```diff
@@ -5,7 +5,9 @@
 
 #include <rclcpp/rclcpp.hpp>
 
+#include <memory>
 #include <string>
+#include <string_view>
 #include <utility>
```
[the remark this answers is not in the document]
bdm-k: fixed

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_utils.hpp
```diff
@@ -69,4 +71,38 @@ bool build_bridge_factory_info(
...
+
+  rcl_context_t * rcl_ctx = rclcpp::contexts::get_global_default_context()->get_rcl_context().get();
+
+  rcl_node_options_t options = rcl_node_get_default_options();
+  options.enable_rosout = false;
+
+  auto del = [](rcl_node_t * node) {
+    if (rcl_node_fini(node) != RCL_RET_OK) {
+      RCUTILS_LOG_ERROR_NAMED(
+        "agnocast_bridge", "Error in destruction of shadow node: %s", rcl_get_error_string().str);
```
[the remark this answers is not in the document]
bdm-k: fixed

src/ros2agnocast/ros2agnocast/verb/node_list_agnocast.py
```diff
@@ -6,6 +6,10 @@
 from ros2node.api import get_node_names
 from ros2topic.verb import VerbExtension
 
+def split_fqn(fqn):
+  namespace, _, name = fqn.rpartition('/')
+  return (namespace or '/'), name
```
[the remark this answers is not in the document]
bdm-k: fixed

bdm-k:
@Koichi98 @kobayu858 
`/parameter_events` subscriber is always created with `rclcpp::Node`, and it cannot be disabled via `NodeOptions`. As a result, the `/parameter_events` subscriber on shadow nodes showed up when running `ros2 node info_agnocast`. So, I changed the implementation to create shadow nodes directly using the rcl API instead. Now, no pub/sub/server/client is ever created on shadow nodes at all.

src/agnocastlib/src/bridge/standard/agnocast_standard_bridge_manager.cpp
bdm-k: This is no longer relevant since it moved to use rcl API to create shadow nodes.

src/agnocastlib/src/bridge/performance/agnocast_performance_bridge_manager.cpp
bdm-k: \[update] All pub/sub/service/client existence is checked. If any of them exist, that node is considered not a shadow node.

Koichi98: The added pub/sub/client introspection may be costly on large systems. Since `start_parameter_services` defaults to true, the service check alone is already fairly reliable for normal rclcpp nodes. I might drop the extra checks later if overhead turns out to be an issue — just flagging.

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_utils.hpp
```diff
@@ -69,4 +73,41 @@ bool build_bridge_factory_info(
...
+      shadow_node != nullptr && strcmp(rcl_node_get_name(shadow_node.get()), name.c_str()) == 0 &&
+      strcmp(rcl_node_get_namespace(shadow_node.get()), ns.c_str()) == 0) {
+      return shadow_node;
+    }
+  }
+
+  rcl_context_t * rcl_ctx = rclcpp::contexts::get_global_default_context()->get_rcl_context().get();
+
+  rcl_node_options_t options = rcl_node_get_default_options();
+  options.enable_rosout = false;
```
Koichi98, line 92: Should we also set `options.use_global_arguments = false` here? The previous `rclcpp::Node` version had it disabled, but it wasn't carried over to the `rcl_node_t` version.

bdm-k: Just setting `options.use_global_arguments = false` somehow did not work in my last experiment 😢
I'll look a bit deeper to figure out the root cause.

bdm-k: fixed in [534dd94](https://github.com/autowarefoundation/agnocast/pull/1300/commits/534dd949bba7e32d573c420c9ac30165e742e718)

Setting `options.use_global_arguments` to `false` requires proper initialization of `options.arguments`.

Koichi98 (approved):

PR #1303, opened 2026-05-01
feat(cie_thread_configurator): add ~/reapply_config service for runtime YAML reload

atsushi421:
## Description

Adds a `~/reapply_config` ROS service to `ThreadConfiguratorNode` so operators can edit the configurator's YAML and reapply scheduling attributes to running threads without restarting the configurator or any target application. Typical tuning loop becomes `vim config.yaml && ros2 service call .../reapply_config`.

**Behavior** (validate-then-apply):

1. Parse the read-only `config_file` parameter.
2. Per-entry validation via `parse_yaml`. Per-entry errors abort with `success=false` and no state mutation. `parse_yaml` also rejects duplicate keys, since the configurator's id maps would otherwise silently collapse duplicates to the last-inserted entry.
3. Carry over `thread_id` from the existing in-memory state into the new vectors by direct lookup in `id_to_*_config_`. `applied` is intentionally NOT carried over — every entry restarts as `applied=false` and is promoted to `true` by the syscall loop only when the new config has actually landed, so the response's `failed_*` array stays consistent with the in-memory state.
4. Swap the in-memory vectors and rebuild the id maps. Added entries (key in new but not in old) enter as `thread_id=-1` / `applied=false`; removed entries (key in old but not in new) drop out and the configurator no longer manages them (their kernel-level scheduling persists).
5. Apply syscalls. Per-entry results land in `applied_*` / `failed_*` / `skipped_*`. `skipped_*` means the thread has not yet announced itself; the new params will take effect on the next announcement. When all entries succeed cleanly the handler logs `Reapply: all entries successfully (re-)applied` so operators have a per-call health signal (`configured_at_least_once_` is one-shot for the process lifetime and is not re-emitted).

**Concurrency:** `non_ros_state_mutex_` protects `non_ros_thread_configs_` against the AUDS-driven listener thread. The non-ROS carry-over, swap, and `unapplied_num_` recompute share one critical section so the listener cannot squeeze a `thread_id` update into the OLD vector between phases. Callback-group state needs no lock — both subscriptions and the reapply handler run on the same `SingleThreadedExecutor`.

**Out of scope:** re-validating `hardware_info` / `rt_throttling` (changes there require a process restart, documented in the `.srv`), cleaning up orphaned `/sys/fs/cgroup/cpuset/<id>` directories.

**Tests:** unit tests cover duplicate-key rejection and same-id-different-domain acceptance. Launch tests cover priority change, added entry, removed entry, invalid policy, and the non-ROS-thread reapply path (separate from the callback-group path because of the mutex). The destructive removal test is named with a `zz_` prefix so it runs last alphabetically — removing a callback group clears the configurator's in-memory state and there is no re-announcement mechanism, so any test that runs after it would see a stale `thread_id=-1`. `test_publisher_component`'s non-ROS worker is now kept alive for the node lifetime so the reapply path has a live tid to operate on; an empty body would let the thread exit and the syscall would get ESRCH.

## How was this PR tested?

- [ ] Autoware (required)
- [ ] \`bash scripts/test/e2e_test_1to1.bash\` (required)
- [ ] \`bash scripts/test/e2e_test_2to2.bash\` (required)
- [x] \`bash scripts/test/run_requires_kernel_module_tests.bash\` (required)
- [x] sample application

- \`need-major-update\`: User API breaking changes
- \`need-minor-update\`: Internal API breaking changes (heaphook/kmod/agnocastlib compatibility)
- \`need-patch-update\`: Bug fixes and other changes

**Important notes:**

- If you need \`need-major-update\` or \`need-minor-update\`, please include this in the PR title as well.
  - Example: \`fix(foo)[needs major version update]: bar\` or \`feat(baz)[needs minor version update]: qux\`
- After receiving approval from reviewers, add the \`run-build-test\` label. The PR can only be merged after the build tests pass.

### DIFF
--- a/src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
+++ b/src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
@@ -256,12 +256,9 @@ void ThreadConfiguratorNode::print_all_unapplied()
 
   RCLCPP_WARN(this->get_logger(), "Following non-ROS threads are not yet configured");
 
-  {
-    std::lock_guard<std::mutex> lk(non_ros_state_mutex_);
-    for (auto & config : non_ros_thread_configs_) {
-      if (!config.applied) {
-        RCLCPP_WARN(this->get_logger(), "  - %s", config.thread_str.c_str());
-      }
+  for (auto & config : non_ros_thread_configs_) {
+    if (!config.applied) {
+      RCLCPP_WARN(this->get_logger(), "  - %s", config.thread_str.c_str());
     }
   }
 }
@@ -575,10 +572,8 @@ void ThreadConfiguratorNode::on_reapply_config_request(
     }
     if (issue_syscalls(cfg)) {
       response->applied_callback_groups.push_back(std::move(key));
-      if (!cfg.applied) {
-        cfg.applied = true;
-        unapplied_num_.fetch_sub(1, std::memory_order_acq_rel);
-      }
+      cfg.applied = true;
+      unapplied_num_.fetch_sub(1, std::memory_order_acq_rel);
     } else {
       response->failed_callback_groups.push_back(std::move(key));
     }
--- a/src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
+++ b/src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
@@ -7,6 +7,7 @@
 #include "yaml-cpp/yaml.h"
 
 #include "agnocast_cie_config_msgs/msg/callback_group_info.hpp"
+#include "agnocast_cie_config_msgs/srv/reapply_config.hpp"
 
 #include <error.h>
 #include <sys/resource.h>
@@ -23,30 +24,33 @@
 using agnocast_cie_thread_configurator::policy_to_sched_const;
 
 ThreadConfiguratorNode::ThreadConfiguratorNode(const rclcpp::NodeOptions & options)
-: Node("thread_configurator_node", options)
+: Node("thread_configurator_node", options),
+  config_file_([this]() {
+    rcl_interfaces::msg::ParameterDescriptor desc;
+    desc.read_only = true;
+    return this->declare_parameter<std::string>("config_file", "", desc);
+  }()),
+  default_domain_id_(agnocast_cie_thread_configurator::get_default_domain_id())
 {
-  const auto config_file = this->declare_parameter<std::string>("config_file", "");
-  if (config_file.empty()) {
+  if (config_file_.empty()) {
     throw std::runtime_error(
       "'config_file' parameter must be provided with a valid YAML file path.");
   }
 
   YAML::Node yaml;
   try {
-    yaml = YAML::LoadFile(config_file);
+    yaml = YAML::LoadFile(config_file_);
   } catch (const std::exception & e) {
-    throw std::runtime_error("Error reading the YAML file '" + config_file + "': " + e.what());
+    throw std::runtime_error("Error reading the YAML file '" + config_file_ + "': " + e.what());
   }
 
   validate_hardware_info(yaml);
   validate_rt_throttling(yaml);
 
-  RCLCPP_INFO(this->get_logger(), "Loaded config from: %s", config_file.c_str());
-
-  size_t default_domain_id = agnocast_cie_thread_configurator::get_default_domain_id();
+  RCLCPP_INFO(this->get_logger(), "Loaded config from: %s", config_file_.c_str());
 
   agnocast_cie_thread_configurator::parse_yaml(
-    yaml, default_domain_id, callback_group_configs_, non_ros_thread_configs_);
+    yaml, default_domain_id_, callback_group_configs_, non_ros_thread_configs_);
 
   unapplied_num_.store(
     static_cast<int>(callback_group_configs_.size() + non_ros_thread_configs_.size()));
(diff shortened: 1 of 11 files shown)


src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
```diff
@@ -486,3 +504,156 @@ void ThreadConfiguratorNode::non_ros_thread_callback(
...
+
+  RCLCPP_INFO(this->get_logger(), "Reapplying configuration from: %s", config_file_.c_str());
+
+  // Phase 1: Parse the YAML.
+  YAML::Node yaml;
+  try {
+    yaml = YAML::LoadFile(config_file_);
+  } catch (const std::exception & e) {
+    response->success = false;
+    response->error_message = std::string("YAML parse error: ") + e.what();
```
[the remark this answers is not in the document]
atsushi421: Done in 5f18c006 - both YAML parse and validation error_message branches now include the config_file path.

src/agnocast_e2e_test/test/functional/test_thread_configurator_reapply_launch.py
```diff
@@ -0,0 +1,301 @@
...
+        stderr=subprocess.STDOUT,
+    )
+    publisher_proc = subprocess.Popen(
+        [publisher_exe],
+        stdout=subprocess.DEVNULL,
+        stderr=subprocess.DEVNULL,
+    )
+
+    time.sleep(5)
+
```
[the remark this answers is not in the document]
atsushi421: Leaving as-is. The literal suggestion (poll for the YAML file) does not fit here: prerun_node only writes template.yaml on shutdown (after SIGINT, in dump_yaml_config), so the file is absent during steady-state by design. A polling alternative would have to scrape the prerun log for "Received CallbackGroupInfo" markers, which adds its own brittleness. The 5s wait is bounded and only used by the prerun fixture (not the test assertions themselves), and the existing fail-fast check at the end raises with the prerun log content if no YAML is produced - so a flake would surface as a clear error rather than a silent assertion miss. Happy to revisit if CI flake data shows this is actually unstable.

src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
```diff
@@ -246,9 +256,12 @@ void ThreadConfiguratorNode::print_all_unapplied()
 
   RCLCPP_WARN(this->get_logger(), "Following non-ROS threads are not yet configured");
 
-  for (auto & config : non_ros_thread_configs_) {
-    if (!config.applied) {
-      RCLCPP_WARN(this->get_logger(), "  - %s", config.thread_str.c_str());
+  {
+    std::lock_guard<std::mutex> lk(non_ros_state_mutex_);
```
kobayu858:
```
  // - print_all_unapplied(): called only after stop() + executor return, so
  //   reads need no lock.
```

The header comment explicitly states that no lock is needed here, yet the implementation acquires one anyway. Is this intentional defensive programming?

src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
```diff
@@ -486,3 +501,120 @@ void ThreadConfiguratorNode::non_ros_thread_callback(
...
+
+  for (auto & cfg : callback_group_configs_) {
+    std::string key = std::to_string(cfg.domain_id) + ":" + cfg.thread_str;
+    if (cfg.thread_id == -1) {
+      response->skipped_callback_groups.push_back(std::move(key));
+      continue;
+    }
+    if (issue_syscalls(cfg)) {
+      response->applied_callback_groups.push_back(std::move(key));
+      if (!cfg.applied) {
```
kobayu858: Since all entries restart with `applied=false` on reapply, this condition should always be true. What is the intent here?

src/agnocast_e2e_test/test/functional/test_thread_configurator_reapply_launch.py
```diff
@@ -0,0 +1,360 @@
...
+        )
+        self.assertEqual(
+            outcomes_total,
+            len(cfg['callback_groups']),
+            'reapply must visit every callback_group entry exactly once',
+        )
+        self.assertGreaterEqual(
+            len(response.applied_callback_groups), 1,
+            'modified callback_group must appear in applied_callback_groups',
+        )
```
kobayu858: `assertGreaterEqual(len(applied), 1)` passes whenever any callback group is applied, not necessarily first (the modified one). If first lands in` skipped_callback_groups `(e.g., its thread hasn't announced yet) while another entry is in applied, this test silently passes despite not exercising what the error message claims.

kobayu858 (changes requested):

src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
atsushi421: Done in 418b59a3 - removed the redundant lock in print_all_unapplied() so the code matches the documented invariant. The function is only called after node->stop() and the executor returns (main.cpp:91-94), so no concurrent access can occur.

src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
atsushi421: You are right - it was dead defensive code. The service handler shares the SingleThreadedExecutor with callback_group_callback, so no concurrent update to cfg.applied is possible on this branch, and every entry starts with applied=false after the swap. Removed the guard in 418b59a3. The analogous check on the non-ROS branch is kept because the listener runs on its own thread and can race between the swap-block lock release and the apply-block lock acquire.

src/agnocast_e2e_test/test/functional/test_thread_configurator_reapply_launch.py
atsushi421: Good catch. Tightened in 418b59a3: the assertion now checks that the modified entry's specific key (`{domain_id}:{id}`) appears in applied_callback_groups, so the test fails if 'first' lands in skipped_callback_groups even when an unrelated entry was applied.

kobayu858 (approved):

PR #1319, opened 2026-05-11
fix(sample_application): add agnocast_ prefix to sample app and thread_configurator

Koichi98:
## Description
Rename shared library targets that get installed to `/opt/ros/jazzy/lib/` so they don't collide with stock ROS 2 Jazzy packages.

### Issue
Installing the `ros-jazzy-agnocast-sample-application` deb fails because `liblistener_component.so` is already shipped by `ros-jazzy-composition`:

  ```
  dpkg: error processing archive ros-jazzy-agnocast-sample-application_...deb (--unpack):
   trying to overwrite '/opt/ros/jazzy/lib/liblistener_component.so',
   which is also in package ros-jazzy-composition
  ```

  Shared libraries are installed flat under `/opt/ros/jazzy/lib/`, so two packages producing `lib<same-name>.so` conflict on deb install.

  ## Changes

  Added `agnocast_` prefix to CMake library targets installed via `LIBRARY DESTINATION lib`:

  **`agnocast_sample_application`**
  - `listener_component` → `agnocast_listener_component` (confirmed collision with `ros-jazzy-composition`)
  - `cie_listener_component` → `agnocast_cie_listener_component` (defensive)
  - `no_rclcpp_listener_component` → `agnocast_no_rclcpp_listener_component` (defensive)

  **`agnocast_cie_thread_configurator`**
  - `thread_configurator` → `agnocast_thread_configurator` (defensive; generic name)

## How was this PR tested?

- [ ] Autoware (required)
- [x] sample application

**Important notes:**

### DIFF
--- a/src/agnocast_cie_thread_configurator/CMakeLists.txt
+++ b/src/agnocast_cie_thread_configurator/CMakeLists.txt
@@ -11,13 +11,13 @@ find_package(rclcpp REQUIRED)
 find_package(agnocast_cie_config_msgs REQUIRED)
 find_package(yaml-cpp REQUIRED)
 
-add_library(thread_configurator SHARED
+add_library(agnocast_thread_configurator SHARED
   src/util.cpp
   src/non_ros_thread_ipc.cpp
   src/thread_config.cpp
 )
-ament_target_dependencies(thread_configurator rclcpp agnocast_cie_config_msgs)
-target_link_libraries(thread_configurator yaml-cpp)
+ament_target_dependencies(agnocast_thread_configurator rclcpp agnocast_cie_config_msgs)
+target_link_libraries(agnocast_thread_configurator yaml-cpp)
 
 add_executable(
   thread_configurator_node
@@ -26,7 +26,7 @@ add_executable(
   src/prerun_node.cpp
 )
 ament_target_dependencies(thread_configurator_node rclcpp agnocast_cie_config_msgs)
-target_link_libraries(thread_configurator_node yaml-cpp thread_configurator)
+target_link_libraries(thread_configurator_node yaml-cpp agnocast_thread_configurator)
 
 target_include_directories(thread_configurator_node PRIVATE
   ${CMAKE_CURRENT_SOURCE_DIR}/include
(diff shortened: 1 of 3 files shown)


atsushi421 (commented):
The PR description states the goal is to prevent current and future `.so` filename collisions under `/opt/ros/<distro>/lib/`, and the renames for `cie_listener_component` / `no_rclcpp_listener_component` / `thread_configurator` are explicitly called "defensive". The same rationale applies cleanly to `src/agnocast_e2e_test/CMakeLists.txt`, which installs seven SHARED libraries flat under `lib/` with very generic names: `test_talker_component`, `test_listener_component`, `test_taker_component`, `test_ros2_listener_component`, `test_ros2_talker_component`, `test_cie_publisher_component`, `test_cie_subscription_component`. `test_listener_component` and `test_talker_component` are especially close to the stock `listener_component` / `talker_component` names that triggered the original collision.

Consider either applying the same `agnocast_` prefix to those targets (and the matching `install(TARGETS ...)` / `ament_export_targets(export_...)` lines), or noting in the PR description that `agnocast_e2e_test` is intentionally out of scope (e.g. because it isn't released as a deb).

atsushi421:
Plus, could you run the `bash scripts/test/run_requires_kernel_module_tests.bash`?

Koichi98:
@atsushi421 
> Consider either applying the same agnocast_ prefix to those targets (and the matching install(TARGETS ...) / ament_export_targets(export_...) lines), 

Added `agnocast_` prefix to e2e_tests too. I didn't add to these files intentionally, since it won't be distributed by apt package. However, it doesn't have any bad effect so I just added it now!

Koichi98:
@atsushi421 
> Plus, could you run the bash scripts/test/run_requires_kernel_module_tests.bash?

Done! All tests have passed.

atsushi421:
@Koichi98 
Renaming the `thread_configurator` package to `agnocast_thread_configurator` seems to be breaking changes, so do we need to preserve the backward compatibility?

Koichi98:
@atsushi421 
The package name isn't changed. Other than directly writing `target_link_libraries(foo agnocast_cie_thread_configurator::thread_configurator)`, which is not normally used, nothing should be affected. That said, it is possible for a user to use it. Do you think we should keep backward compatibility? (Strictly speaking we might need backward compatibility shim)

atsushi421:
> @atsushi421 The package name isn't changed. Other than directly writing `target_link_libraries(foo agnocast_cie_thread_configurator::thread_configurator)`, which is not normally used, nothing should be affected. That said, it is possible for a user to use it. Do you think we should keep backward compatibility? (Strictly speaking we might need backward compatibility shim)

Sorry, I had misunderstood. Because this PR doesn't affect current Autoware implementation, I think we don't need to keep backward compatibility :+1:

atsushi421 (approved):

PR #1333, opened 2026-05-14
fix(cie): disable unnecessary options for cie client node

Koichi98:
## Description
Disable `start_parameter_services` / `start_parameter_event_publisher` / `enable_rosout` on the per-process CIE helper nodes (`create_rclcpp_client_publisher` / `create_agnocast_client_publisher`). These helpers only host a single `CallbackGroupInfo` publisher, so the auto-created parameter services / event publisher / rosout publisher are pure startup overhead with no functional value.

For `agnocast::Node` the impact is more than cosmetic: each auto-created parameter service becomes an `agnocast::Service`, which the bridge proxies (Service + Client per op × 6 = 12 extra child threads per helper) and floods `/agnocast_cie_thread_configurator/callback_group_info` with proxy entries. This also caused `agnocast_e2e_test` tests `TestComponentContainerCIE` / `TestCallbackIsolatedExecutor` to fail with `AssertionError: 4 != 2 : Expected exactly 2 'Received CallbackGroupInfo:' messages`.

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

**Important notes:**

### DIFF
--- a/src/agnocastlib/src/cie_client_utils.cpp
+++ b/src/agnocastlib/src/cie_client_utils.cpp
@@ -88,6 +88,9 @@ create_rclcpp_client_publisher()
   // Disable global arguments so that global "__node" remapping cannot override this name
   // and cause duplicate node names for these per-client nodes.
   options.use_global_arguments(false);
+  options.start_parameter_services(false);
+  options.start_parameter_event_publisher(false);
+  options.enable_rosout(false);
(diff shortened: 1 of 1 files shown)


atsushi421 (commented):
@Koichi98 
Thank you! Could you add the `options.use_clock_thread(false);`?

atsushi421 (approved):

PR #1332, opened 2026-05-14
test(bridge): add `--bridge-only` option and cleanup bridge process in `e2e_test_2to2.bash`

ruth561:
### Summary
This PR adds a performance bridge cleanup process and a `--bridge-only` option to `e2e_test_2to2.bash`. This is a developer-friendly improvement to make E2E testing easier and more efficient.

### Background & Problems
* **Manual Cleanup is Tedious**: To test changes in the performance bridge implementation, we currently have to manually stop existing bridge processes and sometimes delete the `agnocast_bridge_plugins/` directory and rebuild it. This is troublesome and easy to forget.
* **Too Many Test Cases**: We need a way to run only the test cases that involve the bridge process to save time.

### Solution
* **Automated Cleanup**: Added an automatic cleanup process for the performance bridge in `e2e_test_2to2.bash`.
* **New Test Option**: Added the `--bridge-only` option to filter and run only the test cases that use the bridge. This reduces the total number of test cases during specific checks.

## Related links
This PR improves the development workflow related to #1330

## How was this PR tested?

- [ ] Autoware (required)
- [ ] `bash scripts/test/e2e_test_1to1.bash` (required)
- [x] `bash scripts/test/e2e_test_2to2.bash` (required)
- [ ] kunit tests (required when modifying the kernel module)
- [ ] `bash scripts/test/run_requires_kernel_module_tests.bash` (required)
- [ ] sample application

### Supplement
I verified the behavior by running the following commands:

* `bash scripts/test/e2e_test_2to2.bash` (Plain)
* `bash scripts/test/e2e_test_2to2.bash -b performance`
  * Verified that it correctly cleans up the existing performance bridge process.
  * Verified that it generates and builds `agnocast_bridge_plugins` if it does not exist.
* `bash scripts/test/e2e_test_2to2.bash -b standard`
* `AGNOCAST_BRIDGE_MODE=performance bash scripts/test/e2e_test_2to2.bash`

## Version Update Label (Required)

Please add **exactly one** of the following labels to this PR:

- `need-major-update`: User API breaking changes
- `need-minor-update`: Internal API breaking changes (heaphook/kmod/agnocastlib compatibility)
- `need-patch-update`: Bug fixes and other changes

**Important notes:**

- If you need `need-major-update` or `need-minor-update`, please include this in the PR title as well.
  - Example: `fix(foo)[needs major version update]: bar` or `feat(baz)[needs minor version update]: qux`
[...]

Issue #1330, opened 2026-05-14
perf(bridge): optimize the build performance for agnocast_bridge_plugins


ruth561:
### Background

Currently, `agnocast_bridge_plugins` takes too long to build. To provide factory functions like `create_r2a_pubsub_bridge()`, we compile and create separate `.so` files for every message data type. 

### Objective

Significantly reduce the build time of the bridge plugins.

### Proposed Changes

- [ ] Consolidate multiple `.so` files into a single `.so` file.
- [ ] Reduce code duplication as much as possible.
- [ ] Enable `UNITY_BUILD`.
- [ ] Use `ld.lld` instead of `ld` (if available).
- [ ] Replace `Publisher` and `Subscriber` templates with `GenericPublisher` and `GenericSubscriber` and move message (de)serialization into the bridge process.

### How this reduces build time
* **Replacing Templates:** Using `GenericPublisher` and `GenericSubscriber` prevents the compiler from generating massive amounts of code for every single message type. This addresses the root cause of the slow compilation.
* **Consolidating `.so` & UNITY_BUILD:** Reduces the overhead of compiling and linking many separate files.
* **Using `ld.lld`:** Provides a much faster linking phase compared to the default `ld`.

Koichi98:
@ruth561 
I've checked that the compilation time reduces a lot by merging #1336 and #1337, and also confirmed that it successfully passes e2e tests. Thank you so much.

PR #1336, opened 2026-05-15
refactor(bridge-plugins): consolidate .so and enable unity/lld build

ruth561:
### Summary

This PR consolidates multiple `.so` libraries (previously generated for each message type by `ros2 agnocast generate-bridge-plugins`) into a single shared library: `libagnocast_bridge_plugins.so`.

> [!NOTE]
> Please review #1330 before checking this PR.

**Key Changes:**
- [x] Consolidate multiple .so files into a single .so file. 
- [x] Enable UNITY_BUILD.
- [x] Use ld.lld instead of ld (if available).

#### Linker Update (Prioritizing `lld`)
* **Problem:** Because `.so` files are now consolidated, using the default `ld` linker (which does not support multi-threading) might increase link time.
* **Solution:** This PR configures the build system to **prefer `lld` over `ld`** if it is installed in the environment to speed up the linking process.
* **Fallback:** If `lld` is not available, it automatically falls back to the default linker.

#### `.so` File Structure
* **Before:** The `ros2 agnocast generate-bridge-plugins` command generated `libagnocast_bridge_plugins_<msg_name>.so` for each message type. Functions (e.g., `create_r2a_pubsub_bridge`) were defined inside each corresponding `.so` file.
* **After:** A single, all-inclusive `libagnocast_bridge_plugins.so` is generated. To prevent conflicts, the message name is now added as a suffix to the function names.

#### Enabling UNITY_BUILD
* **Background:** It is safe to use here because there are no name conflicts for macros or static variables.
* **Effect:** It reduces duplicated build processes, leading to faster build times.

#### Build Time Performance
This change slightly reduces the build time. 

In my environment, the build time decreased from **~2 minutes to ~1.5 minutes**.
* **CPU:** Intel Core Ultra 9 285K
* **Memory:** 32GiB
* **Workspace:** underlay + agnocast overlay

## Related links

* #1330
* #1332 

## How was this PR tested?

- [ ] Autoware (required)
- [x] `bash scripts/test/e2e_test_2to2.bash` (required) with `-b performance` option (#1332)
- [ ] sample application

**Important notes:**

### DIFF
--- a/src/agnocastlib/src/bridge/performance/agnocast_performance_bridge_loader.cpp
+++ b/src/agnocastlib/src/bridge/performance/agnocast_performance_bridge_loader.cpp
@@ -71,12 +71,10 @@ std::string PerformanceBridgeLoader::convert_type_to_snake_case(const std::strin
   return result;
 }
 
-std::vector<std::string> PerformanceBridgeLoader::generate_library_paths(
-  const std::string & snake_type, bool is_service)
+std::vector<std::string> PerformanceBridgeLoader::generate_library_paths()
 {
   std::vector<std::string> paths;
-  const std::string lib_name =
-    (is_service ? "libservice_bridge_plugin_" : "libpubsub_bridge_plugin_") + snake_type + ".so";
+  const std::string lib_name = "libagnocast_bridge_plugins.so";
 
   // 1. Check environment variable AGNOCAST_BRIDGE_PLUGINS_PATH (colon-separated)
   const char * env_path = std::getenv("AGNOCAST_BRIDGE_PLUGINS_PATH");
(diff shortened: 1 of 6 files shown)


src/agnocastlib/src/bridge/performance/agnocast_performance_bridge_loader.cpp
```diff
@@ -71,12 +71,10 @@ std::string PerformanceBridgeLoader::convert_type_to_snake_case(const std::strin
...
 }
 
-std::vector<std::string> PerformanceBridgeLoader::generate_library_paths(
-  const std::string & snake_type, bool is_service)
+std::vector<std::string> PerformanceBridgeLoader::generate_library_paths()
 {
   std::vector<std::string> paths;
-  const std::string lib_name =
-    (is_service ? "libservice_bridge_plugin_" : "libpubsub_bridge_plugin_") + snake_type + ".so";
+  const std::string lib_name = "libagnocast_bridge_plugins.so";
```
Koichi98, line 77: nit: `generate_library_paths()` now returns a constant value. We can consider caching the library paths and handle (e.g., as member variables) to avoid redundant calls in `get_bridge_factory_symbol()`.  There are some other operations in `generate_library_paths` so we can handle this in the separate PR.

Koichi98 (approved):

PR #1337, opened 2026-05-15
perf(bridge-plugins): make bridge nodes use generic publisher and subscriber

ruth561:
## Description

This PR is a follow-up to #1336.

### Summary

This PR replaces `rclcpp::Publisher<T>` and `rclcpp::Subscriber<T>` used in the performance pub/sub bridge with `rclcpp::GenericPublisher` and `rclcpp::GenericSubscriber`.

> [!NOTE]
> Please review #1330 before checking this PR.

**Key Changes:**
- [x] Reduce code duplication as much as possible.
- [x] Replace Publisher and Subscriber templates with GenericPublisher and GenericSubscriber and move message (de)serialization into the bridge process.

### Background

Previously, compiling `agnocast_bridge_plugins` (generated by the `ros2 agnocast generate-bridge-plugins` command) took too much time. The bottleneck was the template instantiation of `rclcpp::Publisher` and `rclcpp::Subscriber`. To reduce the build time, this PR replaces them with type-erased generic ones.

#### Build Time Performance

This change **significantly** reduces the build time. 
In my environment, the build time decreased from **~1.5 minutes to ~16 seconds**.
* **CPU:** Intel Core Ultra 9 285K
* **Memory:** 32GiB
* **Workspace:** underlay + agnocast overlay

#### Runtime Performance

Runtime performance is slightly lower when using `GenericPublisher` and `GenericSubscriber` compared to the templated ones. However, for this bridge, maximum runtime performance is no longer the main priority. Considering the trade-off, we should prioritize build-time performance over runtime performance.

## Related links

#1330 
#1336

## How was this PR tested?

- [ ] Autoware (required)
- [x] `bash scripts/test/e2e_test_2to2.bash` (required) with `-b performance` option (#1332)
- [ ] sample application

**Important notes:**


src/ros2agnocast/ros2agnocast/templates/pubsub_bridge_plugin.cpp.em
```diff
@@ -23,50 +29,41 @@ extern "C" PerformancePubsubBridgeResult create_r2a_pubsub_bridge_@(snake_type_n
...
-    topic_name,
-    sub_qos,
-    [agno_pub](const @(cpp_type)::ConstSharedPtr msg) {
+  return create_r2a_generic_bridge(
+    node, topic_name, sub_qos, "@(msg_type)",
+    [agno_pub](std::shared_ptr<rclcpp::SerializedMessage> serialized_msg) {
+      static const rclcpp::Serialization<MsgT> serialization;
       auto loaned_msg = agno_pub->borrow_loaned_message();
-      *loaned_msg = *msg;
+      serialization.deserialize_message(serialized_msg.get(), &(*loaned_msg));
```
Koichi98: This seems to throw when it fails. Could you add try/catch just in case?

ruth561: Thank you for pointing that out! Fixed in [1b2ad83](https://github.com/autowarefoundation/agnocast/pull/1337/commits/1b2ad83eb0a424f287b8c6bdce8a86a61de54710).

Koichi98 (approved):

Issue #1176, opened 2026-03-11
Implement unsupported API for AgnocastOnlyExecutor


Koichi98:
The example is `spin_until_future_complete`. There might be some others.

eyr1n:
# AgnocastOnlyExecutor spin API support plan

## Background

`AgnocastOnlyExecutor` class only exposes a single spin-related API `spin()`. While this is sufficient for the simplest execution path, it does not provide a more flexible scheduling interface.

In practice, the lack of `spin_*` family of APIs makes it difficult to implement executor scheduling patterns such as:

- executing only one callback
- spinning until a future completes
- spinning a specific node, etc.

## Goal

Implement the missing spin-related executor APIs for `AgnocastOnlyExecutor`.

At this point, I am planning to create three pull requests. Each PR is based on the implementation in the previous PR.

### PR 1: Spin API primitives

- [ ] [`is_spinning`](https://github.com/ros2/rclcpp/blob/72c5209bfe59cb9fba52e0befb3b1e72f265d391/rclcpp/include/rclcpp/executor.hpp#L412)
  - returns true if the executor is currently spinning
- [ ] [`spin_once`](https://github.com/ros2/rclcpp/blob/72c5209bfe59cb9fba52e0befb3b1e72f265d391/rclcpp/include/rclcpp/executor.hpp#L367)
  -  collect work once and execute the next available work with timeout
- [ ] [`spin_until_future_complete`](https://github.com/ros2/rclcpp/blob/72c5209bfe59cb9fba52e0befb3b1e72f265d391/rclcpp/include/rclcpp/executor.hpp#L384)
  - spin until the future is complete with timeout

### PR 2: Basic spin_* API

- [ ] [`spin_some`](https://github.com/ros2/rclcpp/blob/72c5209bfe59cb9fba52e0befb3b1e72f265d391/rclcpp/include/rclcpp/executor.hpp#L319)
  - collect work once and execute all available work with timeout
- [ ] [`spin_all`](https://github.com/ros2/rclcpp/blob/72c5209bfe59cb9fba52e0befb3b1e72f265d391/rclcpp/include/rclcpp/executor.hpp#L353)
  - collect and execute work repeatedly until no more work is available

### PR 3: Spin API per node

These APIs are based on Step 1 and Step 2. However, each function adds a node to the executor before spinning, and then removes the node.

[...]

PR #1338, opened 2026-05-15
feat(agnocastlib): Implement spin API primitives for AgnocastOnlyExecutor

eyr1n:
## Description

This PR implements spin API primitives from `rclcpp`.

- [`is_spinning`](https://github.com/ros2/rclcpp/blob/72c5209bfe59cb9fba52e0befb3b1e72f265d391/rclcpp/include/rclcpp/executor.hpp#L412)
  - returns true if the executor is currently spinning
- [`spin_once`](https://github.com/ros2/rclcpp/blob/72c5209bfe59cb9fba52e0befb3b1e72f265d391/rclcpp/include/rclcpp/executor.hpp#L367)
  -  collect work once and execute the next available work with timeout
- [`spin_until_future_complete`](https://github.com/ros2/rclcpp/blob/72c5209bfe59cb9fba52e0befb3b1e72f265d391/rclcpp/include/rclcpp/executor.hpp#L384)
  - spin until the future is complete with timeout

## Related links

#1176 
[`rclcpp::Executor` declaration (`executor.hpp`)](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/executor.hpp)
[`rclcpp::Executor` implementation (`executor.cpp`)](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/src/rclcpp/executor.cpp)

## How was this PR tested?

- [ ] Autoware (required)
- [ ] sample application

**Important notes:**


src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
```diff
@@ -127,6 +134,22 @@ class AgnocastOnlyExecutor
...
+    std::chrono::duration<TimeRepT, TimeT> timeout = std::chrono::duration<TimeRepT, TimeT>(-1))
+  {
+    return spin_until_future_complete_impl(
+      std::chrono::duration_cast<std::chrono::nanoseconds>(timeout),
+      [&future](std::chrono::nanoseconds wait_time) { return future.wait_for(wait_time); });
+  }
+
+  AGNOCAST_PUBLIC
+  bool is_spinning();
 };
```
ruth561, line 185: Could you add comments describing each function to match the other `AGNOCAST_PUBLIC` functions?

src/agnocastlib/src/node/agnocast_only_executor.cpp
```diff
@@ -387,4 +387,89 @@ void AgnocastOnlyExecutor::remove_node(const std::shared_ptr<agnocast::Node> & n
...
+  AgnocastExecutable agnocast_exec;
+  if (get_next_agnocast_executable(agnocast_exec, timeout)) {
+    execute_agnocast_executable(agnocast_exec);
+  }
+}
+
+void AgnocastOnlyExecutor::spin_once(std::chrono::nanoseconds timeout)
+{
+  if (spinning_.exchange(true)) {
+    close(agnocast_fd);
```
ruth561: Maybe we should remove this line? Regarding agnocast's error handling, we usually close agnocast_fd only when the process exits. Since this part throws an exception instead of exiting, the user can catch the exception and recover.
```suggestion
```

src/agnocastlib/src/node/agnocast_only_executor.cpp
```diff
@@ -387,4 +387,89 @@ void AgnocastOnlyExecutor::remove_node(const std::shared_ptr<agnocast::Node> & n
...
+  auto end_time = std::chrono::steady_clock::now();
+  std::chrono::nanoseconds timeout_ns =
+    std::chrono::duration_cast<std::chrono::nanoseconds>(timeout);
+  if (timeout_ns > std::chrono::nanoseconds::zero()) {
+    end_time += timeout_ns;
+  }
+  std::chrono::nanoseconds timeout_left = timeout_ns;
+
+  if (spinning_.exchange(true)) {
+    close(agnocast_fd);
```
ruth561: Maybe we should remove this line? (Please see the other comment about this.)
```suggestion
```

src/agnocastlib/src/node/agnocast_only_executor.cpp
```diff
@@ -387,4 +387,89 @@ void AgnocastOnlyExecutor::remove_node(const std::shared_ptr<agnocast::Node> & n
...
+  }
+
+  int next_exec_timeout_ms;
+  if (timeout < std::chrono::nanoseconds::zero()) {
+    next_exec_timeout_ms = -1;
+  } else {
+    next_exec_timeout_ms = std::chrono::duration_cast<std::chrono::milliseconds>(timeout).count();
+  }
+  AgnocastExecutable agnocast_exec;
+  if (get_next_agnocast_executable(agnocast_exec, timeout)) {
```
ruth561, line 446: Are you trying to use `next_exec_timeout_ms` instead of `timeout_ms`?

src/agnocastlib/src/node/agnocast_only_executor.cpp
```diff
@@ -387,4 +387,89 @@ void AgnocastOnlyExecutor::remove_node(const std::shared_ptr<agnocast::Node> & n
...
+void AgnocastOnlyExecutor::spin_once_impl(std::chrono::nanoseconds timeout)
+{
+  if (epoll_update_tracker_.take_update_request()) {
+    add_callback_groups_from_nodes_associated_to_executor();
+    epoll_manager_->prepare_epoll([this](const rclcpp::CallbackGroup::SharedPtr & group) {
+      return is_callback_group_associated(group);
+    });
+  }
+
+  int next_exec_timeout_ms;
```
ruth561: nit: How about using a simpler name, `timeout_ms`?

src/agnocastlib/test/integration/test_agnocast_only_single_threaded_executor.cpp
```diff
@@ -0,0 +1,94 @@
...
+  promise.set_value(true);
+
+  auto start = std::chrono::steady_clock::now();
+  auto shared_future = future.share();
+  auto ret = executor.spin_until_future_complete(shared_future, 1s);
+  executor.remove_node(this->node, true);
+
+  EXPECT_GT(500ms, (std::chrono::steady_clock::now() - start));
+  EXPECT_EQ(rclcpp::FutureReturnCode::SUCCESS, ret);
+}
```
ruth561, line 461: Thank you for adding the test cases for the newly added functions!
I'd like to improve the test coverage a bit more. Could you add the following cases?

`spin_once`:
- `timeout == -1` (No timeout): Ensure it blocks until terminated (e.g., by `cancel()`).
- `timeout == 0` (Non-blocking): Ensure it returns immediately.
- Timeout behavior: Ensure it properly waits for the timeout when no callback is runnable.
- Single execution: Arrange two callbacks and verify that `spin_once()` executes exactly one of them.
- Exception handling: Ensure it throws an exception if another `spin()` primitive is already running (`is_spinning() == true`).
- State update: Ensure it sets `is_spinning` to true.

`spin_until_future_complete`:
- Could you think of similar test cases based on the ones for `spin_once()`?

Sorry that our previous test coverage wasn't sufficient! But since we can easily generate test cases using Gen-AI nowadays, I'd appreciate it if you could give it a try.

Also, this is completely optional, but it would be great if you could structure the test cases using the AAA (Arrange-Act-Assert) pattern!

ruth561 (changes requested):

src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
```diff
@@ -58,6 +59,12 @@ class AgnocastOnlyExecutor
 
   void add_callback_groups_from_nodes_associated_to_executor();
 
+  virtual rclcpp::FutureReturnCode spin_until_future_complete_impl(
+    std::chrono::nanoseconds timeout,
+    const std::function<std::future_status(std::chrono::nanoseconds wait_time)> & wait_for_future);
```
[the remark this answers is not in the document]
eyr1n: fixed.

src/agnocastlib/src/node/agnocast_only_executor.cpp
```diff
@@ -387,4 +387,89 @@ void AgnocastOnlyExecutor::remove_node(const std::shared_ptr<agnocast::Node> & n
...
+  const std::function<std::future_status(std::chrono::nanoseconds wait_time)> & wait_for_future)
+{
+  std::future_status status = wait_for_future(std::chrono::seconds(0));
+  if (status == std::future_status::ready) {
+    return rclcpp::FutureReturnCode::SUCCESS;
+  }
+
+  auto end_time = std::chrono::steady_clock::now();
+  std::chrono::nanoseconds timeout_ns =
+    std::chrono::duration_cast<std::chrono::nanoseconds>(timeout);
```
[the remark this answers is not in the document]
eyr1n: fixed.

src/agnocastlib/test/integration/test_agnocast_only_single_threaded_executor.cpp
```diff
@@ -0,0 +1,94 @@
+#include <agnocast/agnocast.hpp>
+#include <agnocast/node/agnocast_only_executor.hpp>
+
+#include <gtest/gtest.h>
+
+#include <chrono>
```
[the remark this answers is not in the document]
eyr1n: fixed.

src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
```diff
@@ -127,6 +134,22 @@ class AgnocastOnlyExecutor
...
   AGNOCAST_PUBLIC
   void remove_node(const std::shared_ptr<agnocast::Node> & node, bool notify = true);
+
+  AGNOCAST_PUBLIC
+  virtual void spin_once(std::chrono::nanoseconds timeout = std::chrono::nanoseconds(-1));
+
+  template <typename FutureT, typename TimeRepT = int64_t, typename TimeT = std::milli>
+  rclcpp::FutureReturnCode spin_until_future_complete(
+    const FutureT & future,
+    std::chrono::duration<TimeRepT, TimeT> timeout = std::chrono::duration<TimeRepT, TimeT>(-1))
```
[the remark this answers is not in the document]
eyr1n: fixed.

src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
eyr1n: I added comments from rclcpp::Executor.

src/agnocastlib/src/node/agnocast_only_executor.cpp
eyr1n: fixed.

src/agnocastlib/src/node/agnocast_only_executor.cpp
eyr1n: That's my mistake, thanks.

eyr1n:
I updated the PR based on the review comments.

src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
```diff
@@ -4,10 +4,13 @@
...
 #include "agnocast/agnocast_public_api.hpp"
 #include "rclcpp/callback_group.hpp"
+#include "rclcpp/future_return_code.hpp"
 #include "rclcpp/node_interfaces/node_base_interface.hpp"
 #include "rcpputils/thread_safety_annotations.hpp"
 
 #include <atomic>
+#include <chrono>
+#include <future>
 #include <list>
```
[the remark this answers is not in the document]
eyr1n: fixed.

src/agnocastlib/src/node/agnocast_only_executor.cpp
```diff
@@ -387,4 +393,91 @@ void AgnocastOnlyExecutor::remove_node(const std::shared_ptr<agnocast::Node> & n
...
 
+rclcpp::FutureReturnCode AgnocastOnlyExecutor::spin_until_future_complete_impl(
+  std::chrono::nanoseconds timeout,
+  const std::function<std::future_status(std::chrono::nanoseconds wait_time)> & wait_for_future)
+{
+  std::future_status status = wait_for_future(std::chrono::seconds(0));
+  if (status == std::future_status::ready) {
+    return rclcpp::FutureReturnCode::SUCCESS;
+  }
+
```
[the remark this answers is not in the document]
eyr1n: Fixed by updating the while loop condition.

src/agnocastlib/src/node/agnocast_only_executor.cpp
```diff
@@ -387,4 +393,91 @@ void AgnocastOnlyExecutor::remove_node(const std::shared_ptr<agnocast::Node> & n
...
+      return is_callback_group_associated(group);
+    });
+  }
+
+  int timeout_ms;
+  if (timeout < std::chrono::nanoseconds::zero()) {
+    timeout_ms = -1;
+  } else {
+    timeout_ms = std::chrono::ceil<std::chrono::milliseconds>(timeout).count();
+  }
```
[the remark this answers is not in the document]
eyr1n: Fixed.

src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
```diff
@@ -140,6 +154,33 @@ class AgnocastOnlyExecutor
   /// @param notify If true, wake the executor so it picks up the change immediately.
   AGNOCAST_PUBLIC
   void remove_node(const std::shared_ptr<agnocast::Node> & node, bool notify = true);
+
+  /// Collect work once and execute the next available work, optionally within a duration.
+  /// @param timeout timeout The maximum amount of time to spend waiting for work. `-1` is
+  /// potentially block forever waiting for work.
```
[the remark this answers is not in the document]
eyr1n: Fixed.

eyr1n:
I updated the PR based on the review comments.

src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
```diff
@@ -140,6 +155,33 @@ class AgnocastOnlyExecutor
   /// @param notify If true, wake the executor so it picks up the change immediately.
   AGNOCAST_PUBLIC
   void remove_node(const std::shared_ptr<agnocast::Node> & node, bool notify = true);
+
+  /// Collect work once and execute the next available work, optionally within a duration.
+  /// @param timeout The maximum amount of time to spend waiting for work. `-1` blocks forever
+  /// waiting for work.
+  AGNOCAST_PUBLIC
+  virtual void spin_once(std::chrono::nanoseconds timeout = std::chrono::nanoseconds(-1));
```
Koichi98, line 164: `spin_once()` / `spin_until_future_complete()` are inherited by `AgnocastOnlyCallbackIsolatedExecutor`, which doesn't override them. There, `spin_once()` reaches the new `wait_for_work()` -> `add_callback_groups_from_nodes_associated_to_executor()` and claims the callback groups for the parent (`agnocast_only_executor.cpp:471`, `:283`) -- exactly what `CIE::add_node()` documents it must not do (`agnocast_only_callback_isolated_executor.hpp:70-73`). A later `spin()` then hits the double-add check in the child's `add_callback_group()` and calls `exit(EXIT_FAILURE)` (`:168-172`). This path was unreachable from CIE before this PR.

Proper CIE support (and whether `spin_once()` can be supported there at all) deserves its own PR. For now, could you override both in CIE to `throw std::runtime_error("... is not supported in AgnocastOnlyCallbackIsolatedExecutor")`, following `node_topics.cpp:40-65`? Overriding `spin_once()` and `spin_until_future_complete_impl()` makes the whole `wait_for_work()` path unreachable from CIE.

src/agnocastlib/src/node/agnocast_only_executor.cpp
```diff
@@ -387,4 +393,97 @@ void AgnocastOnlyExecutor::remove_node(const std::shared_ptr<agnocast::Node> & n
...
+}
+
+bool AgnocastOnlyExecutor::is_spinning()
+{
+  return spinning_;
+}
+
+void AgnocastOnlyExecutor::wait_for_work(std::chrono::nanoseconds timeout)
+{
+  if (epoll_update_tracker_.take_update_request()) {
```
Koichi98: This is now the third copy of the same block (`agnocast_only_single_threaded_executor.cpp:35-40`, `agnocast_only_multi_threaded_executor.cpp:54-59`). We cannot just delete the copies in `spin()`: `get_next_agnocast_executable()` returns early at `:93-95` when a ready executable exists, so `wait_for_work()` can be skipped. Could we extract it into a small private method and call it from all three places, so they cannot drift?

src/agnocastlib/src/node/agnocast_only_executor.cpp
```diff
@@ -104,6 +104,12 @@ bool AgnocastOnlyExecutor::get_next_agnocast_executable(
   return get_next_ready_agnocast_executable(agnocast_executable);
 }
 
+bool AgnocastOnlyExecutor::get_next_agnocast_executable(
```
Koichi98: This `int` overload only forwards to the `nanoseconds` one, and it has only two callers (`agnocast_only_single_threaded_executor.cpp:43`, `agnocast_only_multi_threaded_executor.cpp:70`). Could we remove it and pass `std::chrono::milliseconds(next_exec_timeout_ms_)` there instead? Two overloads plus a default argument are harder to read.

src/agnocastlib/test/integration/test_agnocast_only_single_threaded_executor.cpp
```diff
@@ -0,0 +1,461 @@
...
+
+  [[maybe_unused]] auto timer1 =
+    this->node->create_wall_timer(1ms, [&]() { callback_count.fetch_add(1); });
+
+  [[maybe_unused]] auto timer2 =
+    this->node->create_wall_timer(1ms, [&]() { callback_count.fetch_add(1); });
+
+  executor.add_node(this->node);
+
+  ASSERT_TRUE(wait_until([&]() { return callback_count.load() == 0; }, 10ms));
```
Koichi98: `wait_until()` returns true as soon as the predicate is true (`:22-35`), and `callback_count` is 0 at this point. So this line always passes and checks nothing. Did you mean `ASSERT_EQ(0, callback_count.load())`?

src/agnocastlib/test/integration/test_agnocast_only_single_threaded_executor.cpp
```diff
@@ -0,0 +1,461 @@
...
+  // Act / Assert
+  EXPECT_THROW(executor.spin_once(0ns), std::runtime_error);
+
+  // Cleanup
+  executor.cancel();
+  spinner.join();
+  executor.remove_node(this->node, true);
+}
+
+TEST_F(AgnocastOnlySingleThreadedExecutorTest, test_spin_once_sets_is_spinning)
```
Koichi98: This test is the same as `test_spin_once_with_no_timeout_blocks_until_cancel` (`:114`), and its assertions are a subset of that one. The same applies to `test_spin_until_future_complete_sets_is_spinning` (`:426`) vs `:279`. Could we drop the two `_sets_is_spinning` tests? Also, `executor.remove_node()` in each Cleanup is not needed: the executor is a local, and its destructor already clears the flags (`agnocast_only_executor.cpp:56-88`).

src/agnocastlib/CMakeLists.txt
```diff
@@ -241,6 +241,18 @@ if(BUILD_TESTING)
     TIMEOUT 300
   )
 
+  # AgnocastOnly executor integration test (requires kernel module, no mock)
+  ament_add_gmock(test_integration_agnocast_only_executor_${PROJECT_NAME}
```
Koichi98: A few small things here:

- The target is `..._agnocast_only_executor_...` but the file is `test_agnocast_only_single_threaded_executor.cpp`. Could we make them match?
- `agnocast_cie_config_msgs` (`:248-249`) looks copied from the CIE test. That test includes the message header directly; this one does not, and the `agnocast` target already depends on it (`:59`).
- `GTEST_DEATH_TEST_STYLE=threadsafe` is set, but there is no death test in this file.
- `TIMEOUT 60` may be tight for 14 tests that each call `agnocast::init()` / `shutdown()`. The CIE test uses 120.

src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
```diff
@@ -140,6 +155,33 @@ class AgnocastOnlyExecutor
...
+
+  /// Spin (blocking) until the future is complete, it times out waiting, or rclcpp is interrupted.
+  /// @param future The future to wait on. If this function returns SUCCESS, the future can be
+  /// accessed without blocking (though it may still throw an exception).
+  /// @param timeout Optional timeout parameter, which gets passed to Executor::spin_node_once. `-1`
```
Koichi98: `Executor::spin_node_once` does not exist in `AgnocastOnlyExecutor`. This looks copied from the rclcpp doc, so it may confuse readers.

src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
```diff
@@ -140,6 +155,33 @@ class AgnocastOnlyExecutor
...
+  /// is block forever, `0` is non-blocking.
+  /// @return The return code, one of `SUCCESS`, `INTERRUPTED`, or `TIMEOUT`.
+  template <typename FutureT, typename TimeRepT = int64_t, typename TimeT = std::milli>
+  rclcpp::FutureReturnCode spin_until_future_complete(
```
Koichi98, line 173: nit: `spin_once()` (`:162`) and `is_spinning()` (`:183`) have `AGNOCAST_PUBLIC`, but this public template does not. Could we add it? The macro is used as a filter for tooling.

src/agnocastlib/src/node/agnocast_only_executor.cpp
```diff
@@ -387,4 +393,97 @@ void AgnocastOnlyExecutor::remove_node(const std::shared_ptr<agnocast::Node> & n
...
+  if (spinning_.exchange(true)) {
+    throw std::runtime_error("spin_once() called while already spinning");
+  }
+  RCPPUTILS_SCOPE_EXIT(this->spinning_.store(false););
+  spin_once_impl(timeout);
+}
+
+bool AgnocastOnlyExecutor::is_spinning()
+{
+  return spinning_;
```
Koichi98: nit: `return spinning_.load();` would match the rest of the file (e.g. `:417`).

Koichi98 (commented):
Thanks for this! I checked the new `spin_until_future_complete_impl()` / `spin_once_impl()` against rclcpp (`rclcpp/executor.hpp`) and the logic matches, so the ported part looks good.

I found one issue that I think should be fixed before merge (the `AgnocastOnlyCallbackIsolatedExecutor` comment below), plus some smaller points inline.

One more thing about test coverage: all tests use a bare `std::promise`, but the main user of `spin_until_future_complete()` is `agnocast::Client`. Its `async_send_request()` promise is set inside the response subscription callback (`agnocast_client.hpp:201`, subscription created at `:207-211`), and `agnocast::Node::create_client()` exists (`agnocast_node.hpp:613`), so "Stage 2 executor + Client" is a supported combination. The existing client tests only use the Stage 1 executor (`test_agnocast_client.cpp:21`), so this combination is not covered anywhere today. Could we add one request/response round-trip test?

src/agnocastlib/src/node/agnocast_only_executor.cpp
```diff
@@ -387,4 +393,97 @@ void AgnocastOnlyExecutor::remove_node(const std::shared_ptr<agnocast::Node> & n
...
+  std::chrono::nanoseconds timeout_left = timeout_ns;
+
+  if (spinning_.exchange(true)) {
+    throw std::runtime_error("spin_until_future_complete() called while already spinning");
+  }
+  RCPPUTILS_SCOPE_EXIT(this->spinning_.store(false););
+
+  while (spinning_.load() && !cancel_requested_.load() && agnocast::ok()) {
+    // Do one item of work.
+    spin_once_impl(timeout_left);
```
Koichi98, line 429: `spin_once()` and `spin_until_future_complete()` default to `timeout = -1`, so this is the first path here that blocks in `epoll_wait()` forever. Every existing spin path polls on purpose -- see the comment in `agnocast_only_multi_threaded_executor.cpp:67-69`.

While blocked, `add_callback_group()` / `add_node()` from another thread is missed, because they ignore `notify`. rclcpp wakes up here via `interrupt_guard_condition_`. Making `notify` actually wake the epoll is the proper fix, but that feels like a separate PR.

src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
eyr1n: Fixed in aac2bef26284832c56fd5101f1a88876e81a2ed2

src/agnocastlib/src/node/agnocast_only_executor.cpp
eyr1n: Fixed in d0f552ef1bcdfaa28a3e38a9119220e043572af8

src/agnocastlib/src/node/agnocast_only_executor.cpp
eyr1n: Fixed in d0f552ef1bcdfaa28a3e38a9119220e043572af8

src/agnocastlib/src/node/agnocast_only_executor.cpp
eyr1n: Fixed in d0f552ef1bcdfaa28a3e38a9119220e043572af8

src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
eyr1n: Fixed in d0f552ef1bcdfaa28a3e38a9119220e043572af8

src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
eyr1n: Fixed in d0f552ef1bcdfaa28a3e38a9119220e043572af8

src/agnocastlib/test/integration/test_agnocast_only_single_threaded_executor.cpp
eyr1n: Fixed in 3ba5343b19a28564fb931b4693879ef7991b96e0

src/agnocastlib/test/integration/test_agnocast_only_single_threaded_executor.cpp
eyr1n: Fixed in 3ba5343b19a28564fb931b4693879ef7991b96e0

src/agnocastlib/CMakeLists.txt
eyr1n: Fixed in 3ba5343b19a28564fb931b4693879ef7991b96e0

PR #1340, opened 2026-05-17
fix: e2e stress workaround

Koichi98:
## Description
 Fix e2e stress tests for Ubuntu 24.04 / Jazzy
  
The old design delayed `ReadyToTest` by the full soak duration, but launch_testing hard-codes a ~15s limit for reaching `ReadyToTest` and aborts otherwise. The previous workaround sed-patched the installed `launch_testing` source at runtime, which is fragile and broke on Jazzy.

Instead:
  - `ReadyToTest` now fires promptly, within the limit.
  - The soak is moved to `setUpClass` (test phase, no limit); `forever=True` nodes keep running during it.
  - Removed the sed workaround from `e2e_test_stress.bash`.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

**Important notes:**

### DIFF
--- a/src/agnocast_e2e_test/test/test_1to1.py
+++ b/src/agnocast_e2e_test/test/test_1to1.py
@@ -1,4 +1,5 @@
 import os
+import time
 import unittest
 
 import launch_testing
@@ -63,7 +64,9 @@ def calc_action_delays(config: dict) -> tuple:
     unit_delay = 1.0
     pub_delay = 0.0 if config['launch_pub_before_sub'] else unit_delay
     sub_delay = 0.01 * EXPECT_INIT_PUB_NUM + unit_delay if config['launch_pub_before_sub'] else 0.0
-    ready_delay = float(TIMEOUT) if TIMEOUT else pub_delay + sub_delay + 10.0
+    # ReadyToTest must fire within launch_testing's ~15s limit; the stress
+    # soak is done in Test1To1.setUpClass.
+    ready_delay = pub_delay + sub_delay + 10.0
     return pub_delay, sub_delay, ready_delay
 
 
@@ -257,6 +260,12 @@ def generate_test_description():
 
 
 class Test1To1(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        # Stress soak: forever=True nodes keep running while we sleep here.
+        if TIMEOUT:
+            time.sleep(float(TIMEOUT))
 
     def test_pub(self, proc_output, test_pub):
         output_text = "".join(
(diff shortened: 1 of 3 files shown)


src/agnocast_e2e_test/test/test_1to1.py
```diff
@@ -63,7 +64,9 @@ def calc_action_delays(config: dict) -> tuple:
     unit_delay = 1.0
     pub_delay = 0.0 if config['launch_pub_before_sub'] else unit_delay
     sub_delay = 0.01 * EXPECT_INIT_PUB_NUM + unit_delay if config['launch_pub_before_sub'] else 0.0
-    ready_delay = float(TIMEOUT) if TIMEOUT else pub_delay + sub_delay + 10.0
+    # ReadyToTest must fire within launch_testing's ~15s limit; the stress
+    # soak is done in Test1To1.setUpClass.
+    ready_delay = pub_delay + sub_delay + 10.0
```
atsushi421, line 69: This formula can silently exceed launch_testing's 15s cap: `sub_delay = 0.01 * EXPECT_INIT_PUB_NUM + 1.0`, so `ready_delay` grows linearly with `pub_qos_depth` and crosses 15.0 around `pub_qos_depth = 400` (with `launch_pub_before_sub` and `pub_transient_local`), re-introducing the exact bug this PR fixes. Consider clamping (`min(pub_delay + sub_delay + 10.0, 14.0)`) or asserting `ready_delay < 15.0` so future config changes fail loudly instead of regressing silently.

src/agnocast_e2e_test/test/test_1to1.py
```diff
@@ -258,6 +261,12 @@ def generate_test_description():
 
 class Test1To1(unittest.TestCase):
 
+    @classmethod
+    def setUpClass(cls):
+        # Stress soak: forever=True nodes keep running while we sleep here.
+        if TIMEOUT:
```
atsushi421: Gating predicates differ from test_2to2.py, which uses `if FOREVER:` (bool). Here `TIMEOUT` is a string-or-None from `os.environ.get`, so `STRESS_TEST_TIMEOUT=0` makes `"0"` truthy and triggers `sleep(0)` while still leaving `FOREVER=True` (publishers never terminate), masking the intent. Either gate on `FOREVER` to match test_2to2.py, or normalize `TIMEOUT` to a float at module scope (e.g., `TIMEOUT = float(_raw) if _raw is not None else 0.0`).

Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1340/commits/ba461a35e4d60177d120257de4d4492ee398b291.

src/agnocast_e2e_test/test/test_1to1.py
Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1340/commits/ba461a35e4d60177d120257de4d4492ee398b291.

atsushi421 (approved):

PR #1341, opened 2026-05-18
refactor(bridge): introduce a builder class for bridge request messages

### LINKED ISSUE #1317: Check snprintf() return values and abort on truncation
clang-tidy reports unchecked `snprintf()` return values in the codebase.

We should check the return value and terminate the program if truncation occurs. This ensures successful `snprintf()` calls as an invariant, simplifying program states and catching buffer size issues early.

bdm-k:
## Description

This PR refactors the bridge request message and its construction pipeline to improve readability, consistency, and maintainability. There are three main changes:

1. Rename `fn_offset`/`fn_offset_reverse` to `fn_offset_r2a`/`fn_offset_a2r` in `BridgeFactoryInfo`. The old names were direction-dependent. The new names make both producer code and consumer code a bit simpler.
2. Introduce `BridgeRequestMsgBuilder` for bridge request message construction, which provides centralized error handling. In relation to this, there is a change in behavior: if a string is truncated by `snprintf()` while constructing a bridge request message, the program (whether it's a normal process or a bridge) terminates. This closes #1317 
3. Removes `symbol_name` from `BridgeFactoryInfo` and adds `in_main_executable` instead. `symbol_name` was used only to determine whether the bridge factories reside in the main executable, so replacing it with a boolean field is more straightforward and memory-efficient.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application


src/agnocastlib/include/agnocast/bridge/agnocast_bridge_utils.hpp
```diff
@@ -67,11 +68,45 @@ bool has_external_ros2_subscriber(const rclcpp::Node * node, const std::string &
...
-  const rclcpp::Logger & logger);
+/// @brief A builder class for creating bridge request messages.
+///
+/// It handles errors such as setting non-existent fields or invalid values for each field, and the
+/// `build_*` member functions return an error reason. However, it does not check whether the
+/// computed message as a whole is valid. It's the caller's responsibility to invoke a correct
+/// set of setters to build a valid message.
+class BridgeRequestMsgBuilder
+{
+  std::variant<MqMsgBridge, MqMsgPerformanceBridge> msg_;
```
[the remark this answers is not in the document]
bdm-k: fixed

src/agnocastlib/src/bridge/agnocast_bridge_utils.cpp
```diff
@@ -232,64 +232,212 @@ bool is_agnocast_service_alive(const std::string & service_name, std::string & r
...
+  standard_msg.factory.fn_offset_a2r = fn_a2r - base_addr;
+
+  return *this;
+}
+
+BridgeRequestMsgBuilder & BridgeRequestMsgBuilder::set_message_type(const char * message_type)
+{
+  std::visit(
+    [this, message_type](auto && msg) {
+      if constexpr (std::is_same_v<std::decay_t<decltype(msg)>, MqMsgPerformanceBridge>) {
```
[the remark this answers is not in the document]
bdm-k: fixed

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_node.hpp
```diff
@@ -343,25 +343,21 @@ void send_standard_pubsub_bridge_request(
...
+                   .set_direction(direction)
+                   .set_is_service(false)
+                   .set_pubsub_target_id(id)
+                   .set_topic_name(topic_name.c_str())
+                   .set_factory(fn_r2a, fn_a2r);
+  auto [msg, reason] = std::move(builder).build_standard_message();
+  if (!reason.empty()) {
+    RCLCPP_ERROR(logger, "Failed to build standard pubsub bridge request: %s", reason.c_str());
+    close(agnocast_fd);
+    exit(EXIT_FAILURE);
```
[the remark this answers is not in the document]
bdm-k: fixed

src/agnocastlib/src/bridge/agnocast_bridge_utils.cpp
```diff
@@ -232,64 +232,212 @@ bool is_agnocast_service_alive(const std::string & service_name, std::string & r
...
+  } else {
+    msg_ = MqMsgPerformanceBridge{};
+  }
+}
+
+BridgeRequestMsgBuilder & BridgeRequestMsgBuilder::fail(const char * format, ...)
+{
+  va_list args;
+  va_start(args, format);
+  int n = vsnprintf(nullptr, 0, format, args);
```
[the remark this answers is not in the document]
bdm-k: fixed

src/agnocastlib/src/bridge/agnocast_bridge_utils.cpp
```diff
@@ -232,64 +232,212 @@ bool is_agnocast_service_alive(const std::string & service_name, std::string & r
...
+  va_end(args);
+
+  std::string buf(n + 1, '\0');
+  va_start(args, format);
+  vsnprintf(buf.data(), n + 1, format, args);
+  va_end(args);
+  buf.resize(n);
+
+  failed_ = true;
```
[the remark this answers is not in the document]
bdm-k: fixed

src/agnocastlib/src/bridge/agnocast_bridge_utils.cpp
```diff
@@ -232,64 +235,223 @@ bool is_agnocast_service_alive(const std::string & service_name, std::string & r
...
+  reason_ = std::move(buf);
+  return *this;
+}
+
+int BridgeRequestMsgBuilder::checked_snprintf(
+  const std::string & member, char * buffer, size_t size, const char * format, ...)
+{
```
kobayu858, line 278: Minor nit: once `failed_` is set to `true`, subsequent calls to `checked_snprintf` still execute. There's no practical harm since the result is discarded, but adding an early return makes the intent explicit:

```cpp
int BridgeRequestMsgBuilder::checked_snprintf(...) {
  if (failed_) return -1;  // already failed, skip
  ...
}
```

src/agnocastlib/src/bridge/agnocast_bridge_utils.cpp
```diff
@@ -232,64 +235,223 @@ bool is_agnocast_service_alive(const std::string & service_name, std::string & r
...
+        NODE_NAME_BUFFER_SIZE, "%s", shadow_node_name);
+    },
+    msg_);
+  return *this;
+}
+
+std::pair<MqMsgBridge, std::string> BridgeRequestMsgBuilder::build_standard_message() &&
+{
+  if (!std::holds_alternative<MqMsgBridge>(msg_)) {
+    fail("Cannot build a standard bridge message from a performance builder");
```
kobayu858: The mismatch check in `build_standard_message()` (and `build_performance_message()`) — calling the wrong `build_*` for the mode — can only happen due to a programming error on the caller's side. Since it's a bug rather than a recoverable runtime condition, I think `assert` would be more appropriate here than returning an error string and letting the caller handle it.

src/agnocastlib/src/bridge/agnocast_bridge_utils.cpp
```diff
@@ -232,64 +235,223 @@ bool is_agnocast_service_alive(const std::string & service_name, std::string & r
...
+{
+  std::visit(
+    [this, topic_name](auto && msg) {
+      this->checked_snprintf(
+        "topic_name", static_cast<char *>(msg.pubsub_target.topic_name), TOPIC_NAME_BUFFER_SIZE,
+        "%s", topic_name);
+    },
+    msg_);
+  return *this;
+}
```
kobayu858, line 378: `set_message_type()` and `set_service_type()` both guard against being called in the wrong mode (e.g. `set_message_type` fails if called on a Standard builder). However, `set_topic_name()` and `set_pubsub_target_id()` unconditionally write to `pubsub_target`, even when `is_service` is `true`. Should these also validate the mode, or at least check `is_service`?

bdm-k: Both standard mode and performance mode have the `topic_name` and `target_id` fields, so the mode validation is not relevant here. Checking `is_service` is indeed reasonable, but the standard I chose here is:

> It's the caller's responsibility to invoke a correct set of setters to build a valid message.

As we do not want to manage whether a complete set of setters is called, I omitted the `is_service` check as well based on the same reasoning.

src/agnocastlib/src/bridge/agnocast_bridge_utils.cpp
bdm-k: fixed in [152bf40](https://github.com/autowarefoundation/agnocast/pull/1341/commits/152bf40b3306d29b8afc9d2da3e46619f3696995)

src/agnocastlib/src/bridge/agnocast_bridge_utils.cpp
bdm-k: fixed in [152bf40](https://github.com/autowarefoundation/agnocast/pull/1341/commits/152bf40b3306d29b8afc9d2da3e46619f3696995)

kobayu858 (approved):

PR #1343, opened 2026-05-19
fix(agnocastlib): pass `use_clock_thread` to Node constructor

Koichi98:
## Description
Propagate the `use_clock_thread` node option to `NodeTimeSource`.

## How was this PR tested?

- [x] Autoware
- [ ] sample application

**Important notes:**

### DIFF
--- a/src/agnocastlib/src/node/agnocast_node.cpp
+++ b/src/agnocastlib/src/node/agnocast_node.cpp
@@ -28,7 +28,7 @@ Node::Node(
     node_base_, options.parameter_overrides(), local_args_.get())),
   node_clock_(std::make_shared<node_interfaces::NodeClock>(RCL_ROS_TIME)),
   node_time_source_(std::make_shared<node_interfaces::NodeTimeSource>(
-    node_clock_, this, rclcpp::ClockQoS(), options.use_clock_thread())),
+    node_clock_, this, options.clock_qos(), options.use_clock_thread())),
   node_logging_(std::make_shared<node_interfaces::NodeLogging>(logger_))
 {
   if (options.start_parameter_services()) {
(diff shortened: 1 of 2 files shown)


src/agnocastlib/src/node/agnocast_node.cpp
```diff
@@ -27,7 +27,8 @@ Node::Node(
   node_parameters_(std::make_shared<node_interfaces::NodeParameters>(
     node_base_, options.parameter_overrides(), local_args_.get())),
   node_clock_(std::make_shared<node_interfaces::NodeClock>(RCL_ROS_TIME)),
-  node_time_source_(std::make_shared<node_interfaces::NodeTimeSource>(node_clock_, this)),
+  node_time_source_(std::make_shared<node_interfaces::NodeTimeSource>(
+    node_clock_, this, rclcpp::ClockQoS(), options.use_clock_thread())),
```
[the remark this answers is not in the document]
Koichi98: Added in https://github.com/autowarefoundation/agnocast/pull/1343/commits/83b95d90e7960f56dcf4c36567fd74a839efa0a4.

atsushi421: `rclcpp::ClockQoS()` is hardcoded here, which silently ignores `options.clock_qos()`. This is the same class of bug this PR is fixing for `use_clock_thread` — upstream `rclcpp::Node` forwards `options.clock_qos()` into the TimeSource, but agnocast does not. Suggest passing `options.clock_qos()` as the third argument so QoS overrides on the `/clock` topic are honored.

Koichi98: True. Thank you. Fixed in https://github.com/autowarefoundation/agnocast/pull/1343/commits/fd073bd4b681b1a059be03bfe8e14a17f6ef3793.

atsushi421 (approved):

sykwer (approved):

PR #1346, opened 2026-05-19
fix(AgnocastOnlyExecutor): add `cancel_requested_` to separate the state

Koichi98:
## Description
Add a `cancel_requested_` flag to `AgnocastOnlyExecutor` so cancellation is not lost when it races with `spin()`:

  - `cancel()` sets `cancel_requested_ = true` (in addition to `spinning_`). It is sticky — never cleared — so the executor is one-shot: a cancelled instance must not be re-spun.
  - Every `spin()` checks it, so a cancel is honored regardless of whether it arrived before or after `spin()` started:
    - single-threaded: added to the `while` loop condition
    - multi-threaded: early-return before spawning workers; added to the worker loop condition
    - callback-isolated: early-return before setup; added to the spawn guard and the monitoring loop condition

With this, `cancel()` is order-independent and `join()` always returns.

A cleaner design would replace the `spinning_` / `cancel_requested_` flag pair with a single atomic state enum (`Idle` / `Spinning` / `Cancelled`), which would also allow re-spinning the same instance. Since a release is imminent, this PR keeps the change minimal and stays with the flag-based fix; the one-shot limitation and the enum-based alternative are recorded as a TODO comment in the header.

### Problem occurs before this fix
  `AgnocastOnlyExecutor` used a single `spinning_` atomic to carry two opposite intents: `spin()` sets it `true` to start, `cancel()` sets it `false` to stop.

  `spin()` begins with `spinning_.exchange(true)`, which unconditionally overwrites whatever `cancel()` wrote, and `cancel()` leaves no other record that a stop was requested. So when `cancel()` runs *before* `spin()` starts, the cancellation is silently lost and `spin()` enters its loop and never exits.

This is hit in practice by short-lived nodes (e.g. the throwaway publisher node in `create_agnocast_client_publisher()`): the clock executor thread is spawned but, before it is scheduled into `spin()`, the node is already being torn down and calls `cancel()` + `join()`. The spawned thread then enters `spin()`, ignores the lost cancel, and loops forever — `join()` blocks permanently and the process hangs.

Spawning a thread and immediately cancelling it is a legitimate sequence; a cancellation API must tolerate `cancel()` arriving before `spin()`.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

**Important notes:**

### DIFF
--- a/src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
+++ b/src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
@@ -80,7 +80,9 @@ public:
   AGNOCAST_PUBLIC
   virtual void spin() = 0;
 
-  /// Request the executor to stop spinning. Causes the current or next spin() call to return.
+  /// Request the executor to stop spinning. Causes the current spin() call to return.
+  /// One-shot: once called, the executor is permanently stopped -- every subsequent spin()
+  /// returns immediately. Create a new instance to spin again.
   AGNOCAST_PUBLIC
   virtual void cancel();
 
--- a/src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
+++ b/src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
@@ -24,13 +24,23 @@ using WeakCallbackGroupsToNodesMap = std::map<
 struct AgnocastExecutable;
 class Node;
 
-/** @brief Base class for Stage 2 executors that handle only Agnocast callbacks (no RMW). Used with
- * agnocast::Node. */
+/**
+ * @brief Base class for Stage 2 executors that handle only Agnocast callbacks (no RMW). Used with
+ * agnocast::Node.
+ *
+ * One-shot: once cancel() is called, spin() will not run again on the same instance -- create a
+ * new executor instead. All current uses (clock executor, CIE child executors) are recreated.
+ * TODO: to support re-spin, replace the spinning_ / cancel_requested_ flags with one atomic state
+ * enum (Idle / Spinning / Cancelled) so spin() can re-arm to Idle on exit.
+ */
 AGNOCAST_PUBLIC
 class AgnocastOnlyExecutor
 {
 protected:
   std::atomic_bool spinning_;
+  // Sticky cancel flag: set by cancel(), never cleared, so a cancel() before spin() is not lost
+  // when spin() does spinning_.exchange(true). Never cleared -> the executor is one-shot.
+  std::atomic_bool cancel_requested_{false};
   std::unique_ptr<EpollManager> epoll_manager_;
   int shutdown_event_fd_;
   pid_t my_pid_;
@@ -66,6 +76,7 @@ public:
   virtual ~AgnocastOnlyExecutor();
 
   /// Block the calling thread and process Agnocast callbacks in a loop until cancel() is called.
+  /// One-shot: if cancel() was already called, spin() returns at once (see class comment).
   AGNOCAST_PUBLIC
   virtual void spin() = 0;
 
--- a/src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
+++ b/src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
@@ -37,7 +37,7 @@ AGNOCAST_PUBLIC
 class AgnocastOnlyExecutor
 {
 protected:
-  std::atomic_bool spinning_;
+  std::atomic_bool spinning_{false};
   // Sticky cancel flag: set by cancel(), never cleared, so a cancel() before spin() is not lost
   // when spin() does spinning_.exchange(true). Never cleared -> the executor is one-shot.
   std::atomic_bool cancel_requested_{false};
--- a/src/agnocastlib/src/node/agnocast_only_callback_isolated_executor.cpp
+++ b/src/agnocastlib/src/node/agnocast_only_callback_isolated_executor.cpp
@@ -127,12 +127,13 @@ void AgnocastOnlyCallbackIsolatedExecutor::spin()
     {
       std::unique_lock<std::mutex> lock(callback_group_created_cv_mutex_);
       callback_group_created_cv_.wait_for(lock, std::chrono::milliseconds(CV_TIMEOUT_MS), [this] {
-        return callback_group_created_ || !spinning_.load() || !agnocast::ok();
+        return callback_group_created_ || !spinning_.load() || cancel_requested_.load() ||
+               !agnocast::ok();
       });
       callback_group_created_ = false;
     }
 
-    if (!spinning_.load() || !agnocast::ok()) {
+    if (!spinning_.load() || cancel_requested_.load() || !agnocast::ok()) {
       break;
     }
 
@@ -164,7 +165,7 @@ void AgnocastOnlyCallbackIsolatedExecutor::spin()
     }
 
     std::lock_guard<std::mutex> guard{child_resources_mutex_};
-    if (!spinning_.load() || !agnocast::ok()) {
+    if (!spinning_.load() || cancel_requested_.load() || !agnocast::ok()) {
       break;
     }
     for (auto & [group, node] : new_groups) {
--- a/src/agnocastlib/src/node/agnocast_only_callback_isolated_executor.cpp
+++ b/src/agnocastlib/src/node/agnocast_only_callback_isolated_executor.cpp
@@ -39,6 +39,10 @@ void AgnocastOnlyCallbackIsolatedExecutor::spin()
 
   RCPPUTILS_SCOPE_EXIT(this->spinning_.store(false););
 
+  if (cancel_requested_.load()) {
+    return;
+  }
+
   std::vector<std::pair<
     rclcpp::CallbackGroup::SharedPtr, rclcpp::node_interfaces::NodeBaseInterface::SharedPtr>>
     groups_and_nodes;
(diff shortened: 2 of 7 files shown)


src/agnocastlib/test/integration/test_agnocast_only_callback_isolated_executor.cpp
```diff
@@ -160,3 +162,41 @@ TEST_F(AgnocastOnlyCallbackIsolatedExecutorTest, cancel_via_base_class_pointer_u
...
+    spin_returned = true;
+  });
+
+  const auto deadline = std::chrono::steady_clock::now() + std::chrono::seconds(10);
+  while (!spin_returned) {
+    ASSERT_LT(std::chrono::steady_clock::now(), deadline)
+      << "spin() did not return after a cancel() that preceded it (cancel-before-spin race)";
+    std::this_thread::sleep_for(std::chrono::milliseconds(50));
+  }
+  spin_thread.join();
```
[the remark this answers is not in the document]
Koichi98: The given solutions don't look like it works, and the current test (=`cancel_via_base_class_pointer_unblocks_spin`) also has the same pattern. I'll leave it for now.

src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
```diff
@@ -66,6 +76,7 @@ class AgnocastOnlyExecutor
   virtual ~AgnocastOnlyExecutor();
 
   /// Block the calling thread and process Agnocast callbacks in a loop until cancel() is called.
+  /// One-shot: if cancel() was already called, spin() returns at once (see class comment).
```
atsushi421, line 79: The class comment now documents one-shot semantics, but the Doxygen on `cancel()` just below (out of this hunk) still reads only "Causes the current or next spin() call to return" — which is the most prominent place a caller looking at the API will land. Consider extending the `cancel()` comment too, e.g. "Once called, the executor is permanently stopped: subsequent spin() calls return at once. Create a new instance to spin again."

src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
```diff
@@ -24,13 +24,23 @@ using WeakCallbackGroupsToNodesMap = std::map<
...
+ * enum (Idle / Spinning / Cancelled) so spin() can re-arm to Idle on exit.
+ */
 AGNOCAST_PUBLIC
 class AgnocastOnlyExecutor
 {
 protected:
   std::atomic_bool spinning_;
+  // Sticky cancel flag: set by cancel(), never cleared, so a cancel() before spin() is not lost
+  // when spin() does spinning_.exchange(true). Never cleared -> the executor is one-shot.
+  std::atomic_bool cancel_requested_{false};
```
atsushi421, line 43: `cancel_requested_` uses an in-class initializer (`{false}`), while `spinning_` is left default-initialized here and explicitly set to `false` in the constructor initializer list of `agnocast_only_executor.cpp`. Mixing the two styles on closely related atomics in the same class is a mild footgun for whoever adds the next flag. Suggest `std::atomic_bool spinning_{false};` here and dropping it from the cpp initializer list.

src/agnocastlib/src/node/agnocast_only_callback_isolated_executor.cpp
```diff
@@ -119,7 +123,7 @@ void AgnocastOnlyCallbackIsolatedExecutor::spin()
   }  // guard child_resources_mutex_
 
   // Monitoring loop: wait for notification when new callback groups are created
-  while (agnocast::ok() && spinning_.load()) {
+  while (agnocast::ok() && spinning_.load() && !cancel_requested_.load()) {
```
atsushi421, line 126: The `wait_for` predicate just below (around line 130, outside this hunk) is the one place in this file that was not updated to observe `cancel_requested_`. It still works today because `cancel()` also clears `spinning_` under the cv mutex, but it is inconsistent with the rest of the loop and with the PR's stated invariant that every spin checks the flag. If the future state-enum refactor (per the TODO in the header) ever updates `cancel_requested_` without clearing `spinning_`, this waiter silently misses the wake. Suggest adding `|| cancel_requested_.load()` to the predicate.

src/agnocastlib/test/integration/test_agnocast_only_callback_isolated_executor.cpp
```diff
@@ -160,3 +162,41 @@ TEST_F(AgnocastOnlyCallbackIsolatedExecutorTest, cancel_via_base_class_pointer_u
...
+  ExecutorT executor;
+  executor.cancel();  // cancel() happens before spin() ever starts
+
+  std::atomic_bool spin_returned{false};
+  std::thread spin_thread([&executor, &spin_returned]() {
+    executor.spin();
+    spin_returned = true;
+  });
+
+  const auto deadline = std::chrono::steady_clock::now() + std::chrono::seconds(10);
```
atsushi421: Two issues with this polling loop. First, if `ASSERT_LT` fires (i.e. on a real regression — exactly what this test exists to catch), the function returns with `spin_thread` neither joined nor detached, so `~thread` calls `std::terminate` and the gtest output is lost. Second, the 50 ms sleep adds up to 50 ms latency on every success run and there is a tiny window between `spin_returned = true` and the thread actually exiting. Consider replacing with `auto fut = std::async(std::launch::async, [&]{ executor.spin(); }); ASSERT_EQ(fut.wait_for(std::chrono::seconds(10)), std::future_status::ready);` — same timeout semantics, no busy-wait, no terminate.

src/agnocastlib/test/integration/test_agnocast_only_callback_isolated_executor.cpp
```diff
@@ -160,3 +162,41 @@ TEST_F(AgnocastOnlyCallbackIsolatedExecutorTest, cancel_via_base_class_pointer_u
...
+  while (!spin_returned) {
+    ASSERT_LT(std::chrono::steady_clock::now(), deadline)
+      << "spin() did not return after a cancel() that preceded it (cancel-before-spin race)";
+    std::this_thread::sleep_for(std::chrono::milliseconds(50));
+  }
+  spin_thread.join();
+  EXPECT_TRUE(spin_returned);
+}
+
+TEST_F(AgnocastOnlyCallbackIsolatedExecutorTest, single_threaded_cancel_before_spin_returns)
```
atsushi421: `single_threaded_cancel_before_spin_returns` and (just below) `multi_threaded_cancel_before_spin_returns` exercise `AgnocastOnlySingleThreadedExecutor` and `AgnocastOnlyMultiThreadedExecutor`, but they live in the CIE test file under `AgnocastOnlyCallbackIsolatedExecutorTest`. The repo already has dedicated test files for ST and MT, and cross-executor cases are templated in `test_agnocast_init_ok_shutdown.cpp`. `--gtest_filter=AgnocastOnlyCallbackIsolatedExecutorTest*` will now run ST/MT tests too, and discovery is misleading. Suggest moving them to their respective executor test files, or template-ing the helper into the cross-executor file.

src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1346/commits/dd4a7bd1459109fae3b2ea06fdf0903e145a8a7f.

src/agnocastlib/include/agnocast/node/agnocast_only_executor.hpp
Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1346/commits/db6de0b5caca8747e93b54ac0dc77e6e6f4b3ce1.

src/agnocastlib/src/node/agnocast_only_callback_isolated_executor.cpp
Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1346/commits/ed86d6eb615043027a00abd80410bea60a82f861.

src/agnocastlib/test/integration/test_agnocast_only_callback_isolated_executor.cpp
Koichi98: `std::async` fixes the busy-wait but not the `std::terminate`: a future from `std::async(std::launch::async, …)` has a destructor that blocks until the task finishes, so on a real regression the failing assert just turns into an indefinite hang in ~future instead.  
Moved it to `test_agnocast_init_ok_shutdown.cpp` and reused its `wait_until_or_fail` helper. `FAIL()` stays inside the helper, so `join()` still runs — no terminate. Polling drops 50 ms → 2 ms.

Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1346/commits/29910437ceac383ae45e983bda51ea9d1ab8dc46.

src/agnocastlib/test/integration/test_agnocast_only_callback_isolated_executor.cpp
Koichi98: Fixed at the same time with the above comment.
https://github.com/autowarefoundation/agnocast/pull/1346/commits/29910437ceac383ae45e983bda51ea9d1ab8dc46

atsushi421 (approved):
Thanks!

atsushi421 (approved):

sykwer (approved):

PR #1350, opened 2026-05-21
feat: per-IPC-namespace discovery daemon + AgnocastDaemonState msgs (shared base)

k1832:
## Description

Adds the shared building blocks for two upcoming features — cross-IPC-namespace / cross-ECU observability (#1351) and cross-IPC-namespace bridge auto-generation (#1352 + #1353):

- **`ros2agnocast_discovery_msgs`** package — three msgs (`AgnocastEndpoint`, `AgnocastTopic`, `AgnocastDaemonState`) describing the full Agnocast state of one IPC namespace at one timestamp.
- **`ros2agnocast_discovery_agent`** package — Python rclpy node (one process per IPC namespace) that snapshots the local Agnocast state via the existing `libagnocast_ioctl_wrapper.so` and publishes it on `/_agnocast_discovery` every 1 s with QoS `Reliable + TransientLocal + KeepLast(1) + Liveliness Automatic 30 s`.

The agent is intentionally a thin Python node so reviewers can verify it line-by-line without C++ template machinery; the CLI verb extension (#1351) and the bridge-generation feature (#1352 agnocastlib data layer + #1353 discovery agent: connect everything) build on this base.

This PR is **additive only** — it does not touch `agnocastlib`, heaphook, or the kernel module.

## Related links

- Design doc (cross-IPC-NS / cross-ECU observability — #1351 builds on this): https://tier4.atlassian.net/wiki/x/VAGzNgE (TIER IV internal link)
- Design doc (cross-IPC-NS bridge auto-generation — #1352 + #1353 build on this): https://tier4.atlassian.net/wiki/x/2YFtNwE (TIER IV internal link)
- Jira parent epic: T4DEV-52095 (TIER IV internal)
- Merge order: **#1350 (this PR)** → #1351 (CLI gossip subscribe) → #1352 (agnocastlib data layer) → #1353 (discovery agent: connect everything)

## How was this PR tested?

- [ ] Autoware
- [ ] `bash scripts/test/e2e_test_1to1.bash` (required) — N/A: no agnocastlib / kmod / heaphook changes, no data-path code touched
- [ ] `bash scripts/test/e2e_test_2to2.bash` (required) — N/A: same reason
- [ ] kunit tests (required when modifying the kernel module) — N/A: kmod untouched
- [ ] `bash scripts/test/run_requires_kernel_module_tests.bash` (required) — N/A: same reason
- [x] sample application
- [x] Added automated tests (run locally on this branch, all pass):
  - `colcon test --packages-select ros2agnocast_discovery_agent` → 6 pytest cases in `test/test_agent.py` covering ioctl-to-msg conversion, machine-id fallback, and multi-endpoint topic snapshots.
  - `bash scripts/test/e2e_test_discovery_agent.bash` → single-namespace smoke verifying that `/_agnocast_discovery` carries an `AgnocastDaemonState` with the expected fields and QoS profile (RELIABLE / TRANSIENT_LOCAL / Liveliness AUTOMATIC).

### Manual test: cross-IPC-namespace observability

> **Caveat**: This has only been exercised on a single physical machine (two IPC namespaces on the same host via `unshare`). Cross-ECU verification on real hardware is still outstanding.

Single-namespace coverage is already automated above. For real cross-IPC-NS verification (requires `CAP_SYS_ADMIN`, not available in CI), run:

```bash
# Prereqs: agnocast kmod loaded; workspace built; three separate shells.

# --- Shell A: IPC NS "alpha" + a talker. The talker launch defaults
#     `discovery_agent` to `false`; pass `:=true` here so the alpha NS
#     gets its own agent (exactly one agent per IPC NS).
sudo -E unshare --ipc bash -c '
  ros2 launch agnocast_sample_application talker.launch.xml discovery_agent:=true &
  sleep 120; wait
'

# --- Shell B: IPC NS "beta", agent only (no Agnocast traffic in this NS) ---
sudo -E unshare --ipc bash -c '
  ros2 launch ros2agnocast_discovery_agent discovery_agent.launch.xml &
  sleep 120; wait
'

# --- Shell C: host NS, verify both namespaces are visible on the gossip topic ---
# If a stale `ros2 daemon` is running with old PYTHONPATH (started before
# install/setup.bash was sourced), it won't know about the new msg package
# and `ros2 topic echo` will report "invalid message type". Force a fresh
# ros2 daemon under this shell's env.
ros2 daemon stop && sleep 1 && ros2 daemon start && sleep 3
# Expect two AgnocastDaemonState publishers, one per IPC NS.
ros2 topic info /_agnocast_discovery --verbose | grep "Publisher count"
# The gossip publisher uses Reliable + TransientLocal QoS, so `ros2 topic echo`
# needs matching QoS flags or the subscription's defaults (BestEffort + Volatile)
# will silently fail to match on `durability`. Use `--once` to grab the
# late-joiner snapshot via TransientLocal in one shot.
ros2 topic echo --once \
  --qos-reliability reliable \
  --qos-durability transient_local \
  /_agnocast_discovery
```

> **Caveats**:
> - `sudo -E` is required so that `sudo` preserves `RMW_IMPLEMENTATION` (and other ROS env vars) into the unshared bash. Without `-E`, the inner agent falls back to the default RMW (Fast-DDS on most systems) while the host shell uses Cyclone DDS — DDS discovery still succeeds via the standard RTPS UDP path so `topic info` shows the publishers, but data delivery between the mixed RMW implementations is unreliable and `topic echo` silently receives nothing.
> - The `ros2 daemon stop && start` step is needed once per fresh shell, after sourcing `install/setup.bash`, so the ros2 daemon picks up the new `ros2agnocast_discovery_msgs` package in its PYTHONPATH (otherwise `topic echo` rejects the resolved type as "invalid").

Expected:

- Two publishers on `/_agnocast_discovery` (one per IPC NS).
- The alpha snapshot lists `/talker_node` under `topics[]`; the beta snapshot lists nothing under `topics[]` (no Agnocast process in beta).
- Both snapshots share `host_uuid` (same machine) but differ in `ipc_ns_inode`.

### Known design deviation (acceptable; documented here so reviewers don't need to flag it)

- **Publish strategy: 1 Hz timer-only**, not the design's "change-detection immediate publish + 0.1 Hz heartbeat". Worst-case detection latency is therefore ≤ 1 s instead of ~ms; the change-detection optimization is a follow-up.

### Why this is `need-patch-update`

This PR is two **new** ROS 2 packages plus a tiny Python wrapper around an already-shipped `libagnocast_ioctl_wrapper.so` symbol. No changes to `agnocastlib`, heaphook, kmod, MQ payloads, or any existing ABI. See CONTRIBUTING.md §"Versioning rules".

- `need-patch-update`

### DIFF
--- a/scripts/test/e2e_test_discovery_agent.bash
+++ b/scripts/test/e2e_test_discovery_agent.bash
@@ -1,0 +1,99 @@
+#!/bin/bash
+#
+# E2E regression smoke test for ros2agnocast_discovery_agent.
+#
+# Starts one discovery agent in the current IPC namespace, lets it run for a
+# couple of polling cycles, and verifies that an AgnocastDaemonState message
+# appears on /_agnocast_discovery with the expected fields populated. This
+# does not require any Agnocast pub / sub to be running -- the message is
+# expected to carry an empty `topics` list in that case.
+#
+# Requires:
+#   * agnocast kernel module loaded (sudo insmod agnocast_kmod/agnocast.ko)
+#   * workspace built (bash scripts/dev/build_all.bash)
+#
+# Usage:
+#   bash scripts/test/e2e_test_discovery_agent.bash
+#
+# Exit codes:
+#   0 - all checks passed
+#   1 - prerequisite failure (kmod / workspace)
+#   2 - assertion failure (msg fields didn't match expectation)
+
+ROOT_DIR=$(cd "$(dirname "$0")/../.." && pwd)
+ECHO_TIMEOUT_SEC=${ECHO_TIMEOUT_SEC:-5}
+DAEMON_WARMUP_SEC=${DAEMON_WARMUP_SEC:-2}
+
+red()    { printf '\033[31m%s\033[0m\n' "$*"; }
+green()  { printf '\033[32m%s\033[0m\n' "$*"; }
+yellow() { printf '\033[33m%s\033[0m\n' "$*"; }
+
+# ----- prerequisites -----
+if ! grep -q "^agnocast " /proc/modules; then
+    red "ERROR: agnocast kmod not loaded."
+    echo "  -> sudo insmod $ROOT_DIR/agnocast_kmod/agnocast.ko" >&2
+    exit 1
+fi
+
+if [ ! -f "$ROOT_DIR/install/setup.bash" ]; then
+    red "ERROR: workspace not built ($ROOT_DIR/install/setup.bash missing)."
+    echo "  -> bash $ROOT_DIR/scripts/dev/build_all.bash" >&2
+    exit 1
+fi
+
+# shellcheck disable=SC1091
+source "$ROOT_DIR/install/setup.bash"
+set -u
+green "✓ kmod loaded and workspace sourced"
+
+# ----- start daemon -----
+LOG_DIR=$(mktemp -d)
+cleanup() {
+    pkill -P $$ 2>/dev/null || true
+    sleep 1
+    rm -rf "$LOG_DIR"
+}
+trap cleanup EXIT
+
+yellow "Starting agnocast_discovery_agent..."
+ros2 run ros2agnocast_discovery_agent discovery_agent > "$LOG_DIR/agent.log" 2>&1 &
+sleep "$DAEMON_WARMUP_SEC"
+
+if ! grep -q "discovery_agent up" "$LOG_DIR/agent.log"; then
+    red "ERROR: daemon did not log startup within ${DAEMON_WARMUP_SEC}s."
+    echo "----- agent log -----" >&2
+    cat "$LOG_DIR/agent.log" >&2
+    exit 2
+fi
+green "✓ daemon started"
+
+# ----- assert msg shape -----
+fail() {
+    red "ERROR: $1"
+    [ -n "${2:-}" ] && { echo "----- output -----" >&2; printf '%s\n' "$2" >&2; }
+    echo "----- agent log -----" >&2
+    cat "$LOG_DIR/agent.log" >&2
+    exit 2
+}
+
+msg=$(timeout "$ECHO_TIMEOUT_SEC" ros2 topic echo --once /_agnocast_discovery 2>&1) \
+    || fail "ros2 topic echo on /_agnocast_discovery timed out or failed" "$msg"
+
+grep -q '^schema_version: 1$'                  <<<"$msg" || fail "schema_version != 1" "$msg"
+grep -q '^host_uuid: [0-9a-f-]\+$'             <<<"$msg" || fail "host_uuid missing or malformed" "$msg"
+grep -q '^host_hostname: '                     <<<"$msg" || fail "host_hostname missing" "$msg"
+grep -q '^ipc_ns_inode: [0-9]\+$'              <<<"$msg" || fail "ipc_ns_inode missing or non-numeric" "$msg"
+grep -q '^timestamp:'                          <<<"$msg" || fail "timestamp missing" "$msg"
+grep -q '^topics:'                             <<<"$msg" || fail "topics field missing" "$msg"
+green "✓ AgnocastDaemonState shape OK (schema_version, host_uuid, host_hostname, ipc_ns_inode, timestamp, topics)"
+
+# ----- QoS sanity (Reliable + TransientLocal + Liveliness Automatic) -----
+qos=$(timeout "$ECHO_TIMEOUT_SEC" ros2 topic info /_agnocast_discovery --verbose 2>&1) \
+    || fail "ros2 topic info on /_agnocast_discovery failed" "$qos"
+grep -q "Reliability: RELIABLE"          <<<"$qos" || fail "QoS reliability != RELIABLE" "$qos"
+grep -q "Durability: TRANSIENT_LOCAL"    <<<"$qos" || fail "QoS durability != TRANSIENT_LOCAL" "$qos"
+grep -q "Liveliness: AUTOMATIC"          <<<"$qos" || fail "QoS liveliness != AUTOMATIC" "$qos"
+green "✓ QoS profile: RELIABLE + TRANSIENT_LOCAL + Liveliness AUTOMATIC"
+
+green ""
+green "===== ALL CHECKS PASSED ====="
--- a/scripts/test/e2e_test_discovery_agent.bash
+++ b/scripts/test/e2e_test_discovery_agent.bash
@@ -79,13 +79,16 @@ fail() {
 msg=$(timeout "$ECHO_TIMEOUT_SEC" ros2 topic echo --once /_agnocast_discovery 2>&1) \
     || fail "ros2 topic echo on /_agnocast_discovery timed out or failed" "$msg"
 
-grep -q '^schema_version: 1$'                  <<<"$msg" || fail "schema_version != 1" "$msg"
-grep -q '^host_uuid: [0-9a-f-]\+$'             <<<"$msg" || fail "host_uuid missing or malformed" "$msg"
-grep -q '^host_hostname: '                     <<<"$msg" || fail "host_hostname missing" "$msg"
-grep -q '^ipc_ns_inode: [0-9]\+$'              <<<"$msg" || fail "ipc_ns_inode missing or non-numeric" "$msg"
-grep -q '^timestamp:'                          <<<"$msg" || fail "timestamp missing" "$msg"
-grep -q '^topics:'                             <<<"$msg" || fail "topics field missing" "$msg"
-green "✓ AgnocastDaemonState shape OK (schema_version, host_uuid, host_hostname, ipc_ns_inode, timestamp, topics)"
+grep -q '^schema_version: 1$'                       <<<"$msg" || fail "schema_version != 1" "$msg"
+grep -q '^agnocast_version: \(.*\)$'                <<<"$msg" || fail "agnocast_version missing" "$msg"
+# Assert agnocast_version is non-empty so a misconfigured importlib.metadata
+# lookup is caught (would silently fall back to '' otherwise).
+grep -q '^agnocast_version: ..*$'                   <<<"$msg" || fail "agnocast_version is empty" "$msg"
+grep -q '^host_uuid: [0-9a-f-]\+$'                  <<<"$msg" || fail "host_uuid missing or malformed" "$msg"
+grep -q '^host_hostname: '                          <<<"$msg" || fail "host_hostname missing" "$msg"
+grep -q '^ipc_ns_inode: [0-9]\+$'                   <<<"$msg" || fail "ipc_ns_inode missing or non-numeric" "$msg"
+grep -q '^topics:'                                  <<<"$msg" || fail "topics field missing" "$msg"
+green "✓ AgnocastDaemonState shape OK (schema_version, agnocast_version, host_uuid, host_hostname, ipc_ns_inode, topics)"
 
 # ----- QoS sanity (Reliable + TransientLocal + Liveliness Automatic) -----
 qos=$(timeout "$ECHO_TIMEOUT_SEC" ros2 topic info /_agnocast_discovery --verbose 2>&1) \
@@ -94,6 +97,20 @@ grep -q "Reliability: RELIABLE"          <<<"$qos" || fail "QoS reliability != R
 grep -q "Durability: TRANSIENT_LOCAL"    <<<"$qos" || fail "QoS durability != TRANSIENT_LOCAL" "$qos"
 grep -q "Liveliness: AUTOMATIC"          <<<"$qos" || fail "QoS liveliness != AUTOMATIC" "$qos"
 green "✓ QoS profile: RELIABLE + TRANSIENT_LOCAL + Liveliness AUTOMATIC"
+
+# ----- pop a talker in the same IPC NS and assert its endpoint shows up -----
+yellow "Starting agnocast_sample_application talker (discovery_agent=false to avoid a 2nd agent)..."
+ros2 launch agnocast_sample_application talker.launch.xml discovery_agent:=false \
+    > "$LOG_DIR/talker.log" 2>&1 &
+# Give the talker time to call its Publisher<T> ctor and the daemon time to
+# observe it on its next 1 Hz tick.
+sleep 3
+msg_with_topic=$(timeout "$ECHO_TIMEOUT_SEC" ros2 topic echo --once /_agnocast_discovery 2>&1) \
+    || fail "ros2 topic echo (with talker running) timed out" "$msg_with_topic"
+
+grep -q 'topic_name: /my_topic'      <<<"$msg_with_topic" || fail "talker's /my_topic not seen in snapshot" "$msg_with_topic"
+grep -q 'node_name: /talker_node'    <<<"$msg_with_topic" || fail "talker_node not seen in snapshot" "$msg_with_topic"
+green "✓ talker's /my_topic + /talker_node appear in the snapshot"
 
 green ""
 green "===== ALL CHECKS PASSED ====="
--- a/src/agnocast_sample_application/launch/talker.launch.xml
+++ b/src/agnocast_sample_application/launch/talker.launch.xml
@@ -1,4 +1,14 @@
 <launch>
+    <!--
+      `discovery_agent` arg controls whether the per-IPC-namespace Agnocast
+      discovery agent is spawned alongside the talker. Exactly one daemon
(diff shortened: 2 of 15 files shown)


src/ros2agnocast_discovery_agent/setup.py
```diff
@@ -0,0 +1,29 @@
...
+    maintainer_email='keita.morisaki@tier4.jp',
+    description='Per-IPC-namespace daemon publishing Agnocast state on '
+                '/_agnocast_discovery for cross-NS/ECU observability and bridge generation.',
+    license='Apache License 2.0',
+    tests_require=['pytest'],
+    entry_points={
+        'console_scripts': [
+            'discovery_agent = ros2agnocast_discovery_agent.agent:main',
+        ],
+    },
```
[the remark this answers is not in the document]
k1832: Verified empirically: dropping the manual `lib/<pkg>` install makes the `console_scripts` wrapper land in `bin/` (where `ros2 run` does not look), and `ros2 run ros2agnocast_discovery_agent discovery_agent` then fails with "No executable found". The two install paths don't actually collide (`bin/` vs `lib/<pkg>/`), and the manual install is the standard ament_python idiom for exposing console scripts to `ros2 run`. Keeping both.

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/agent.py
```diff
@@ -0,0 +1,275 @@
...
+
+class DiscoveryAgent(Node):
+    """rclpy Node that publishes the local Agnocast state every PUBLISH_INTERVAL_SEC.
+
+    Also subscribes to its own gossip topic so the bridge decider can observe
+    remote namespace state; the callback records the latest snapshot keyed by
+    ``(host_uuid, ipc_ns_inode)``.
+    """
+
+    def __init__(self, ioctl_lib=None):
```
Koichi98: Do we need this `ioctl_lib`? Is this for the future mock tests?

src/ros2agnocast_discovery_agent/package.xml
```diff
@@ -0,0 +1,28 @@
...
+  <name>ros2agnocast_discovery_agent</name>
+  <version>2.3.4</version>
+  <description>
+    Per-IPC-namespace daemon that reads the local Agnocast state via the
+    existing NS-scoped ioctl and publishes it on /_agnocast_discovery so
+    other namespaces and ECUs running ros2agnocast tooling can build a
+    cross-NS / cross-ECU view and request bridges on demand.
+  </description>
+
+  <maintainer email="keita.morisaki@tier4.jp">Keita Morisaki</maintainer>
```
Koichi98, line 13: Could you also add me, @sykwer, and @ruth561? Same for some other files.

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/agent.py
```diff
@@ -0,0 +1,275 @@
...
+        msg.host_hostname = self._host_hostname
+        now = self.get_clock().now().to_msg()
+        msg.timestamp = Time(sec=now.sec, nanosec=now.nanosec)
+        msg.ipc_ns_inode = self._ipc_ns_inode
+        msg.topics = read_local_topics(self._lib)
+        return msg
+
+    def _on_remote_state(self, msg: AgnocastDaemonState) -> None:
+        # Skip our own messages.
+        if msg.host_uuid == self._host_uuid and msg.ipc_ns_inode == self._ipc_ns_inode:
```
Koichi98, line 271: We have `ignore_local_publications` option for rclcpp, but seems like there's none for rclpy...?

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/agent.py
```diff
@@ -0,0 +1,275 @@
...
+    AgnocastEndpoint,
+    AgnocastTopic,
+)
+
+
+GOSSIP_TOPIC = '/_agnocast_discovery'
+SCHEMA_VERSION = 1
+PUBLISH_INTERVAL_SEC = 1.0
+LIVELINESS_LEASE_SEC = 30.0
+MACHINE_ID_PATH = '/etc/machine-id'
```
Koichi98: nit: `/etc/machine-id` comes from systemd, but Agnocast does not depend on systemd. We do not need an ID that lasts across reboots here — we just need all daemons on one host to get the same ID. `/proc/sys/kernel/random/boot_id` does that, comes from the kernel, and works in containers.

Koichi98: Also, if the image build writes a value into `/etc/machine-id`, it gets baked into the image and all ECUs flashed with it might share one machine-id — so `host_uuid` can collide. This is avoidable, but safer not to rely on `/etc/machine-id`.

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/agent.py
```diff
@@ -0,0 +1,275 @@
...
+        mirror it here so the in-memory cache agrees.
+
+        ``now_sec`` is exposed for tests; production code lets the method read
+        ``self.get_clock()`` directly.
+        """
+        if now_sec is None:
+            now_sec = self.get_clock().now().nanoseconds / 1e9
+        stale_keys = [
+            key for key, msg in self._remote_states.items()
+            if now_sec - (msg.timestamp.sec + msg.timestamp.nanosec / 1e9)
```
Koichi98: Using the remote ECU timestamp? This might be a bit fragile since the time isn't exactly the same between ECUs.

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/agent.py
```diff
@@ -0,0 +1,275 @@
...
+    ``(host_uuid, ipc_ns_inode)``.
+    """
+
+    def __init__(self, ioctl_lib=None):
+        super().__init__('agnocast_discovery_agent')
+        self._lib = ioctl_lib if ioctl_lib is not None else _load_ioctl_wrapper()
+        self._host_uuid = _read_machine_id()
+        self._host_hostname = socket.gethostname()
+        self._ipc_ns_inode = _read_ipc_ns_inode()
+        self._agnocast_version = os.environ.get('AGNOCAST_VERSION', '')
```
Koichi98: We don't have `AGNOCAST_VERSION`. It might be a good way to get the agent's own package version — populating it from importlib.metadata.
https://github.com/autowarefoundation/agnocast/blob/1b308a4f27278b573a4a9c97cd77b5f1389fe741/src/ros2agnocast/ros2agnocast/verb/version.py#L193

src/agnocast_sample_application/launch/talker.launch.xml
```diff
@@ -1,4 +1,14 @@
 <launch>
+    <!--
+      `discovery_agent` arg controls whether the per-IPC-namespace Agnocast
+      discovery agent is spawned alongside the talker. Exactly one daemon
+      should run per IPC namespace; set this to false when including this
+      launch from another launch that already spawns the daemon.
+    -->
+    <arg name="discovery_agent" default="true"/>
```
Koichi98: The agent startup approach in Autoware is converging, but the general recommendation for starting the daemon is still open, so I'd like to at least default `discovery_agent` to `false` here — that keeps the talker/listener launches from spawning two daemons in one IPC namespace. 
The `Manual Test` with the sample application was really helpful, thanks!

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/agent.py
Koichi98: Might be better to add assert for the version field in `scripts/test/e2e_test_discovery_agent.bash`.

scripts/test/e2e_test_discovery_agent.bash
```diff
@@ -0,0 +1,99 @@
+#!/bin/bash
+#
+# E2E regression smoke test for ros2agnocast_discovery_agent.
+#
+# Starts one discovery agent in the current IPC namespace, lets it run for a
+# couple of polling cycles, and verifies that an AgnocastDaemonState message
+# appears on /_agnocast_discovery with the expected fields populated. This
+# does not require any Agnocast pub / sub to be running -- the message is
+# expected to carry an empty `topics` list in that case.
```
Koichi98, line 9: Can't we just add one or two pub/sub and verify the `topic` list too? Or maybe you intentionally avoid doing this for stability?

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/agent.py
```diff
@@ -0,0 +1,275 @@
...
+def _read_machine_id() -> str:
+    """Return /etc/machine-id formatted as a UUID, fallback to a random UUID."""
+    try:
+        with open(MACHINE_ID_PATH) as fp:
+            raw = fp.read().strip()
+        if len(raw) == 32 and all(c in '0123456789abcdef' for c in raw):
+            return str(uuid.UUID(raw))
+    except OSError:
+        pass
+    return str(uuid.uuid4())
```
ruth561: How about logging a warning like "Failed to read /etc/machine-id, using a random value"?
This would help us identify the root cause more easily if it unexpectedly falls back to a random value and causes issues.

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/agent.py
ruth561: I agree. How about using the local timestamp recorded when the previous `AgnocastDaemonState` message arrived?
If we use the local timestamp, we won't need the `timestamp` field in `AgnocastDaemonState` anymore, right?

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/agent.py
```diff
@@ -0,0 +1,275 @@
...
+        would leave its last snapshot in ``_remote_states`` forever, slowly
+        growing memory in long-running deployments. The DDS Liveliness lease
+        already forces the publisher off the topic at the same timescale; we
+        mirror it here so the in-memory cache agrees.
+
+        ``now_sec`` is exposed for tests; production code lets the method read
+        ``self.get_clock()`` directly.
+        """
+        if now_sec is None:
+            now_sec = self.get_clock().now().nanoseconds / 1e9
```
ruth561: Since the per-IPC daemon is for sharing the latest node and topic states for each IPC namespace and ECU, can we prevent the use of simulation time by using system_clock explicitly?
Or, do we need to consider cases where the IPC daemon process uses simulation time?

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/agent.py
```diff
@@ -0,0 +1,275 @@
...
+class DiscoveryAgent(Node):
+    """rclpy Node that publishes the local Agnocast state every PUBLISH_INTERVAL_SEC.
+
+    Also subscribes to its own gossip topic so the bridge decider can observe
+    remote namespace state; the callback records the latest snapshot keyed by
+    ``(host_uuid, ipc_ns_inode)``.
+    """
+
+    def __init__(self, ioctl_lib=None):
+        super().__init__('agnocast_discovery_agent')
```
Koichi98: Might be better to add (hostname or host_uuid) + ipc_ns_inode, since to make the node name unique. There will be several `agnocast_discovery_agent` nodes in the same system.

src/ros2agnocast_discovery_agent/package.xml
k1832: Done.

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/agent.py
k1832: Dropped — `ioctl_lib` had no callers and no mock test wired to it.

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/agent.py
k1832: rclpy doesn't expose it. The callback already skips `(self_host_uuid, self_ipc_ns_inode)`, which is the equivalent filter.

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/agent.py
k1832: Done — switched to `/proc/sys/kernel/random/boot_id`.

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/agent.py
k1832: Done — dropped `msg.timestamp`. The agent stores `(msg, received_at)` and prunes against the local arrival time.

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/agent.py
k1832: Done — `importlib.metadata.version()` with a `PackageNotFoundError` fallback. Added a non-empty assertion in `e2e_test_discovery_agent.bash`.

src/agnocast_sample_application/launch/talker.launch.xml
k1832: Done — both talker and listener default to `false`.

scripts/test/e2e_test_discovery_agent.bash
k1832: Done — the e2e now launches a talker and asserts `/my_topic` + `/talker_node` appear in a fresh snapshot.

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/agent.py
k1832: Done.

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/agent.py
k1832: Done — `parameter_overrides=[Parameter('use_sim_time', False)]` on the node, plus a dedicated `Clock(ClockType.SYSTEM_TIME)` for prune / receive timestamps.

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/agent.py
k1832: Done — node name is `agnocast_discovery_agent_{host8hex}_{ipc_ns_inode}`.

src/ros2agnocast_discovery_agent/package.xml
```diff
@@ -0,0 +1,31 @@
...
+  <maintainer email="sykwer@gmail.com">Takahiro Ishikawa-Aso</maintainer>
+  <maintainer email="koichi.imai.2@tier4.jp">Koichi Imai</maintainer>
+  <maintainer email="ryuta.kambe@tier4.jp">Ryuta Kambe</maintainer>
```
Koichi98:
```suggestion
  <maintainer email="primenumber_2_3_5@yahoo.co.jp">Takumi Jin</maintainer>
```

src/ros2agnocast_discovery_agent/setup.py
```diff
@@ -0,0 +1,28 @@
...
+        ('share/ament_index/resource_index/packages',
+            ['resource/' + package_name]),
+        ('share/' + package_name, ['package.xml']),
+        ('share/' + package_name + '/launch', ['launch/discovery_agent.launch.xml']),
+        ('lib/' + package_name, ['scripts/discovery_agent']),
+    ],
+    install_requires=['setuptools'],
+    zip_safe=True,
+    maintainer='Keita Morisaki',
+    maintainer_email='keita.morisaki@tier4.jp',
```
Koichi98: Could you also add me and the others too?

k1832: Done — added Takahiro Ishikawa-Aso, Koichi Imai, and Takumi Jin to `setup.py` so it matches `package.xml`. Also swapped Ryuta Kambe for Takumi Jin in `ros2agnocast_discovery_msgs/package.xml` to keep both new packages on the same set.

Koichi98 (approved):

ruth561 (approved):

PR #1351, opened 2026-05-21
feat: ros2agnocast CLI verbs see Agnocast endpoints across IPC namespaces and ECUs

k1832:
## Description

**`ros2agnocast` CLI verbs now show Agnocast endpoints from every IPC namespace and ECU on the `ROS_DOMAIN_ID`, not just the caller's own NS.**

Affected verbs: `ros2 topic list_agnocast` / `info_agnocast`, `ros2 node list_agnocast` / `info_agnocast`.

### How
- Verbs read endpoint state from the `/_agnocast_discovery` gossip published per-NS by the discovery agent (#1350), instead of the NS-scoped ioctl. Every agent on the domain (including the caller's own NS) gossips its state.
- **ioctl fallback**: when no agent is gossiping (agent down, RMW mismatch), the CLI synthesises the caller's-NS view from `libagnocast_ioctl_wrapper`, so local output keeps working.
- **Per-NS bridge label**: a topic that crosses a boundary (endpoints in 2+ NSes, or a NS plus a ROS 2 endpoint) but lacks a bridge its real endpoints need shows `(WARN: one or more necessary bridges are not running)` instead of silently `(Agnocast enabled)` — so a talker and listener in different NSes that aren't actually connected are made explicit.

### Notes
- New hidden `--gossip-timeout` flag (default `2.0` s); to be replaced by a ros2-daemon-driven stop condition (TODO in `discovery.py`).
- Staleness is bounded by the publisher's DDS Liveliness lease (`AUTOMATIC`, 30 s); no separate freshness filter.
- New modules: `ros2agnocast/discovery.py` (gossip + projection + bridge label) and `ros2agnocast/_ioctl.py` (fallback).
- **Additive only** — no agnocastlib / kmod / heaphook changes.

## Related links

- Design doc: https://tier4.atlassian.net/wiki/x/VAGzNgE (TIER IV internal link)
- Jira parent epic: T4DEV-52095 (TIER IV internal)
- Merge order: #1350 (discovery msgs + agent) → **#1351 (this PR)** → #1352 (agnocastlib data layer) → #1353 (discovery agent: connect everything)

## How was this PR tested?

- [ ] Autoware
- [ ] sample application
- [x] Added automated tests (run locally on this branch, all pass):
  - `colcon test --packages-select ros2agnocast` → 17 pytest cases in `test/test_discovery.py` covering gossip collection, the per-verb projection helpers (`all_topic_names` / `all_nodes` / `topic_endpoints` / `topics_of_node`), the per-NS `evaluate_bridge_label` states (enabled / bridged / WARN, incl. the cross-NS no-bridge case), `_resolve_spin_node` for `NodeStrategy` unwrap, and the three branches of `collect_announcements_with_fallback` (gossip success, gossip-empty → ioctl, gossip-empty + ioctl unavailable).
  - `bash scripts/test/e2e_test_cross_ns_cli.bash` → smoke verifying every patched verb accepts `--gossip-timeout` and emits the expected output when an agent + talker are running. (Note: requires a clean `ROS_DOMAIN_ID` partition because of brittleness in DDS state contamination on dev machines — run with e.g. `ROS_DOMAIN_ID=99 bash ...` if your host has stale DDS participants.)

### Manual test: cross-IPC-NS CLI visibility

> **Caveat**: This has only been exercised on a single physical machine (two IPC namespaces on the same host via `unshare`). Cross-ECU verification on real hardware is still outstanding.

This is the user-facing benefit of this PR and **must** be exercised before approving.

```bash
# Prereqs: agnocast kmod loaded; workspace built; three separate shells.

# --- Shell A: IPC NS "alpha" with a talker. The talker launch defaults
#     `discovery_agent` to `false`; pass `:=true` here so the alpha NS
#     gets its own agent (exactly one agent per IPC NS).
sudo -E unshare --ipc bash -c '
  ros2 launch agnocast_sample_application talker.launch.xml discovery_agent:=true &
  sleep 120; wait
'

# --- Shell B: IPC NS "beta" with an agent only (no Agnocast traffic) ---
sudo -E unshare --ipc bash -c '
  ros2 launch ros2agnocast_discovery_agent discovery_agent.launch.xml &
  sleep 120; wait
'

# --- Shell C (host NS): run the CLI verbs and check both NSs show up ---
# Force a fresh ros2 daemon under this shell's env so the new
# `ros2agnocast_discovery_msgs` package is in PYTHONPATH (any ros2 daemon
# inherited from before `install/setup.bash` was sourced will reject the
# resolved type with "invalid message type").
ros2 daemon stop && sleep 1 && ros2 daemon start && sleep 3
ros2 topic list_agnocast            # expect /my_topic listed with "(Agnocast enabled)"
ros2 node  list_agnocast            # expect /talker_node from alpha
ros2 topic info_agnocast /my_topic  # expect "Agnocast Publisher count: 1"
ros2 node  info_agnocast /talker_node  # expect /my_topic in the Pub list
```

> **Caveats**:
> - `sudo -E` preserves the caller's `RMW_IMPLEMENTATION` into the unshared bash. Without `-E` the agent falls back to the default RMW (Fast-DDS on most systems) while the host shell uses Cyclone DDS — discovery still works via RTPS UDP, but data delivery between mixed RMW implementations is unreliable.
> - The `ros2 daemon stop && start` step is required once per fresh shell, after sourcing `install/setup.bash`, so the ros2 daemon's PYTHONPATH includes the new discovery msg package.

Optional cross-ECU run: replace shells A and B with two separate ECUs sharing the same `ROS_DOMAIN_ID`. Same expectations apply.

### Ioctl fallback (no agent on the domain)

If no `discovery_agent` is gossiping on the domain, every verb falls back to a synthetic snapshot built from `libagnocast_ioctl_wrapper`, scoped to the caller's NS. A NOTE is printed to stderr; CLI output is the same as before this PR.

```bash
# Same shell, no agent running, no `--gossip-timeout`:
ros2 launch agnocast_sample_application talker.launch.xml discovery_agent:=false &
sleep 3
ros2 topic list_agnocast
# expected stderr:
#   NOTE: no /_agnocast_discovery agent visible; showing local NS only via ioctl.
#   Start one with `ros2 run ros2agnocast_discovery_agent discovery_agent`
#   to see other NSes / ECUs.
# expected stdout: /my_topic listed with "(Agnocast enabled)"
```

CLI-only changes plus a new helper module. No `agnocastlib` / kmod / heaphook / MQ / ABI changes. See CONTRIBUTING.md §"Versioning rules".

- `need-patch-update`


src/ros2agnocast/package.xml
```diff
@@ -22,6 +22,8 @@
 
   <exec_depend>agnocast_ioctl_wrapper</exec_depend>
   <exec_depend>ament_index_python</exec_depend>
+  <exec_depend>rclpy</exec_depend>
+  <exec_depend>ros2agnocast_discovery_msgs</exec_depend>
```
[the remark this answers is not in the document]
k1832: Done — added `<test_depend>` entries for `python3-pytest`, `ament_copyright`, `ament_flake8`, and `ament_pep257`, matching the set already declared by the neighbour `ros2agnocast_discovery_agent`. Now `rosdep install` brings the pytest runner alongside `ament_python` for the suite added in this PR.

src/ros2agnocast/ros2agnocast/discovery.py
```diff
@@ -0,0 +1,175 @@
...
+        return
+    if snapshots:
+        return
+    print(
+        'WARNING: no /_agnocast_discovery announcements received within '
+        f'{timeout_sec:.1f}s — cross-namespace / cross-ECU Agnocast endpoints '
+        "will not appear. Start the discovery agent in this IPC namespace via "
+        '`ros2 run ros2agnocast_discovery_agent discovery_agent`, or pass '
+        '`--gossip-timeout 0` to skip this check.',
+        file=sys.stderr)
```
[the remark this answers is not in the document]
k1832: Done — reworded `warn_if_no_announcements()`. The new text drops the same-NS implication and instead lists the realistic causes: (1) no discovery agent in this `ROS_DOMAIN_ID`, (2) the ros2 daemon was started before `install/setup.bash` was sourced and is missing the discovery msg package on its `PYTHONPATH` (suggests `ros2 daemon stop && start`), (3) QoS / type mismatch. The docstring was updated to match. Test `test_warn_if_no_announcements_emits_hint_when_empty` now asserts the wording change.

src/ros2agnocast/ros2agnocast/verb/list_agnocast.py
```diff
@@ -112,7 +128,17 @@ def remove_service_topic(topic_names):
                 lib.free_agnocast_topics(agnocast_topic_array, topic_count)
 
             agnocast_topics = remove_service_topic(agnocast_topics)
-            
+
+            # Merge in topics seen via /_agnocast_discovery gossip (other IPC
+            # namespaces and other ECUs in the same ROS_DOMAIN_ID).
+            raw_snapshots = collect_announcements(
```
Koichi98: Seems like the agent already reads the local NS and publishes it as `AgnocastDaemonState`. So the list/info verbs (not only `topic list_agnocast`) might be able to drop their own ioctl path.

Maybe gossip first, ioctl as fallback? node info_agnocast's service detection would probably still need ioctl.

Not necessary needed for #1351 — could just be a later PR. Mostly wanted to ask: was this considered?

src/ros2agnocast/ros2agnocast/discovery.py
```diff
@@ -0,0 +1,181 @@
...
+    and want enough history buffer to absorb a burst at startup. Mismatched
+    ``depth`` does not affect QoS compatibility (it's a per-endpoint
+    setting).
+    """
+    return QoSProfile(
+        reliability=ReliabilityPolicy.RELIABLE,
+        durability=DurabilityPolicy.TRANSIENT_LOCAL,
+        history=HistoryPolicy.KEEP_LAST,
+        depth=64,
+        liveliness=LivelinessPolicy.AUTOMATIC,
```
Koichi98, line 61: How about using `MANUAL_BY_TOPIC`?

src/ros2agnocast/ros2agnocast/verb/list_agnocast.py
```diff
@@ -112,7 +128,17 @@ def remove_service_topic(topic_names):
...
+                node, timeout_sec=args.gossip_timeout)
+            warn_if_no_announcements(raw_snapshots, args.gossip_timeout)
+            snapshots = filter_fresh(raw_snapshots, node=node)
+            for name in all_topic_names(snapshots):
+                if not name.startswith('/AGNOCAST_SRV_'):
```
Koichi98: Can we use this?
https://github.com/autowarefoundation/agnocast/pull/1351/changes#diff-282ef76e7d348cd7d8d6b5cc0a5f2a2f4ceace61bf14494e6d7657f39af38690L100

src/ros2agnocast/ros2agnocast/verb/topic_info_agnocast.py
```diff
@@ -36,6 +44,12 @@ def add_arguments(self, parser, cli_name):
             '-d',
             action='store_true',
             help='Include internal bridge nodes (agnocast_bridge_node_*) in the output')
+        parser.add_argument(
+            '--gossip-timeout',
```
Koichi98: This flag will go away once we switch to ros2-daemon-driven discovery, so let's hide it (`help=argparse.SUPPRESS`) and warn on use, treating it as an unsupported API

src/ros2agnocast/ros2agnocast/verb/list_agnocast.py
Koichi98: It might be good keep the ioctl path while the daemon stabilizes, but drop the local-NS snapshot in `collect_announcements` (same `(host_uuid, ipc_ns_inode)` filter the agent uses). Then ioctl and gossip don't overlap and we can remove the per-verb dedup entirely.

src/ros2agnocast/ros2agnocast/verb/topic_info_agnocast.py
Koichi98: I suggest adding the following in the PR Description as a TODO: replace the fixed `--gossip-timeout` wait with a daemon-driven stop condition. The CLI would query the ros2 daemon for the publisher count on `/_agnocast_discovery` and return once a snapshot has been received from each, with `--gossip-timeout` kept only as a safety fallback.

src/ros2agnocast/ros2agnocast/discovery.py
```diff
@@ -0,0 +1,181 @@
...
+    on the same ``ROS_DOMAIN_ID`` with matching QoS / message type — it
+    does *not* have to live in the same IPC namespace as the CLI. The
+    warning lists the common causes without implying a local-agent
+    requirement. Best-effort: does not change exit codes.
+    """
+    if timeout_sec <= 0:
+        return
+    if snapshots:
+        return
+    print(
```
Koichi98: Could call `get_publishers_info_by_topic` to split this into "no publisher visible" vs "publishers visible but we received nothing". This is local rmw read so there is no big costs.

src/ros2agnocast/ros2agnocast/discovery.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def on_msg(msg: AgnocastDaemonState) -> None:
+        snapshots[(msg.host_uuid, msg.ipc_ns_inode)] = msg
+
+    sub = node.create_subscription(
+        AgnocastDaemonState, GOSSIP_TOPIC, on_msg, gossip_qos())
+    try:
+        deadline = time.monotonic() + timeout_sec
+        while time.monotonic() < deadline:
+            rclpy.spin_once(node, timeout_sec=0.05)
```
Koichi98: Seems like we are passing `NodeStrategy` directly to `spin_once`. `NodeStrategy` is just a dispatcher that routes between the daemon and its own `DirectNode` — it is not an `rclpy.Node`, so passing it directly to an executor to spin is outside its intended use. The
  underlying node should be extracted (e.g. `node_strategy.direct_node.node`) and spun instead.

src/ros2agnocast/ros2agnocast/verb/topic_info_agnocast.py
```diff
@@ -203,10 +256,13 @@ def main(self, *, args):
 
             # check if topic exists
             if not topic_types:
-                if sub_topic_info_ret_count.value == 0 and pub_topic_info_ret_count.value == 0:
+                if (sub_topic_info_ret_count.value == 0
+                        and pub_topic_info_ret_count.value == 0
+                        and not pub_topic_info_rets
+                        and not sub_topic_info_rets):
```
Koichi98: Do we need to add these checks in this PR?
>  and not pub_topic_info_rets
    and not sub_topic_info_rets):

src/ros2agnocast/ros2agnocast/verb/node_list_agnocast.py
Koichi98:
```suggestion
                total_nodes = len(merged_node_name)
```
Could you fix this as above? This is the bug we have had from before this PR, but nice to be fixed along.

src/ros2agnocast/ros2agnocast/discovery.py
```diff
@@ -0,0 +1,181 @@
...
+    finally:
+        node.destroy_subscription(sub)
+
+    return list(snapshots.values())
+
+
+def warn_if_no_announcements(snapshots: list, timeout_sec: float) -> None:
```
Koichi98: This runs before `filter_fresh`, so "received but all stale" produces an empty merge with no warning. Might be better to consider checking post-filter and noting the dropped-stale count.
Such as

```
raw = collect_announcements(node, timeout_sec=args.gossip_timeout)
snapshots = filter_fresh(raw, node=node)
warn_if_no_fresh_announcements(snapshots, raw, args.gossip_timeout)
```

src/ros2agnocast/ros2agnocast/verb/list_agnocast.py
Koichi98: `get_bridge_status()` probably needs to be updated to also factor in gossip endpoints. Fine-grained classification isn't necessary — just a flag indicating whether non-shared-memory traffic is happening on an agnocast-enabled topic. Whether to source this from ioctl or from gossip should align with what we settle on in https://github.com/autowarefoundation/agnocast/pull/1351/changes#r3289075550.

src/ros2agnocast/ros2agnocast/verb/list_agnocast.py
k1832: Done (option 2). `collect_announcements()` drops snapshots from the local `(host_uuid, ipc_ns_inode)`; per-verb dedup in `topic_info_agnocast.py` is gone since ioctl and gossip are disjoint by construction.

src/ros2agnocast/ros2agnocast/discovery.py
k1832: Done — `MANUAL_BY_TOPIC` on both publisher and subscriber. `publish()` is the per-topic assertion, so no extra code needed.

src/ros2agnocast/ros2agnocast/verb/list_agnocast.py
k1832: Done — using `remove_service_topic`.

src/ros2agnocast/ros2agnocast/verb/topic_info_agnocast.py
k1832: Done — `argparse.SUPPRESS` + stderr WARN on explicit use, extracted into `add_gossip_timeout_arg` / `warn_if_gossip_timeout_overridden` and applied to all four verbs. PR description has the TODO for the ros2-daemon-driven replacement.

src/ros2agnocast/ros2agnocast/discovery.py
k1832: Done — `warn_if_no_announcements` now splits between "no publisher visible" and "publisher visible but no snapshot received" via `get_publishers_info_by_topic`.

src/ros2agnocast/ros2agnocast/discovery.py
k1832: Done — added `_resolve_spin_node()` that unwraps `NodeStrategy.direct_node.node`.

src/ros2agnocast/ros2agnocast/verb/topic_info_agnocast.py
k1832: Needed — `pub_topic_info_rets` / `sub_topic_info_rets` now hold ioctl + gossip endpoints, so without these checks a topic visible only via remote-NS gossip would be reported as `'Unknown topic'`. Keeping.

src/ros2agnocast/ros2agnocast/verb/node_list_agnocast.py
k1832: Done.

src/ros2agnocast/ros2agnocast/discovery.py
k1832: Resolved by D1's follow-up — `filter_fresh` / `is_stale` are gone (they relied on the removed `msg.timestamp`); the DDS Liveliness lease bounds staleness.

src/ros2agnocast/ros2agnocast/verb/list_agnocast.py
k1832: Done (coarse). `get_bridge_status` OR's in `is_bridge` from gossip via a new `gossip_has_bridge_endpoint()`. Fine-grained per-NS classification deferred to a follow-up after #1353.

src/ros2agnocast/ros2agnocast/discovery.py
```diff
@@ -0,0 +1,217 @@
+"""Shared helper for subscribing to /_agnocast_discovery and aggregating gossip.
+
+The local NS is excluded from gossip so the CLI does not double-count
+what it already reads via ioctl. Staleness is bounded by the publisher's
+DDS Liveliness lease; no separate freshness filter.
+"""
```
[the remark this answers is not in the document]
k1832: Good catch — the PR description was the side that was stale. Updated the body to drop the "snapshots older than 10 s are filtered as stale" / `--include-stale` flag / "freshness filtering" mentions and to spell out what we actually do today: staleness is bounded by the publisher's DDS Liveliness lease (`AUTOMATIC`, 30 s), no separate CLI-side freshness filter. The module docstring you pointed to is now consistent with the PR description.

src/ros2agnocast/ros2agnocast/discovery.py
```diff
@@ -0,0 +1,217 @@
...
+    else:
+        print(
+            f'WARNING: /_agnocast_discovery has {len(publishers)} publisher(s) '
+            f'visible but no snapshot was received in {timeout_sec:.1f}s. Common '
+            'causes: (1) the ros2 daemon was started before `install/setup.bash` '
+            'was sourced and is missing the discovery msg package on its '
+            'PYTHONPATH (try `ros2 daemon stop && ros2 daemon start`); '
+            '(2) QoS / type mismatch on the publisher side. Pass '
+            '`--gossip-timeout 0` to skip this check.',
+            file=sys.stderr)
```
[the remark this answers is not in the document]
k1832: Fixed in `ee0a9aef` (initial) + `9bfbe661` (tightened). `collect_announcements` now returns `(remote_snapshots, saw_local)`; `warn_if_no_announcements` takes the new `saw_local` and suppresses the "publisher visible but no snapshot" branch **only when `len(publishers) == 1`** — i.e. the single visible publisher is our own NS agent. With 2+ visible publishers and no remote snapshot at least one remote agent is silent, so the warning still fires there (verified by the new `test_warn_if_no_announcements_warns_when_remote_publisher_is_silent`). The benign single-NS case is covered by `test_warn_if_no_announcements_silent_when_only_local_publisher_visible`.

scripts/test/e2e_test_cross_ns_cli.bash
```diff
@@ -0,0 +1,115 @@
...
+# This script does not actually create multiple IPC namespaces (that
+# requires sudo unshare); instead it runs the daemon and the CLI in the
+# same namespace and checks that:
+#
+#   * Each verb accepts `--gossip-timeout`.
+#   * Each verb finishes without raising when the daemon is publishing.
+#   * `topic list_agnocast` surfaces an Agnocast topic via gossip even
+#     when the caller's own ioctl path is the only source (single-NS
+#     baseline; cross-NS visibility is exercised by manual unshare-based
+#     tests).
```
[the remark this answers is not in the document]
k1832: Fixed the header in `ee0a9aef`. New wording drops the "surfaces an Agnocast topic via gossip" claim and explicitly notes that `collect_announcements()` drops the caller's own NS, so the topic/node/endpoint output in this single-NS script comes from the ioctl path; cross-NS gossip visibility is exercised by separate manual unshare-based tests. Did not force `GOSSIP_TIMEOUT=0` since the #30 fix already silences the warning in this single-NS setup, and keeping `--gossip-timeout > 0` still exercises the flag plumbing the script advertises.

Koichi98:
@k1832 

> ros2 topic list_agnocast            # expect /my_topic listed with "(Agnocast enabled)"
> ros2 node  list_agnocast            # expect /talker_node from alpha
> ros2 topic info_agnocast /my_topic  # expect "Agnocast Publisher count: 1"
> ros2 node  info_agnocast /talker_node  # expect /my_topic in the Pub list

I've checked that these commands ran successfully as expected, even the endpoints exists across ECUs.

TetsuKawa:
@k1832 
Just to confirm, is it correct to assume that this PR does not include any changes related to the bridge, and its sole purpose is to collect, merge, and display all information from each NS and ECU?

---

When running a talker in Shell A, and not just an agent but also a listener in Shell B, the output for `/my_topic` becomes as follows:

```bash
❯ ros2 topic info_agnocast -v /my_topic
Type: <UNKNOWN>

ROS 2 Publisher count: 0
Agnocast Publisher count: 1

Node name: talker_node
Node namespace: /
Topic type: <UNKNOWN>
Endpoint type: PUBLISHER (Agnocast enabled)
QoS profile:
  History (Depth): KEEP_LAST (1)
  Durability: VOLATILE

ROS 2 Subscription count: 0
Agnocast Subscription count: 1

Node name: listener_node
Node namespace: /
Topic type: <UNKNOWN>
Endpoint type: SUBSCRIPTION (Agnocast enabled)
QoS profile:
  History (Depth): KEEP_LAST (1)
  Durability: VOLATILE
```

This output could potentially mislead users into thinking that the Pub/Sub communication is actually established between them.

Therefore, I believe we should differentiate topics with the same name if they belong to different IPC namespaces. What do you think about displaying them distinctively in this PR, and then reverting to the combined view in the future once the bridge functionality is developed and those topics can genuinely be treated as identical?

k1832:
@TetsuKawa 
> @k1832 Just to confirm, is it correct to assume that this PR does not include any changes related to the bridge, and its sole purpose is to collect, merge, and display all information from each NS and ECU?

Yes.

> This output could potentially mislead users into thinking that the Pub/Sub communication is actually established between them.
> 
> Therefore, I believe we should differentiate topics with the same name if they belong to different IPC namespaces. What do you think about displaying them distinctively in this PR, and then reverting to the combined view in the future once the bridge functionality is developed and those topics can genuinely be treated as identical?

That's a very reasonable concern. However, since we're aiming to support the cross-ECU/cross-NS bridge generation in very near future (~ one month), we can probably leave the CLI output as is for now. If we want to fill the gap just for this small time window, we would have to pay more cost than the benefit (e.g. this PR gets move complex, having to revert that code in a month).

k1832:
Follow up to the comment above:

We will update the user doc to avoid this confusion
> This output could potentially mislead users into thinking that the Pub/Sub communication is actually established between them.

Koichi98:
@k1832 @TetsuKawa 
I'd like to propose a refinement to the semantics of the `bridged` label / WARN in `list_agnocast` and `node_info_agnocast`, as a stepping stone toward the upcoming cross-NS / cross-ECU bridge support.Bridge presence is evaluated by OR-aggregating across all snapshots currently, which breaks down when the same topic name is observable from more than one NS. 

Let me propose the semantics for the 'bridged' label onwards.
- **`(Agnocast enabled, bridged)`** — Every NS that participates in a required data flow has the bridge(s) it locally needs.
- **`(WARN: bridge is not active)`** — Some NS lacks a bridge required by its own non-bridge endpoints.
- **`(Agnocast enabled)`** — No bridging is required.

The case @TetsuKawa raised — a talker and a listener living in different NSes shown as if they were communicating on the same topic — surfaces **automatically as a WARN** under this semantics:

So we keep the current "one line per topic name" presentation, and the misleading scenario is no longer silent — without needing a separate doc note for this case.

### Toward future cross-ns, cross-ECU bridging
Future cross-NS / cross-ECU bridging is presumably going to look like "each NS hosts its own bridge, and DDS stitches them together". The per-NS semantics match that:
- **Today (no cross-NS bridge)**: Multi-NS instances of the same topic **predominantly produce WARN** — correctly reporting that data isn't flowing across NSes in the current configuration.
- **Near future (cross-NS bridge supported)**: The same reachability rule naturally returns `(bridged)`. No relabeling and no restructuring of the decision logic are needed.

Conversely, the current OR aggregation can't distinguish "fully wired cross-NS bridge" from "half-wired with a bridge on only one side", so we'd have to revisit `bridged` semantics in the cross-NS bridge PR anyway.

k1832:
@Koichi98 @TetsuKawa Thank you for the comments!

> Let me propose the semantics for the 'bridged' label onwards.
> 
> * **`(Agnocast enabled, bridged)`** — Every NS that participates in a required data flow has the bridge(s) it locally needs.
> * **`(WARN: bridge is not active)`** — Some NS lacks a bridge required by its own non-bridge endpoints.
> * **`(Agnocast enabled)`** — No bridging is required.
> 
> The case @TetsuKawa raised — a talker and a listener living in different NSes shown as if they were communicating on the same topic — surfaces **automatically as a WARN** under this semantics:

This plan looks good. Let me implement this labeling in this PR!

src/ros2agnocast/ros2agnocast/verb/list_agnocast.py
```diff
@@ -1,75 +1,55 @@
...
-                    for n in nodes:
-                        if n.is_bridge:
-                            has_pub_bridge = True
-                        else:
-                            has_agnocast_pub = True
+                for snap in snapshots:
+                    for topic in snap.topics:
+                        if topic.topic_name != topic_name:
+                            continue
+                        for n in topic.subscribers:
```
TetsuKawa: Instead of running the triple-nested loop for each topic every time, how about scanning the snapshots just once beforehand to build a dictionary mapping topic names to their results?This would optimize the lookup time to $O(1)$ and significantly speed up the command, especially when there are many topics.

k1832: Good catch. I'd like to apply this change, but since we're a bit hurry to this PR, do you mind if I just add a TODO comment for this, and apply the change in a follow up?

src/ros2agnocast/ros2agnocast/discovery.py
```diff
@@ -185,18 +185,55 @@ def topics_of_node(snapshots: list, node_name: str) -> tuple:
...
-    return has_pub_bridge, has_sub_bridge
+            real_pub = any(not ep.is_bridge for ep in topic.publishers)
+            real_sub = any(not ep.is_bridge for ep in topic.subscribers)
+            if not (real_pub or real_sub):
+                continue
+            inbound_bridge = any(ep.is_bridge for ep in topic.publishers)
+            outbound_bridge = any(ep.is_bridge for ep in topic.subscribers)
+            ns_roles.append((real_pub, real_sub, inbound_bridge, outbound_bridge))
+
+    required = len(ns_roles) >= 2 or (len(ns_roles) >= 1 and ros2_has_endpoint)
```
Koichi98: The current `evaluate_bridge_label` raises WARN whenever a real endpoint lacks its bridge once `required` is True. But if no peer actually needs the data, no bridge is expected to spawn — and the WARN is misleading. For example:

- NS-A has `agnocast_pub`, ROS 2 has only a publisher (no subscriber), no bridge in NS-A → currently WARN, but no bridge should be running here.

Could we make the per-NS check direction-aware?
- A `real_pub` needs an outbound bridge only if some other NS or ROS 2 has a non-bridge subscriber.
- A `real_sub` needs an inbound bridge only if some other NS or ROS 2 has a non-bridge publisher.

That way WARN only fires when a bridge is actually expected to be running but isn't.

src/ros2agnocast/ros2agnocast/verb/list_agnocast.py
TetsuKawa: Sounds good! Go ahead and add a TODO. We can handle it in a follow-up PR

k1832: There was actually another critical comment, so I applied this change too!
https://github.com/autowarefoundation/agnocast/pull/1351/commits/14e53d83b42acf3c299532473aafb3ea1ee41ce1

src/ros2agnocast/ros2agnocast/discovery.py
k1832: Done
https://github.com/autowarefoundation/agnocast/pull/1351/commits/b5dc618ee172249b2a7424942adb9b2bfe3815d9

Koichi98 (approved):

TetsuKawa (approved):

PR #1362, opened 2026-05-28
fix(scripts): fix logic for sucessful installation for `switch_kmod.bash`

Koichi98:
## Description
Fix the post-modprobe verification in `switch_kmod.bash` Step 5.

The old logic snapshotted `dmesg | wc -l` before modprobe and used `tail -n +N` to isolate new lines. Steps 1–4 (unload / apt purge / dkms remove / apt install) take ~10s+, during which the kernel ring buffer can roll. When old lines are evicted, the snapshotted absolute line number exceeds the current dmesg length, `tail -n +N` returns empty, and verification fails even though `Agnocast installed! vX` is present.

Since Steps 1–4 unload the old module, purge every `agnocast-kmod-v*` package, and install exactly one target package, the last `Agnocast installed! vX` line in dmesg is guaranteed to be from this modprobe. Replaced the diff with a simple `grep | tail -n 1`.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

**Important notes:**

### DIFF
--- a/scripts/switch_kmod.bash
+++ b/scripts/switch_kmod.bash
@@ -172,21 +172,18 @@ sudo apt-get install -y "${target_package}"
 
 echo "[5/5] Load agnocast and verify"
 
-# Snapshot dmesg line count so we can isolate messages emitted by THIS
-# modprobe (avoids mistaking a stale 'Agnocast installed!' line from an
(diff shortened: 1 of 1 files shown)


ruth561 (approved):
LGTM

PR #1363, opened 2026-05-28
feat(ros2agnocast): add discovery_daemon_status verb for daemon liveness

k1832:
### Background
#1350 added the *discovery agent*, one rclpy process per IPC namespace (NS) that publishes the local Agnocast state on `/_agnocast_discovery`. When the agent isn't running (or is running in a different NS than the operator expects), cross-NS observability silently stops working — `ros2agnocast` CLI verbs that subscribe to the gossip simply see fewer endpoints, with no error.

### What's added
Adds `ros2 agnocast discovery-daemon-status`, a liveness check for the **current** IPC namespace:

| verdict (exit) | default | `--verbose` |
|---|---|---|
| **OK** — running (0) | `OK. The discovery agent is running.` | `IPC namespace inode: <inode>`<br>`  process:       OK (holds the singleton lock)`<br>`  gossip:        OK (snapshot received on /_agnocast_discovery)`<br>`  type_registry: 1 live registration(s) …`<br>_(blank)_<br>`OK. The discovery agent is running.` |
| **NG** — not running (1) | `NG. The discovery agent is not running.` | `IPC namespace inode: <inode>`<br>`  process:       NG (singleton lock is free …)`<br>`  gossip:        skipped (agent not running)`<br>`  type_registry: no … registered yet … (1 stale …)`<br>_(blank)_<br>`NG. The discovery agent is not running.` |
| **NG** — running but not publishing (1) | `NG. The discovery agent is running but not publishing.` | `IPC namespace inode: <inode>`<br>`  process:       OK (holds the singleton lock …)`<br>`  gossip:        NG (no snapshot within 3.0s)`<br>`  type_registry: …`<br>_(blank)_<br>`NG. The discovery agent is running but not publishing.` |

The verdict comes from two internal checks:

| Check | What it asserts |
|---|---|
| **process** | the agent is alive — it holds an exclusive `flock(2)` on its per-NS singleton lock (`${AGNOCAST_TMPFS_DIR:-/dev/shm}/agnocast_discovery_agent_<ipc_ns_inode>.lock`) for its whole lifetime, so we probe that lock. |
| **gossip** | an `AgnocastDaemonState` tagged with **this** namespace's `ipc_ns_inode` is received on `/_agnocast_discovery` within the timeout (snapshots from other namespaces sharing the topic don't count). `gossip` OK implies `process` OK, so it is the primary signal; `process` only distinguishes "not running" from "running but not publishing". |
| **type_registry** (info, `--verbose` only) | how many live Agnocast processes have registered under `${AGNOCAST_TMPFS_DIR:-/dev/shm}/agnocast_type_registry/<ipc_ns_inode>/`. Informational — an empty registry just means nothing has registered yet, and never fails. |

The verb is intentionally per-NS — it inspects only the IPC namespace it runs in, so operators run it inside the namespace they want to interrogate.

## Related links

- Jira: T4DEV-51976 (TIER IV internal)
- Builds on (already merged):
  - **#1356** — adds the tmpfs type registry. The verb's `type_registry` line looks for files the writer in `#1356` creates.

## How was this PR tested?

- [ ] Autoware
- [ ] `bash scripts/test/e2e_test_1to1.bash` — N/A: no change to the pub / sub path.
- [ ] `bash scripts/test/e2e_test_2to2.bash` — N/A: same reason.
- [ ] kunit tests — N/A: kmod untouched.
- [ ] `bash scripts/test/run_requires_kernel_module_tests.bash` — N/A.
- [ ] sample application
- [x] Added automated tests, all pass locally (`colcon test --packages-select ros2agnocast`). The new test file covers the flock-probe liveness check (held / free / missing lock), the informational `type_registry` description (live / stale / empty / missing), the path helpers, and the verdict logic (running / not running / running-but-not-publishing, default vs `--verbose`).

## Notes for reviewers — manual liveness test

Manual run-through, all on one host:

```bash
cd ~/agnocast

# --- Happy path: a talker (so type_registry has a writer) + a discovery
#     agent, both in the current IPC namespace ------------------------------
ros2 launch agnocast_sample_application talker.launch.xml discovery_agent:=true \
    > /dev/null 2>&1 &
sleep 6
ros2 agnocast discovery-daemon-status            # default: single verdict
ros2 agnocast discovery-daemon-status --verbose  # + inode / per-check / type_registry

# --- Agent down: stop the launch tree (its discovery agent dies with it) ---
kill %1 && wait %1 2>/dev/null
sleep 2
ros2 agnocast discovery-daemon-status            # default: single verdict
ros2 agnocast discovery-daemon-status --verbose  # + inode / per-check / type_registry
```

Observed output:

```
# Happy path — default
OK. The discovery agent is running.

# Happy path — --verbose
IPC namespace inode: 4026531839
  process:       OK (holds the singleton lock (/dev/shm/agnocast_discovery_agent_4026531839.lock))
  gossip:        OK (snapshot received on /_agnocast_discovery)
  type_registry: 1 live registration(s) in /dev/shm/agnocast_type_registry/4026531839

OK. The discovery agent is running.

# Agent down — default
NG. The discovery agent is not running.

# Agent down — --verbose
IPC namespace inode: 4026531839
  process:       NG (singleton lock is free (/dev/shm/agnocast_discovery_agent_4026531839.lock))
  gossip:        skipped (agent not running)
  type_registry: no Agnocast process has registered yet in /dev/shm/agnocast_type_registry/4026531839 (1 stale <pid>.txt awaiting daemon cleanup)

NG. The discovery agent is not running.
```

Cross-NS variant: run the talker inside `sudo -E unshare --ipc bash -c '...'` and observe that `ros2 agnocast discovery-daemon-status` run *outside* that NS reports `NG` (the agent's singleton lock for the host NS is free / absent).

- `need-patch-update`

Additions only — one new verb file, one new test file, one entry-point line in `setup.py`. No changes to kmod ABI, existing ioctls, public `agnocastlib` API, or wire structs.

### DIFF
--- a/src/ros2agnocast/ros2agnocast/verb/discovery_daemon_status.py
+++ b/src/ros2agnocast/ros2agnocast/verb/discovery_daemon_status.py
@@ -5,17 +5,20 @@ for cross-namespace observability. If the agent is not running (or
 running in a different IPC namespace), that observability silently stops
 working — this verb gives the operator a single place to confirm liveness.
 
+This command only inspects the **current** IPC namespace (the one the
+command itself runs in); run it inside the namespace you want to check.
+
 Checks performed (each prints OK / NG with detail):
 
-  * **process** — any `ros2agnocast_discovery_agent` process is running
-    in **this** IPC namespace (matched by `/proc/<pid>/ns/ipc`).
-  * **gossip** — `/_agnocast_discovery` has at least one DDS publisher
-    on the current `ROS_DOMAIN_ID` and a snapshot is received within
-    the timeout.
+  * **process** — any ``ros2agnocast_discovery_agent`` process is running
+    in **this** IPC namespace (matched by ``/proc/<pid>/ns/ipc``).
+  * **gossip** — a snapshot from **this** IPC namespace is received on
+    ``/_agnocast_discovery`` within the timeout (snapshots from other
+    namespaces sharing the topic don't count).
   * **type_registry** — the tmpfs directory
-    `/dev/shm/agnocast_type_registry/<ipc_ns_inode>/` exists and contains
-    at least one ``<pid>.txt`` (proves at least one Agnocast process has
-    registered).
+    ``${AGNOCAST_TMPFS_DIR:-/dev/shm}/agnocast_type_registry/<ipc_ns_inode>/``
+    holds at least one ``<pid>.txt`` whose process is still alive (proves
+    a live Agnocast process has registered).
 
 Exit code:
 
@@ -24,20 +27,20 @@ Exit code:
 """
 
 import os
-import time
 
-import rclpy
 from ros2cli.node.strategy import NodeStrategy
 from ros2cli.verb import VerbExtension
 
-from ros2agnocast_discovery_msgs.msg import AgnocastDaemonState
-
-
-_GOSSIP_TOPIC = '/_agnocast_discovery'
+from ros2agnocast.discovery import (
+    add_gossip_timeout_arg,
+    collect_announcements,
+    GOSSIP_TOPIC,
+    warn_if_gossip_timeout_overridden,
+)
 
 
 def _type_registry_base() -> str:
-    """Resolve the tmpfs root, honoring `AGNOCAST_TMPFS_DIR` like the writer."""
+    """Resolve the tmpfs root, honoring ``AGNOCAST_TMPFS_DIR`` like the writer."""
     root = os.environ.get('AGNOCAST_TMPFS_DIR') or '/dev/shm'
     return os.path.join(root, 'agnocast_type_registry')
 
@@ -52,6 +55,7 @@ def _check_daemon_process(my_ns_inode):
     for pid_str in os.listdir('/proc'):
         if not pid_str.isdigit():
             continue
+
         pid = int(pid_str)
         # The discovery agent runs as a python script — its ``comm`` is the
         # python interpreter — so match on cmdline instead.
@@ -60,6 +64,7 @@ def _check_daemon_process(my_ns_inode):
                 cmdline = fp.read().replace(b'\0', b' ').decode('utf-8', errors='replace')
         except (FileNotFoundError, PermissionError):
             continue
+
         # Match the actual agent binary path so we don't also pick up
         # ``ros2 run ros2agnocast_discovery_agent discovery_agent``
         # wrapper processes (whose cmdline contains the same package name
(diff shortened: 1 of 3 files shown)


src/ros2agnocast/ros2agnocast/verb/discovery_daemon_status.py
```diff
@@ -0,0 +1,150 @@
...
+    return True, f'{ns_dir} has {len(files)} registration file(s)'
+
+
+class DiscoveryDaemonStatusVerb(VerbExtension):
+    """Check the per-IPC-namespace Agnocast discovery agent's liveness."""
+
+    def add_arguments(self, parser, cli_name):
+        parser.add_argument(
+            '--gossip-timeout', type=float, default=2.0,
+            help='Seconds to wait for a /_agnocast_discovery snapshot (default: 2.0)')
```
ruth561: Should we use `add_gossip_timeout_arg()`?

src/ros2agnocast/setup.py
```diff
@@ -28,6 +28,7 @@
         'ros2agnocast.verb': [
             'generate-bridge-plugins = ros2agnocast.verb.generate_bridge_plugins:GenerateBridgePluginsVerb',
             'version = ros2agnocast.verb.version:VersionVerb',
+            'discovery_daemon_status = ros2agnocast.verb.discovery_daemon_status:DiscoveryDaemonStatusVerb',
```
ruth561: I noticed an inconsistency in the naming conventions. Some use hyphens (`-`), while others use underscores (`_`) (e.g., `ros2 agnocast generate-bridge-plugins` and `ros2 topic list_agnocast`).

Since renaming might require a `need-major-update`, we should decide on this carefully. How about using hyphens (`-`) here to align with the `ros2 agnocast` command's naming convention?

src/ros2agnocast/ros2agnocast/verb/discovery_daemon_status.py
```diff
@@ -0,0 +1,150 @@
...
+
+    ``NodeStrategy`` initializes rclpy itself; do not call ``rclpy.init``
+    here or the second call raises ``Context.init() must only be called once``.
+    """
+    with NodeStrategy(None) as node:
+        from ros2agnocast.discovery import gossip_qos
+        received = []
+
+        def cb(msg: AgnocastDaemonState) -> None:
+            received.append(msg)
```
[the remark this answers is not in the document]
ruth561: In addition, could you update the CLI help message to mention that this command only supports checking the status of daemons in the current namespace?

src/ros2agnocast/ros2agnocast/verb/discovery_daemon_status.py
```diff
@@ -83,62 +86,78 @@ def _check_daemon_process(my_ns_inode):
...
-        finally:
-            node.destroy_subscription(sub)
-
-        if not received:
-            return False, (
-                f'no AgnocastDaemonState received on {_GOSSIP_TOPIC} within '
-                f'{timeout_sec}s')
-        return True, f'received {len(received)} snapshot(s) on {_GOSSIP_TOPIC}'
+    mine = [s for s in snapshots if s.ipc_ns_inode == my_ns_inode]
+    if mine:
```
k1832: nit: use `any()`?

src/ros2agnocast/ros2agnocast/verb/discovery_daemon_status.py
```diff
@@ -83,62 +86,78 @@ def _check_daemon_process(my_ns_inode):
...
+            live += 1
+        else:
+            stale += 1
+
+    if live == 0:
+        detail = f'{ns_dir} has no registration from a live process'
+        if stale:
+            detail += f' ({stale} stale <pid>.txt awaiting daemon cleanup)'
+        return False, detail
+    return True, f'{ns_dir} has {live} live registration file(s)'
```
k1832: Reverse the condition:
```
if live:
   ....
else:
   ....
```

Also, as shown above, we could leave the `else` there to emphasize the symmetricity to the `true` path for readability

src/ros2agnocast/ros2agnocast/verb/discovery_daemon_status.py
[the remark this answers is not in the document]
k1832: Done (0b7be8c9) — the gossip check now filters snapshots by the current-NS inode, and the module/class docstrings (which back the CLI help) now state the current-NS-only scope.

src/ros2agnocast/ros2agnocast/verb/discovery_daemon_status.py
```diff
@@ -0,0 +1,150 @@
...
+            received.append(msg)
+
+        sub = node.create_subscription(
+            AgnocastDaemonState, _GOSSIP_TOPIC, cb, gossip_qos())
+        try:
+            deadline = time.monotonic() + timeout_sec
+            while time.monotonic() < deadline and not received:
+                rclpy.spin_once(node, timeout_sec=0.05)
+        finally:
+            node.destroy_subscription(sub)
```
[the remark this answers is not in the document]
k1832: Done (0b7be8c9) — switched to `collect_announcements()`, which resolves the `NodeStrategy` to the underlying node before spinning.

src/ros2agnocast/ros2agnocast/verb/discovery_daemon_status.py
```diff
@@ -0,0 +1,150 @@
...
+    """Return (ok, detail) for the tmpfs type registry check."""
+    ns_dir = os.path.join(_type_registry_base(), str(my_ns_inode))
+    if not os.path.isdir(ns_dir):
+        return False, f'directory missing: {ns_dir}'
+    files = [f for f in os.listdir(ns_dir) if f.endswith('.txt')]
+    if not files:
+        return False, f'{ns_dir} contains no <pid>.txt registrations'
+    return True, f'{ns_dir} has {len(files)} registration file(s)'
+
+
```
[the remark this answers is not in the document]
k1832: Done (0b7be8c9)

src/ros2agnocast/ros2agnocast/verb/discovery_daemon_status.py
k1832: Done (0b7be8c9)

src/ros2agnocast/setup.py
k1832: Done (0b7be8c9)

src/ros2agnocast/ros2agnocast/verb/discovery_daemon_status.py
```diff
@@ -0,0 +1,181 @@
...
+            with open(f'/proc/{pid}/cmdline', 'rb') as fp:
+                cmdline = fp.read().replace(b'\0', b' ').decode('utf-8', errors='replace')
+        except (FileNotFoundError, PermissionError):
+            continue
+
+        # Match the actual agent binary path so we don't also pick up
+        # ``ros2 run ros2agnocast_discovery_agent discovery_agent``
+        # wrapper processes (whose cmdline contains the same package name
+        # as an argv slot rather than as the executable path).
+        if '/ros2agnocast_discovery_agent/lib/ros2agnocast_discovery_agent/discovery_agent' \
```
ruth561: In this line, a specific file path is hardcoded. However, when I run the `discovery_agent` executable, the actual path is `/ros2agnocast_discovery_agent/bin/discovery_agent` instead of `/ros2agnocast_discovery_agent/lib/ros2agnocast_discovery_agent/discovery_agent`. Because of this difference, the agent process ID cannot be found.

I am concerned that specifying such a detailed path might cause issues if the usage or environment changes in the future. Could we consider an alternative approach?

I'm not entirely sure what the best way to identify the current IPC-daemon process is, but how about an idea like this?

* When `discovery_agent` starts, it creates a file at `/dev/shm/agnocast_discovery_agents/<namespace>/<pid>.txt`.
* The CLI command reads the files in `/dev/shm/agnocast_discovery_agents/<namespace>/*` to get the PIDs, and pushes them to `found_pids`.

What do you think?

src/ros2agnocast/ros2agnocast/verb/discovery_daemon_status.py
```diff
@@ -0,0 +1,181 @@
...
+        else:
+            stale += 1
+
+    if live:
+        return True, f'{ns_dir} has {live} live registration file(s)'
+
+    detail = f'{ns_dir} has no registration from a live process'
+    if stale:
+        detail += f' ({stale} stale <pid>.txt awaiting daemon cleanup)'
+    return False, detail
```
ruth561: I have a question about `_check_type_registry()`. Should we avoid reporting an error when no Agnocast process has registered yet?

Having zero `<pid>.txt` files doesn't mean the discovery agent is broken. It just means no Agnocast publisher or subscriber has started in this namespace yet.

The current check mixes up the agent's status with actual Agnocast usage, which might confuse users.

Could we change this to an informational message instead of an error?

src/ros2agnocast/ros2agnocast/verb/discovery_daemon_status.py
k1832: Oh, true! Good catch.
To fix this, I'm reusing the flock that's already created for enforcing a single agent per IPC NS. Here is the lifecycle of the lock.
1. The lock file is created when an agent starts
2. The lock is held while the agent is alive
3. The lock is released when the agent dies for whatever reason
4. The lock file itself will be kept in the file system even though the agent is gone

So probing whether the lock is held is a reliable way to check the agent's status.

src/ros2agnocast/ros2agnocast/verb/discovery_daemon_status.py
k1832: > Having zero <pid>.txt files doesn't mean the discovery agent is broken. It just means no Agnocast publisher or subscriber has started in this namespace yet.

That's true. Thanks for the catch. I made the `type_registry` check informational only.

k1832:
I'd like to test the current status.

k1832:
Manual test in my env successfully passed. Making the PR "ready for review" again.

## Manual test

Verified the verb end-to-end in the current IPC namespace — happy path (agent up) and agent-down.

```bash

# --- Happy path -----------------------------------------------------------
# Spawn one talker (so `type_registry` has a writer) and one discovery agent
# in the current IPC NS, then check liveness.
ros2 launch agnocast_sample_application talker.launch.xml discovery_agent:=true \
    > /dev/null 2>&1 &
sleep 5
ros2 agnocast discovery-daemon-status
# → process OK / gossip OK, exit 0

# --- Negative: agent down -------------------------------------------------
# Stop the talker's launch tree (its child discovery agent dies with it) and
# re-check from the same shell.
kill %1 && wait %1 2>/dev/null
sleep 2
ros2 agnocast discovery-daemon-status
# → process NG / gossip NG, exit non-zero
```

Output:

```
# Happy path
IPC namespace inode: 4026531839
  process       OK: agent is running: a live process holds the singleton lock (could not acquire /dev/shm/agnocast_discovery_agent_4026531839.lock)
  gossip        OK: received a snapshot from this IPC namespace on /_agnocast_discovery
  type_registry INFO: 1 live registration(s) in /dev/shm/agnocast_type_registry/4026531839

# After killing the launch tree (agent + talker gone)
IPC namespace inode: 4026531839
  process       NG: agent is NOT running: nothing holds the singleton lock (freely acquired /dev/shm/agnocast_discovery_agent_4026531839.lock)
  gossip        NG: no AgnocastDaemonState received on /_agnocast_discovery within 2.0s
  type_registry INFO: no Agnocast process has registered yet in /dev/shm/agnocast_type_registry/4026531839 (1 stale <pid>.txt awaiting daemon cleanup)
```

Notes:
- `process` flips OK → NG correctly when the agent dies — the singleton `flock` is released by the kernel on process exit, so a probe that *can* acquire it means no live agent.
- `type_registry` is informational only: it reports `1 live registration` while the talker runs, and after the talker exits it reports no live registration plus the leftover `1 stale <pid>.txt` — it never drives the exit code.

src/ros2agnocast/ros2agnocast/verb/discovery_daemon_status.py
```diff
@@ -0,0 +1,189 @@
...
+        print(f'IPC namespace inode: {my_ns_inode}')
+
+        any_ng = False
+        for name, fn in [
+            ('process', lambda: _check_daemon_process(my_ns_inode)),
+            ('gossip', lambda: _check_gossip(my_ns_inode, args.gossip_timeout)),
+        ]:
+            ok, detail = fn()
+            status = 'OK' if ok else 'NG'
+            print(f'  {name:<14}{status}: {detail}')
```
Koichi98: Users of this command don't know Agnocast's internals, so it seems like surfacing the process / gossip checks as separate OK/NG lines by default is more detail than they need. Also, since gossip OK already implies process OK, the two checks aren't independent — gossip is the stronger, end-to-end signal. Can we collapse the default output to one verdict and only breaking down the internals when something is actually wrong:

```
  # healthy
  OK. The discovery agent is running.

  # process is gone
  NG. The discovery agent is not running.   (process: lock free)

  # alive but not publishing
  NG. The discovery agent is running but not publishing.
      (process OK: holds singleton lock / gossip: no snapshot within 3s)
```

  The per-check breakdown still appears on NG, so the "alive but not publishing" vs "dead" distinction isn't lost.

In my opinion, anything beyond liveness — the IPC namespace inode header and the type_registry info line — isn't part of the verdict and shouldn't affect the exit code. What about gating those behind a --verbose flag?

k1832: I agree that the current log format can be confusing or too much detail for users.

> the IPC namespace inode header and the type_registry info line

Right, the inode header and the type_registry check are already out of the verdict scope (neither affects the exit code). I'll move both behind a `--verbose` flag so they're off by default.

> Can we collapse the default output to one verdict and only breaking down the internals when something is actually wrong:

Yes, I'll apply this later. Thanks for the comment!

k1832: Done! The manual test mentioned in the PR passed too.

Koichi98 (approved):

PR #1365, opened 2026-05-29
feat(agnocastlib): add deferred service callback support

### LINKED ISSUE #1323: Asynchronous service callbacks
rclcpp offers an asynchronous form of service callbacks, which is not yet supported in Agnocast. See https://github.com/ijnek/nested_services_rclcpp_demo.

bdm-k:
closes #1323 

## Description

This PR introduces a **deferred service callback** pattern that decouples receiving a service request from sending the response. With deferred service callbacks, it's possible to send responses asynchronously, for example, from a time, an I/O completion handler, or another thread.

### Usage

```cpp
auto service = agnocast::create_service<ServiceT>(
  node, "my_service",
  [](std::shared_ptr<agnocast::Service<ServiceT>> service,
     ipc_shared_ptr<ServiceT::Request> && request) {
    // Respond later from any context (timer, thread, I/O callback, etc.)
    auto response = service->borrow_loaned_response(request);
    response->result = compute(request->data);
    service->send_response(std::move(request), std::move(response));
  }, qos, group);
```

### Implementation details

The `BasicService`'s constructor determines whether the passed callback type is basic or deferred. If it is a basic callback, the subscription callback is generated by `wrap_basic_service_callback_for_subscriber()`; if it is a deferred callback, the subscription callback is generated by `wrap_deferred_service_callback_for_subscriber()`.

Two public member functions are added. They are expected to be called from within the context of the deferred callback:

- `send_response()` sends a deferred response. The `response` parameter must be acquired via `borrow_loaned_response()`. The `request` parameter is also required in order to know which node initiated the service call.
- `borrow_loaned_response()` allocates a response message in shared memory and sets the internal sequence number for correct request-response matching.

## How was this PR tested?

- [ ] Autoware
- [x] sample application

Created a sample application that uses a deferred service callback and verified its functionality.


src/agnocastlib/include/agnocast/agnocast_service.hpp
```diff
@@ -111,14 +104,83 @@ class BasicService
...
+   * @brief Allocate a service response message in shared memory. This function is expected to be
+   * used in deferred response callbacks.
+   * @param request The request that initiated the service call.
+   * @return Owned pointer to the response message in shared memory.
+   */
+  // AGNOCAST_PUBLIC
+  ipc_shared_ptr<typename ServiceT::Response> borrow_loaned_response(
+    ipc_shared_ptr<typename ServiceT::Request> request)
+  {
+    auto internal_request = static_ipc_shared_ptr_cast<RequestT>(std::move(request));
```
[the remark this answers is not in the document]
bdm-k: fixed

src/agnocastlib/include/agnocast/agnocast_service.hpp
```diff
@@ -111,14 +104,83 @@ class BasicService
...
+    NodeT * node, const std::string & service_name, Func && callback, const rclcpp::QoS & qos,
+    rclcpp::CallbackGroup::SharedPtr group)
+  : node_(node),
+    service_name_(node->get_node_services_interface()->resolve_service_name(service_name)),
+    // TransientLocal durability is not allowed for services.
+    qos_(rclcpp::QoS(qos).durability_volatile())
+  {
+    static_assert(
+      is_basic_cb<Func>::value || is_deferred_cb<Func>::value,
+      "Callback must be callable with on of the following argument pairs:\n"
```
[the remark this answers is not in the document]
bdm-k: fixed

src/agnocastlib/include/agnocast/agnocast_service.hpp
```diff
@@ -111,14 +104,83 @@ class BasicService
...
+   */
+  // AGNOCAST_PUBLIC
+  void send_response(
+    ipc_shared_ptr<typename ServiceT::Request> && request,
+    ipc_shared_ptr<typename ServiceT::Response> && response)
+  {
+    auto internal_request = static_ipc_shared_ptr_cast<RequestT>(std::move(request));
+    auto internal_response = static_ipc_shared_ptr_cast<ResponseT>(std::move(response));
+    auto publisher = get_or_create_publisher_for(internal_request->_node_name);
+    publisher->publish(std::move(internal_response));
```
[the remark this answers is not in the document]
bdm-k: This problem is fixed by #1367.

src/agnocastlib/include/agnocast/agnocast_service.hpp
```diff
@@ -111,14 +104,83 @@ class BasicService
...
   }
+
+  /**
+   * @brief Sends a response to the client that initiated the service call. This function is
+   * expected to be used in deferred response callbacks.
+   * @param request The request that initiated the service call.
+   * @param response The response to send. Must be acquired by calling borrow_loaned_response().
+   */
+  // AGNOCAST_PUBLIC
+  void send_response(
```
Koichi98, line 163: Since `borrowed_publisher_num` is thread_local, deferred callback keeps the window of borrow->publish open across threads, so `send_response` from another thread decrements from 0 → `exit(EXIT_FAILURE)`.

We might have to have `send_response` take a finished response and do borrow→copy→publish synchronously inside the call. One copy at send time.

Please also document that the response is copied into shared memory at send time in both the Agnocast user docs and the `send_response` doc comment, if you accept this solution.

bdm-k: `Publisher` makes it the user's responsibility that the borrow-publish pair happens on a single thread. By the same logic, I think it's natural to require the user to ensure the borrow-send pair also happens on the same thread. Plus, in a typical deferred service callback, the response is constructed right before calling `send_response()`. What do you think about just clarifying this in the doc comment?

Koichi98: In my opinion, I don't think the Publisher analogy holds. For Publisher, borrow→publish is synchronous and local: it happens in one function with a local handle, so "same thread" is satisfied automatically unless you deliberately break it. Deferred is asynchronous by design — the response is produced later, often on a worker thread the user doesn't choose. So "borrow and send on the same thread" is a fragile requirement here, and sometimes the user can't even guarantee it.
What do you think?

bdm-k: I may not have made myself clear, but my point is that the flow of borrow→populate→send is typically synchronous even in a deferred service callback. You can see this, for example, in the R2A service bridge implementation. The ROS 2 response is allocated, populated, and sent in a single callback ([code](https://github.com/autowarefoundation/agnocast/blob/bfdcf257035befa188e7fb2dc7bc774ff50668f4/src/agnocastlib/include/agnocast/bridge/agnocast_bridge_node.hpp#L275-L277)). The canonical example ([repo](https://github.com/ijnek/nested_services_rclcpp_demo)) also follows this pattern: borrow→populate→send within a single callback.

Koichi98: Finally got it. In the basic case the library allocates the response, but in the deferred case the user borrows and publishes themselves — so it's enough to do borrow→publish under the user's own responsibility, within the same callback that calls `send_response`. I had somehow completely misunderstood the API and how it's meant to be used. Anyway, this was a good discussion and I understand it much better now, thanks!

So, as you suggested, could you go ahead and just add the documentation?

bdm-k: Sure 👍

Koichi98 (approved):

Issue #1328, opened 2026-05-13
Enable kmod tests (kunit, e2e, `requires_kernel_module`) in GitHub CI


k1832:
## Background

The repository has substantial automated test coverage for the kernel module —
kunit suites under `agnocast_kmod/agnocast_kunit/`, the
`scripts/test/run_kunit.bash` runner, `scripts/test/run_requires_kernel_module_tests.bash`,
and the `e2e_test_1to1.bash` / `e2e_test_2to2.bash` parameter sweeps — but
**none of them run in GitHub Actions**. The PR template
(`.github/pull_request_template.md`) reflects this: contributors are required to
tick each of those boxes manually.

What CI does today for the kmod:

- `build-kmod.yaml` — runs `make` against the runner's kernel headers (label-gated). Never `insmod`s.
- `sparse-check.yaml`, `checkpatch.yaml` — static-only.
- `build-and-test-agnocastlib-components-heaphook.yaml` — excludes the kernel-module-dependent suites:
  `--ctest-args -LE "requires_kernel_module|requires_heaphook"` (see line 175).

This means a PR can pass CI while breaking, for example,
`get_topic_publisher_info`, `add_subscriber`, `do_exit` cleanup, or the entire
e2e publish path, with the regression surfacing only when a reviewer or the
author remembers to run the scripts locally.

## Why this is worth revisiting now

The historical reason for not running the kmod on hosted runners was that the
module hooked `do_exit` via **kprobes**, which are not enabled in many distro
kernels (and not on GitHub-hosted Ubuntu runners by default). That blocker is
gone:

- PR #1089 replaced the kprobe with `tracepoint_probe_register` on
  `sched_process_exit`.
- The current `agnocast_kmod/` sources include only standard kernel headers
  (`tracepoint.h`, `ipc_namespace.h`, `proc_fs.h`, `rwsem.h`, `hashtable.h`,
  `mm.h`, …). All are unconditionally present in Ubuntu generic kernels.
- Tracepoints are available unconditionally on every supported kernel
  (`CONFIG_TRACEPOINTS=y` is required by huge swaths of the kernel and is
  always on).

So `insmod agnocast.ko` on a stock `ubuntu-22.04` / `ubuntu-24.04` GitHub-hosted
runner is expected to work — these runners are full VMs with root.

## Proposal

Split the work into two phases because they have very different infrastructure
cost.

### Phase 1 — e2e + `requires_kernel_module` + heaphook gtests on stock runners

[...]

PR #1366, opened 2026-05-29
ci(agnocast_kmod): run KUnit suite under QEMU in CI

k1832:
### TL;DR
Adds a `kunit` GitHub Actions workflow that runs the KUnit suite (`agnocast_kmod/agnocast_kunit/`) in CI. Until now, we have been asking the PR author to manually run `scripts/test/run_kunit.bash` locally, and there was no enforcement.

### What's implemented
The job stages the Agnocast sources into a pinned Linux tree (`v6.12` LTS), builds the KUnit suite into the kernel, boots it under QEMU with `kunit.py`, and checks the KTAP results. It runs on PRs that touch `agnocast_kmod/**` (matching the `checkpatch` / `build-kmod` workflows).

This implements the **Phase 2 ("kunit in CI") goal of #1328** — "kunit runs in CI on every PR that touches `agnocast_kmod/`" — as a QEMU variant of that issue's option (C) (`kunit.py` runner). Phase 1 (e2e / `requires_kernel_module` / heaphook gtests via `insmod` on stock runners) and the PR-template update are out of scope here.

### Why kunit.py + QEMU

- **No `insmod`.** `scripts/test/run_kunit.bash` builds a `.ko` and `insmod`s it, which CI runners block via kernel lockdown ([actions/runner-images#8623](https://github.com/actions/runner-images/issues/8623)). `kunit.py` instead links the suite into the kernel and boots it, so no module loading, root, or real hardware is needed.
- **`--arch=x86_64`, not the default UML.** Several tests call `copy_to_user`, which UML can't service from the no-userspace KUnit context (it panics with `Segfault with no mm`); x86_64 under QEMU runs the whole suite.

## Related links

- Addresses Phase 2 of #1328 (enable kmod tests in GitHub CI)
- Why `insmod` is unusable on hosted runners: actions/runner-images#8623

## Files

- `.github/workflows/kunit.yaml` — the CI job.
- `.github/kunit/kunitconfig` — kernel config fragment consumed by `kunit.py`.
- `.github/kunit/Kbuild` — in-tree build file, used only by CI to link the suite into the kernel; reuses `agnocast_kunit/Makefile.inc`.

## Running locally

Prerequisites: `build-essential flex bison libelf-dev libssl-dev bc python3 qemu-system-x86`.

```bash
# Assumes this repo and the kernel tree sit side by side under your home dir:
#   ~/agnocast  (this repo)   ~/linux  (the kernel, cloned in step 1).

# 1. Clone the pinned kernel once.
git clone --depth 1 --branch v6.12 \
  https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git "$HOME/linux"

# 2. Stage the Agnocast sources into the kernel tree (same as the workflow).
dst="$HOME/linux/drivers/agnocast"
mkdir -p "$dst/agnocast_kunit"
cp "$HOME"/agnocast/agnocast_kmod/*.c "$HOME"/agnocast/agnocast_kmod/*.h "$dst/"
cp "$HOME"/agnocast/agnocast_kmod/agnocast_kunit/*.c \
   "$HOME"/agnocast/agnocast_kmod/agnocast_kunit/*.h \
   "$HOME"/agnocast/agnocast_kmod/agnocast_kunit/Makefile.inc "$dst/agnocast_kunit/"
cp "$HOME/agnocast/.github/kunit/Kbuild" "$dst/Makefile"
grep -q 'drivers/agnocast' "$HOME/linux/drivers/Makefile" \
  || echo 'obj-$(CONFIG_KUNIT) += agnocast/' >> "$HOME/linux/drivers/Makefile"

# 3. Build + boot + run.
"$HOME/linux/tools/testing/kunit/kunit.py" run \
  --kunitconfig="$HOME/agnocast/.github/kunit/kunitconfig" \
  --arch=x86_64 --jobs="$(nproc)" 'agnocast*'
```

## CI run time

Measured on `ubuntu-24.04` for the `agnocast-kunit` job:

- **Cold (cache miss — first run on a branch, or after a `LINUX_REF` bump): ~4 min.** Shallow-clones the kernel and does a full kernel build.
- **Warm (cache hit): ~55 s.** Restores the cached kernel source + build dir; only the staged Agnocast objects recompile and the kernel image relinks (~25 s), then it boots and runs (~4 s).

Both the source tree and the build dir are cached, so the steady state on PRs is the ~55 s path.

### From PR template
- [ ] Autoware
- [ ] sample application

### Additional testing
- [x] Reproduced the steps above locally on Linux `v6.12`: all 161 tests pass. Confirmed green on the actual CI run (see CI run time above)
- [x] CI actually fails when there is an issue: `agnocast_ioctl_get_version` was temporarily changed to return an empty version string. The `kunit` job went red as expected ([failed run](https://github.com/autowarefoundation/agnocast/actions/runs/26619423325/job/78441906244)):

## Notes for reviewers

- The build dir is kept outside the kernel source tree so the source cache does not absorb build artifacts, and `restore-keys` lets a warm build dir be reused so only the Agnocast objects recompile.
- The 7-file core object list in `.github/kunit/Kbuild` mirrors the `KUNIT_BUILD=y` branch of `agnocast_kmod/Makefile`. The 26 test objects are shared via `Makefile.inc`, so only this short list needs syncing if a new top-level source file is added.
- `KUNIT_FILTER='agnocast*'` plus a "no tests ran" guard prevents a false green if the suite is renamed (`kunit.py` exits 0 on a zero-match filter).
- **Coverage (gcov/lcov) is intentionally left to a follow-up PR.** This `kunit.py` + QEMU flow has no userspace (the kernel runs the tests during boot and then halts), so it cannot export the gcov data that `run_kunit.bash` reads locally from `/sys/kernel/debug/gcov`. (UML *would* write `.gcda` straight to the build dir, but the suite can't run there.) Coverage needs a separate harness: a GCOV-enabled kernel booted with a tiny initramfs that copies `/sys/kernel/debug/gcov` out over a 9p share, then `lcov` on the host. This was verified end-to-end with a local PoC (≈88% line coverage captured), so it is a concrete follow-up, not an open question.
- An `--arch=arm64` matrix axis could be added later if we need to catch arch-specific regressions.

`need-patch-update` — CI-only change, no API or compatibility impact.


Koichi98:
If the x86_64 QEMU run turns out to be slow, refactoring to separate out `copy_to_user` and switching to UML would be an option. For now, both cold and warm runs look acceptable to me, LGTM.

Koichi98 (approved):

PR #1369, opened 2026-06-01
feat(ros2agnocast): add ros2agnocast verb `ros2 agnocast bridge-daemo…

ruth561:
## Description

Add a new ros2cli verb `ros2 agnocast bridge-daemon-status` that checks the liveness of Agnocast Bridge daemon processes.

### Overview

This change introduces a Unix domain socket (UDS) control socket (in the **abstract namespace**) into the bridge daemon and a corresponding CLI verb to query it.

**agnocastlib (`IpcEventLoopBase`):**
- Each bridge daemon now creates an **abstract namespace** UDS server socket with the name `\0agnocast_bridge/<ipc_ns_inode>/<pid>` during startup.
- When a client connects, the daemon immediately sends a JSON status payload and closes the connection.
- The socket is purely for debug/liveness use. Because the abstract namespace is used, the OS automatically releases the socket when the process terminates — no stale socket files are left behind.

**agnocastlib (bridge managers):**
- Both `PerformanceBridgeManager` and `StandardBridgeManager` register a callback that returns a JSON payload indicating their type, IPC namespace inode, and PID:

```
{"type": "performance"|"standard", "ipc_ns": <int>, "pid": <int>}
```

**ros2agnocast CLI (`ros2 agnocast bridge-daemon-status`):**
- Discovers running bridge daemon processes by scanning `/proc/net/unix` for abstract socket names matching the `agnocast_bridge/<ipc_ns_inode>/` prefix.
- Connects to each socket to verify the daemon is alive and retrieves its type.
- Validates consistency (e.g., no multiple performance bridges, no mixed performance+standard).
- Reports a single-line `OK` / `NG` summary. `--verbose` (`-v`) shows per-bridge details.
- `--timeout SECONDS` controls the per-socket connect timeout (default: 1.0 s).

### Usage Examples

1. When a performance bridge is running:

```
$ ros2 agnocast bridge-daemon-status
OK. A performance bridge is running. (PID: 2127913)
```

With `--verbose`:

```
$ ros2 agnocast bridge-daemon-status --verbose
IPC namespace inode: 4026531839

Summary:
  OK. A performance bridge is running. (PID: 2127913)

Bridge Status:
  PID 2127913 (Performance) OK
```

2. When no bridge daemon is running:

```
$ ros2 agnocast bridge-daemon-status
NG. There is no bridge daemon process.
```

### OK / NG Conditions

The command outputs **OK** or **NG** based on the health of Agnocast Bridge daemon processes found in the current IPC namespace.

**OK cases**

- **A single performance bridge is running** — exactly one performance bridge process is alive and responding correctly via its control socket.
- **One or more standard bridges are running** — one or multiple standard bridge processes are alive and responding correctly.

**NG cases**

| Condition | Message |
|---|---|
| No bridge daemon process exists | `NG. There is no bridge daemon process.` |
| Multiple performance bridge processes exist | `NG. There are multiple performance bridge processes. (PIDs: ...)` |
| Both performance and standard bridge processes coexist | `NG. Both performance and standard bridge processes exist. (PIDs: ...)` |
| A bridge process is alive but does not respond via the control socket | `NG. The/A ... bridge process is not working. (PID: ...)` |
| The response from the bridge cannot be parsed (version mismatch) | `NG. A bridge process is not working. (PID: ...)` — with a message indicating a version incompatibility |
| The response contains an unknown bridge type | Same as above |
| The response contains mismatched PID or IPC namespace data | Same as above |

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Notes for reviewers

The UDS socket uses the **abstract namespace** (`sun_path[0] == '\0'`) instead of filesystem paths so that the socket is automatically released by the OS when the process terminates. This eliminates stale socket files caused by abnormal exits and removes the need for any cleanup logic.

The socket is intentionally write-only from the daemon's perspective (no `recv()`/`read()` on the server side). The act of connecting is the trigger; the daemon responds immediately and closes the connection. This keeps the control path simple and avoids blocking the main `IpcEventLoop`.

Abstract sockets are scoped to the **network namespace**. The CLI discovers bridge processes by scanning `/proc/net/unix` for matching abstract socket names (prefixed with `@agnocast_bridge/<ipc_ns_inode>/`).

`need-patch-update` — adds a new feature (CLI verb + socket infrastructure) without breaking any existing user or internal API.


src/ros2agnocast/ros2agnocast/verb/bridge_daemon_status.py
```diff
@@ -0,0 +1,278 @@
...
+        try:
+            sock.settimeout(timeout_sec)
+            sock.connect(self._sock_path)
+            while True:
+                chunk = sock.recv(256)
+                if not chunk:
+                    break
+                chunks.append(chunk)
+        except Exception:
+            pass
```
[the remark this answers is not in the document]
ruth561: Fixed in [cd46fc4](https://github.com/autowarefoundation/agnocast/pull/1369/commits/cd46fc4dcefe6764001f2b46778680ab756f315a).

src/ros2agnocast/ros2agnocast/verb/bridge_daemon_status.py
```diff
@@ -0,0 +1,278 @@
...
+
+        bridge_pids = _find_bridge_pids(ipc_inode)
+        alive_pids, stale_pids = _partition_pids(bridge_pids)
+
+        results = self._collect_bridge_results(alive_pids, ipc_inode, timeout_sec)
+        stale_paths = [_sock_path_for_pid(ipc_inode, pid) for pid in stale_pids]
+
+        summary, is_ng = _build_summary(results)
+        
+        # output
```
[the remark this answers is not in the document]
ruth561: Fixed in [cd46fc4](https://github.com/autowarefoundation/agnocast/pull/1369/commits/cd46fc4dcefe6764001f2b46778680ab756f315a).

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_ipc_event_loop_base.hpp
```diff
@@ -11,12 +11,16 @@
 #include <mqueue.h>
 #include <sys/epoll.h>
 #include <sys/signalfd.h>
+#include <sys/socket.h>
 #include <sys/stat.h>
+#include <sys/time.h>
+#include <sys/un.h>
```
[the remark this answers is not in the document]
ruth561: Fixed in [cd46fc4](https://github.com/autowarefoundation/agnocast/pull/1369/commits/cd46fc4dcefe6764001f2b46778680ab756f315a).

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_ipc_event_loop_base.hpp
```diff
@@ -177,6 +249,60 @@ inline void IpcEventLoopBase::setup_epoll()
...
+// response. Because no request payload is expected, recv()/read() is
+// intentionally absent from handle_socket().
+inline void IpcEventLoopBase::setup_socket()
+{
+  const char * env = std::getenv("AGNOCAST_TMPFS_DIR");
+  const std::string root_dir =
+    std::string(env != nullptr && *env != '\0' ? env : "/dev/shm") + "/agnocast_bridge_control";
+  if (mkdir(root_dir.c_str(), 0755) == -1 && errno != EEXIST) {
+    throw std::system_error(errno, std::generic_category(), "mkdir failed: " + root_dir);
+  }
```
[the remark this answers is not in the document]
ruth561: You're right. Thank you for pointing that out! Fixed in [cd46fc4](https://github.com/autowarefoundation/agnocast/pull/1369/commits/cd46fc4dcefe6764001f2b46778680ab756f315a).

src/agnocastlib/src/bridge/standard/agnocast_standard_bridge_manager.cpp
```diff
@@ -134,6 +135,12 @@ void StandardBridgeManager::on_signal()
   }
 }
 
+std::string StandardBridgeManager::on_socket_request()
```
Koichi98: `get_self_ipc_ns_inode()` can throw, and it's called on every socket connection here. Since the IPC namespace inode never changes during the process lifetime, how about resolving it once at startup and caching it, then reusing the cached value here? That avoids a per-connection debug path that can throw into the main loop.

src/ros2agnocast/ros2agnocast/verb/bridge_daemon_status.py
```diff
@@ -0,0 +1,279 @@
...
+        try:
+            sock.settimeout(timeout_sec)
+            sock.connect(self._sock_path)
+            while True:
+                chunk = sock.recv(256)
+                if not chunk:
+                    break
+                chunks.append(chunk)
+        except OSError:
+             # Treat any connect/recv error as a liveness failure.
```
kobayu858: There is extra space (before the #).

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_ipc_event_loop_base.hpp
```diff
@@ -177,6 +248,77 @@ inline void IpcEventLoopBase::setup_epoll()
...
+
+  const std::string base_dir = root_dir + "/" + std::to_string(get_self_ipc_ns_inode());
+  if (mkdir(base_dir.c_str(), 0755) == -1 && errno != EEXIST) {
+    RCLCPP_WARN(
+      logger_, "setup_socket: mkdir failed for '%s': %s (socket disabled)", base_dir.c_str(),
+      strerror(errno));
+    return;
+  }
+
+  socket_path_ = base_dir + "/" + std::to_string(getpid()) + ".sock";
```
Koichi98: I'm planning to remove mqueue because it leaves file objects behind, so I'd like to avoid file-backed UDS for the same reason. This one uses the socket file for PID discovery, so if a replacement isn't quick, it's fine to merge as-is and fix it afterwards.

src/ros2agnocast/ros2agnocast/verb/bridge_daemon_status.py
ruth561: Thank you for pointing that out! Fixed in [498dca3](https://github.com/autowarefoundation/agnocast/pull/1369/commits/498dca34caaeade73caa6f77a13ab089ef0843e0).

src/agnocastlib/src/bridge/standard/agnocast_standard_bridge_manager.cpp
ruth561: Thank you for pointing that out! Fixed in [10914b4](https://github.com/autowarefoundation/agnocast/pull/1369/commits/10914b48b8f95353bd22da21ead37366f29c2e3b).

kobayu858 (dismissed):

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_ipc_event_loop_base.hpp
ruth561: I agree with your concern about leaving file objects behind. To address this, I have replaced the file-backed UDS with an abstract namespace UDS. Since this is associated with the process, the system automatically cleans it up even if the process exits abnormally. Could you take a look at the changes in [299b838](https://github.com/autowarefoundation/agnocast/pull/1369/commits/299b83830964318d8181ad927395b3513b18cb13)?

src/ros2agnocast/ros2agnocast/verb/bridge_daemon_status.py
```diff
@@ -0,0 +1,277 @@
...
+def _find_bridge_pids(ipc_inode: int) -> list[int]:
+    """Find all bridge process PIDs via /proc/net/unix.
+
+    Abstract sockets are listed with an '@' prefix in the Path column.
+    Only entries matching our prefix and ipc_inode are considered.
+    Because abstract sockets are released by the OS on process exit,
+    every entry found here belongs to a living process.
+    """
+    target_prefix = f'@{_ABSTRACT_SOCKET_PREFIX}/{ipc_inode}/'
+    pids: list[int] = []
```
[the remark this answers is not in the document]
ruth561: Updated PR description.

src/ros2agnocast/test/test_bridge_daemon_status.py
```diff
@@ -0,0 +1,320 @@
...
+The verb itself talks to /proc and Unix domain sockets, but the helpers are
+exercised here with mocked /proc/net/unix content and in-process abstract
+socket servers in place of the real bridge daemon.
+"""
+
+import os
+import socket
+import threading
+from io import StringIO
+from unittest.mock import mock_open, patch
```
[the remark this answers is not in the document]
ruth561: Fixed in [0264831](https://github.com/autowarefoundation/agnocast/pull/1369/commits/0264831e18c4a3501dd2bc65d62de6097864441a).

src/ros2agnocast/ros2agnocast/verb/bridge_daemon_status.py
```diff
@@ -0,0 +1,277 @@
...
+            default=DEFAULT_TIMEOUT_SEC,
+            metavar='SECONDS',
+            help=f'Timeout in seconds for each ping (default: {DEFAULT_TIMEOUT_SEC})',
+        )
+        parser.add_argument(
+            '--verbose', '-v',
+            action='store_true',
+            default=False,
+            help='Print detailed output including bridge status and stale files',
+        )
```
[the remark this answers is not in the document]
ruth561: Fixed in [0264831](https://github.com/autowarefoundation/agnocast/pull/1369/commits/0264831e18c4a3501dd2bc65d62de6097864441a).

kobayu858 (dismissed):

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_ipc_event_loop_base.hpp
```diff
@@ -129,6 +136,14 @@ inline bool IpcEventLoopBase::spin_once(int timeout_ms)
       if (s == sizeof(struct signalfd_siginfo)) {
         handle_signal();
       }
+    } else if (fd == socket_fd_) {
+      int client_fd = accept4(socket_fd_, nullptr, nullptr, SOCK_CLOEXEC | SOCK_NONBLOCK);
+      if (client_fd == -1) {
```
Koichi98, line 141: nit: We could skip the `EAGAIN/EWOULDBLOCK` warning here. Since we only `accept4()` after the fd is readable, it should almost never happen — but apparently it still can, depending on timing. So treating it as a non-error would be a bit cleaner.

ruth561: Thank you for the suggestion! Since it's a quick fix, I've updated it in this PR: [c0d6025](https://github.com/autowarefoundation/agnocast/pull/1369/commits/c0d60258172a7ba858a70cad6c39ad2c7996749c).

Koichi98 (dismissed):

Koichi98 (approved):

PR #1372, opened 2026-06-03
feat(agnocastlib, ros2agnocast_discovery_agent): auto-generate cross-namespace bridges (standard mode only)

k1832:
## Description

> **Scope: standard bridge mode only.** Performance mode is wired up separately in #1385.

Auto-generate bridges between Agnocast endpoints that live in different IPC namespaces (or ECUs).

Each per-namespace discovery daemon compares its local Agnocast state against the remote gossip snapshots. When a topic has a local endpoint whose opposite-role peer lives in another namespace, the daemon asks its own in-namespace `bridge_manager` to build the matching half of the relay:

- local **publisher** + remote **subscriber** → an **A2R** bridge (Agnocast → DDS)
- local **subscriber** + remote **publisher** → an **R2A** bridge (DDS → Agnocast)

Each namespace builds only its own half; the halves connect over DDS (e.g. NS-A's A2R → DDS → NS-B's R2A), so the two endpoints reach each other.

Changes:

- **agnocastlib**: a new daemon→manager request `MqMsgDaemonBridge` on its own MQ (`create_mq_name_for_daemon_bridge`), which the bridge event loop also listens on (`register_aux_mq`).
  - **What `bridge_manager` already does:** it builds and keeps a bridge for a topic only when it sees a *local* DDS counterpart — i.e. demand-driven from what's visible in its own namespace.
  - **When the agent sends a request to `bridge_manager`:** the daemon sends the request when it sees the counterpart outside its namespace — e.g. NS-A (namespace-A) has an Agnocast subscriber and NS-B an Agnocast publisher. (Without this PR, no bridge will be created in this case.) The `bridge_manager` keeps the created bridge(s) as long as the agent keeps sending the request. Once the agent stops sending the request, the `bridge_manager` drops the bridge(s).
- **ros2agnocast_discovery_agent**: `bridge_decider` computes the needed requests each tick and dispatches `MqMsgDaemonBridge` to the owning process's standard MQ (pid resolved from the tmpfs type registry) and the per-NS performance MQ.

This reuses the existing factory mechanism (offset-based, `#1276`/`#1341`), so no new factory transfer path is added. Additive only — no existing ABI changes.

## Related links

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

In addition:

- **Unit tests**: `bridge_decider` (17 cases — direction/role logic, dedup, type fallback, registry-pid targeting, wire format) and a C++ test pinning the `MqMsgDaemonBridge` wire layout (`sizeof == 524`, field offsets) to the daemon's Python `struct` format.
- **Manual cross-NS test** (standard mode): talker in IPC ns A, listener in IPC ns B, one discovery daemon per namespace.

  ```bash
  # run both from the agnocast workspace root (sudo keeps the cwd; avoid `~`,
  # which expands to /root under sudo)

  # shell A (IPC ns A): publisher
  sudo unshare --ipc bash -c 'source install/setup.bash \
    && export AGNOCAST_BRIDGE_MODE=standard \
    && ros2 launch agnocast_sample_application talker.launch.xml discovery_agent:=true'

  # shell B (IPC ns B): subscriber
  sudo unshare --ipc bash -c 'source install/setup.bash \
    && export AGNOCAST_BRIDGE_MODE=standard \
    && ros2 launch agnocast_sample_application listener.launch.xml discovery_agent:=true'
  ```

  Result: within a few seconds the listener prints `subscribe message: id=N`, i.e. data crossed ns A → DDS → ns B purely via the auto-generated bridges. Confirmed the per-NS `agnocast_daemon_bridge@<pid>` MQ is created, the R2A bridge appears in NS-B's gossip (`is_bridge: true` publisher on `/my_topic`), and `/my_topic` flows on DDS. With `discovery_agent:=false` (no daemon) the listener receives nothing.

- `need-patch-update`

### DIFF
--- a/src/agnocastlib/CMakeLists.txt
+++ b/src/agnocastlib/CMakeLists.txt
@@ -150,7 +150,8 @@ if(BUILD_TESTING)
     test/unit/test_cie_client_utils.cpp
     test/unit/test_agnocast_context.cpp
     test/unit/test_diagnostic_updater.cpp
-    test/unit/test_type_registry_writer.cpp)
+    test/unit/test_type_registry_writer.cpp
+    test/unit/test_daemon_bridge_mq.cpp)
   target_include_directories(test_unit_${PROJECT_NAME} PRIVATE include src)
   target_link_libraries(test_unit_${PROJECT_NAME} agnocast)
   ament_target_dependencies(test_unit_${PROJECT_NAME} std_msgs diagnostic_msgs diagnostic_updater)
--- a/src/agnocastlib/include/agnocast/agnocast_mq.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_mq.hpp
@@ -86,10 +86,33 @@ struct MqMsgPerformanceBridge
   bool is_service;
 };
 
+// Cross-IPC-namespace bridge request from the per-NS daemon to a same-NS
+// bridge_manager. The daemon holds no process-local factory pointers, so it
+// names the target by topic (standard mode reuses the factory the manager
+// already cached from this process's intra-NS requests) and by type
+// (performance mode resolves a plugin). QoS is sent explicitly since the
+// bridge_manager cannot query the originating endpoint's QoS on its own.
+struct MqMsgDaemonBridge
+{
+  char topic_name[TOPIC_NAME_BUFFER_SIZE];
+  char type_name[MESSAGE_TYPE_BUFFER_SIZE];
+  BridgeDirection direction;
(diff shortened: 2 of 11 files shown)


k1832:
Superseded by #1389. Restructured the stack to drop standard-mode daemon support and lead with performance mode (per the GenericSubscription/GenericPublisher design that unifies the bridges), on top of the request→registration rename (#1388). This PR's shared infra + discovery agent moved to #1389 (performance-only); the performance manager is in #1390.

PR #1383, opened 2026-06-09
feat(bridge): add performance-mode A2R service bridging

bdm-k:
## Description

This PR implements A2R service bridging and extracts the service bridge lifecycle into a self-contained `ServiceBridgeItem` class with a state-machine.

### Overview by file

Thanks for reviewing this huge diff. Below is a per-file overview with the significance of each change to help guide your review.

| File | Changes | Significance |
| ---- | ------- | ------------ |
| `agnocast_service_bridge` | Added `ServiceBridgeItem`. | **High** |
| `service_bridge_plugin.cpp.em` | (1) Added `create_a2r_service_bridge()` template; (2) Updated `create_r2a_service_bridge()` template to use an Agnocast client with no bridging policy. | **High** |
| `agnocast_performance_bridge_manager` | Wired up `ServiceBridgeItem`. | **High** |
| `agnocast_performance_bridge_loader` | Added `create_a2r_service_bridge()`. | Low |
| `agnocast_performance_bridge_plugin_api.hpp` | (1) Renamed `PerformanceServiceBridgeResult` to `ServiceBridgeEntity` because this will also be used in standard bridges in due time; (2) Added C interface of `create_a2r_service_bridge()`. | Low |
| `agnocast_client.hpp` | Changed `Client` to `BasicClient` for encapsulating bridge policies. | Low |
| `agnocast_service.hpp` | Made the constructors directly call `request_bridge()` to satisfy compiler the compiler.  | Low |
| `agnocast_bridge_node.hpp` | (1) Added `AgnocastToRosServiceRequestPolicy`; (2) Updated `RosToAgnocastServiceBridge` to use an Agnocast client with no bridging policy; (3) Fixed `RosToAgnocastServiceRequestPolicy::request_bridge`'s template parameter order to satisfy the compiler. | Low |
| `agnocast_bridge_utils` | Added `split_full_node_name()`. | Low |

### Implementation details

The most significant change is in `ServiceBridgeItem`. An instance of this class represents a service bridge for a specific service name. Please refer to the design document for the underlying state machine's transition rules; this will help you review the code. While the class has many member functions, they can be classified into three categories based on their code path.

1.  Functions used when a bridging request comes in. The entry point is `handle_request()`, with `update_configuration()` and `handle_request_with_direction()` as sub-functions.
2. Functions used when periodically probing the system state. The entry point is `check_and_update()`, with a sub-function for each state except NONE.
3. Utility functions that may be called from both code paths. Only `start_r2a_bridge()` and `start_a2r_bridge()` modify the data members; all others are const member functions.

The performance bridge manager wires `ServiceBridgeItem` up. It manages a set of `ServiceBridgeItem`s using an `unordered_map`. The key point here is that every managed `ServiceBridgeItem` is in a state other than NONE.

The A2R service bridge template is added in `service_bridge_plugin.cpp.em`. It's exactly the opposite of the R2A service bridge; it uses a pair of ROS 2 client and Agnocast service. The deferred service callbacks that we recently implemented are used here.

## Related links

- Design document ([TIER IV internal link](https://tier4.atlassian.net/wiki/x/FYH0IQE))

## How was this PR tested?

- [ ] Autoware
- [x] sample application
- [x] Exercise every state transition path to confirm the functionality

## Notes for reviewers

While I'm aware that a large refactoring for bridges is underway, I don't yet have the full picture. I've defined `ServiceBridgeItem` in a separate file, but if that is suboptimal, please let me know and we can discuss it.


src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
```diff
@@ -0,0 +1,428 @@
...
+#include <cerrno>
+#include <map>
+#include <utility>
+
+namespace
+{
+
+// The global error string used by ServiceBridgeItem.
+std::string error_string;
+
```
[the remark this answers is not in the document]
bdm-k: The bridge manager operates in a single thread.

src/ros2agnocast/ros2agnocast/templates/service_bridge_plugin.cpp.em
```diff
@@ -51,3 +48,39 @@ extern "C" PerformanceServiceBridgeResult create_r2a_service_bridge_@(snake_type
...
+{
+  using ServiceT = @(cpp_type);
+  using AgnoService = agnocast::BasicService<ServiceT, agnocast::NoBridgeRequestPolicy>;
+
+  auto srv_cb_group = node->create_callback_group(rclcpp::CallbackGroupType::Reentrant);
+  auto client_cb_group = node->create_callback_group(rclcpp::CallbackGroupType::Reentrant);
+
+  auto ros_client =
+    node->create_client<ServiceT>(service_name, qos.get_rmw_qos_profile(), client_cb_group);
+
```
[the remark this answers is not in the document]
bdm-k: fixed in [a5cd768](https://github.com/autowarefoundation/agnocast/pull/1383/commits/a5cd768487c3d268d9f9f4a3f4babf73da675384)

src/agnocastlib/include/agnocast/agnocast_client.hpp
```diff
@@ -154,14 +152,14 @@ class Client
   }
```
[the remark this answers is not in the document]
bdm-k: fixed in [b051a48](https://github.com/autowarefoundation/agnocast/pull/1383/commits/b051a48fb6e11ebc920f14b5b24e0ac7c3ad5570)

src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
```diff
@@ -0,0 +1,428 @@
...
+std::shared_ptr<rcl_node_t> ServiceBridgeItem::find_or_create_shadow_node(
+  const std::pair<std::string, std::string> & identity)
+{
+  auto it = g_shadow_nodes.find(identity);
+  if (it != g_shadow_nodes.end()) {
+    auto shadow_node = it->second.lock();
+    if (shadow_node != nullptr) {
+      return shadow_node;
+    }
+  }
```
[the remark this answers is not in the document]
bdm-k: fixed in [9486a37](https://github.com/autowarefoundation/agnocast/pull/1383/commits/9486a37187194169c8b6023e6a9b5b820b70ac4d)

src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
```diff
@@ -0,0 +1,413 @@
...
+  if (shadow_node_identity_.has_value()) {
+    const auto & identity = *shadow_node_identity_;
+    if ((shadow_node = find_or_create_shadow_node(identity)) == nullptr) {
+      return -1;
+    }
+  }
+
+  ServiceBridgeEntity entity;
+  if (service_type_.has_value() && ctx.performance_loader != nullptr) {
+    entity = ctx.performance_loader->create_r2a_service_bridge(
```
Koichi98: Can we add try/catch here as we do for pub/sub?
https://github.com/autowarefoundation/agnocast/pull/1383/changes#diff-a1cd490d4e2387b4c006621cf77bc5e5266ad0ec44e903cdc61d8ac6425404d1L445

bdm-k: fixed in [66a80ad](https://github.com/autowarefoundation/agnocast/pull/1383/commits/66a80ad8e47818b9dad6e54e18a705558ee27acc)

src/agnocastlib/include/agnocast/bridge/agnocast_service_bridge.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+#include <memory>
+#include <optional>
+#include <string>
+#include <utility>
+
+namespace agnocast
+{
+
+struct BridgeManagerContext
```
Koichi98: Might be better to rename this to `ServiceBridgeDeps` or something that describe this is just a plain bundle of injected dependencies handed to `ServiceBridgeItem`.

src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
```diff
@@ -0,0 +1,429 @@
...
+      return shadow_node;
+    }
+  }
+
+  rcl_context_t * rcl_ctx = rclcpp::contexts::get_global_default_context()->get_rcl_context().get();
+
+  rcl_node_options_t options = rcl_node_get_default_options();
+  options.enable_rosout = false;
+  options.use_global_arguments = false;
+  if (rcl_parse_arguments(0, nullptr, options.allocator, &(options.arguments)) != RCL_RET_OK) {
```
Koichi98, line 59: Seems like we need to call `rcl_node_options_fini()`.

src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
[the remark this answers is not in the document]
Koichi98: > The bridge manager operates in a single thread.

This is true, but I'd like to avoid process-global mutable state as a general principle. 
Concretely here, the shared `error_string` introduces an implicit set→get coupling, which makes the code harder to maintain and reason about. I'd prefer we scope it to the instance — e.g. make it a member of ServiceBridgeItem — rather than keeping a true global.

src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
```diff
@@ -0,0 +1,429 @@
...
+    return -1;
+  }
+
+  if (entity.entity_handle == nullptr) {
+    set_error_string("Bridge loader returned nullptr");
+    return -1;
+  }
+
+  state_ = ServiceBridgeState::A2R;
+  entity_ = entity;
```
Koichi98: nit: `std::move` here?

src/agnocastlib/src/bridge/performance/agnocast_performance_bridge_manager.cpp
```diff
@@ -130,7 +130,20 @@ void PerformanceBridgeManager::on_bridge_message(const void * data, std::size_t
         return;
       }
       const auto & payload = msg.payload.service;
-      create_service_bridge_if_needed(payload, payload.direction);
+      BridgeManagerContext ctx{container_node_, executor_, logger_, loader_};
+      std::string service_name = static_cast<const char *>(payload.service_name);
```
Koichi98, line 134: nit:  I'd lean toward reading these fixed-size buffers with a bounded `strnlen` instead of relying on null-termination. No need to change it in this PR since the same pattern exists throughout the codebase, but worth doing codebase-wide as a follow-up.

src/agnocastlib/src/bridge/performance/agnocast_performance_bridge_manager.cpp
```diff
@@ -357,28 +370,19 @@ void PerformanceBridgeManager::check_and_remove_pubsub_bridges()
...
 
-void PerformanceBridgeManager::check_and_remove_service_bridges()
+void PerformanceBridgeManager::check_and_update_service_bridges()
 {
-  auto r2a_srv_it = active_r2a_service_bridges_.begin();
-  while (r2a_srv_it != active_r2a_service_bridges_.end()) {
-    const std::string & service_name = r2a_srv_it->first;
+  auto it = active_service_bridges_.begin();
+  while (it != active_service_bridges_.end()) {
+    BridgeManagerContext ctx{container_node_, executor_, logger_, loader_};
```
Koichi98: Can we place this outside of the `while` loop?

src/agnocastlib/src/bridge/performance/agnocast_performance_bridge_manager.cpp
bdm-k: Noted

src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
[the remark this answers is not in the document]
bdm-k: fixed in [61a0c38](https://github.com/autowarefoundation/agnocast/pull/1383/commits/61a0c38ba61102911ef46351b4b1c70c2e9c5ebf)

I was actually not sure about the previous design. Thanks for your advice.

src/agnocastlib/include/agnocast/bridge/agnocast_service_bridge.hpp
bdm-k: changed to `ServiceBridgeDeps` in [d24477b](https://github.com/autowarefoundation/agnocast/pull/1383/commits/d24477bd7f84f00e9fb2e3d07b949425ab224356)

src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
bdm-k: fixed in [21b1652](https://github.com/autowarefoundation/agnocast/pull/1383/commits/21b165221c8f01cad6664a81d1ba238555f7a82a)

src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
bdm-k: fixed in [3ac6fe3](https://github.com/autowarefoundation/agnocast/pull/1383/commits/3ac6fe31821705efcddfee1d82d79656938e7e36)

src/agnocastlib/src/bridge/performance/agnocast_performance_bridge_manager.cpp
bdm-k: fixed in [3c67049](https://github.com/autowarefoundation/agnocast/pull/1383/commits/3c6704926e28f38a798201bad5203045ac2212ad)

Koichi98 (dismissed):

Koichi98 (approved):

PR #1385, opened 2026-06-10
feat(agnocastlib): handle daemon cross-NS bridge requests in performance mode

k1832:
## Description

Auto-generate cross-IPC-namespace (and cross-ECU) bridges under **performance mode** too. The per-namespace discovery daemon already dispatches `MqMsgDaemonBridge` to the per-NS performance MQ (#1372); this PR makes the performance `bridge_manager` act on it.

Changes:

- **agnocastlib**: the performance `bridge_manager` listens on the per-NS daemon MQ (`/agnocast_daemon_bridge_perf`, one per IPC namespace — not keyed by pid like standard mode). While a request is unexpired it creates the bridge **without** a same-graph DDS counterpart and shields it from demand-based teardown; that counterpart cannot exist until the peer namespace's bridge is also up (chicken-and-egg). The force expires shortly after the daemon stops asserting, restoring the normal on-demand lifecycle.
- Unlike standard mode — which reuses the process-local factory cached from its own intra-NS requests — performance mode has no such factory, so the bridge is resolved from the request's **message type** (plugin) and its QoS is rebuilt from the request's explicit fields (`daemon_request_qos`). This is the asymmetry already documented on `MqMsgDaemonBridge`.
- `DAEMON_FORCE_TTL` moves into the shared `bridge_utils` header so both managers share one source of truth for the force-lease duration (it must stay longer than the daemon's publish interval).

No agent-side changes — the decider already targets the per-NS performance MQ (#1372). Additive only; no existing ABI changes.

## Related links

- Standard-mode counterpart: #1372

## How was this PR tested?

- [ ] Autoware
- [x] `bash scripts/test/e2e_test_1to1.bash` (required) — standard mode
- [x] `bash scripts/test/e2e_test_2to2.bash` (required) — standard mode
- [ ] sample application

In addition:

- **Unit tests** (`test_unit_agnocastlib`, `DaemonBridgeMqTest`): the existing `MqMsgDaemonBridge` wire-layout / MQ-naming cases, plus two new cases pinning `daemon_request_qos` — the performance-mode QoS rebuilt from the request's explicit depth / reliability / durability fields.
- **Manual cross-NS test** (performance mode) — **confirmed PASS**: listener in IPC ns B received the talker (IPC ns A) via the auto-generated performance bridge. Performance mode needs a one-time setup step first — it does not work by default (see [#1386](https://github.com/autowarefoundation/agnocast/issues/1386)): generate the bridge plugin for the topic's type, then run talker/listener in separate IPC namespaces:

  ```bash
  # source this workspace first; local_setup.bash avoids pulling in other overlays (e.g. autoware)
  source /opt/ros/humble/setup.bash && source install/local_setup.bash
  ros2 agnocast generate-bridge-plugins --message-types agnocast_sample_interfaces/msg/DynamicSizeArray
  colcon build --packages-select agnocast_bridge_plugins && source install/local_setup.bash

  # talker (ns A) and listener (ns B), each in its own shell:
  sudo unshare --ipc bash -c 'sysctl -w fs.mqueue.msg_max=1024 >/dev/null; source install/setup.bash; \
    AGNOCAST_BRIDGE_MODE=performance ros2 launch agnocast_sample_application talker.launch.xml discovery_agent:=true'
  sudo unshare --ipc bash -c 'sysctl -w fs.mqueue.msg_max=1024 >/dev/null; source install/setup.bash; \
    AGNOCAST_BRIDGE_MODE=performance ros2 launch agnocast_sample_application listener.launch.xml discovery_agent:=true'
  ```

  Expected: within a few seconds the listener prints `subscribe message: id=N` (with `discovery_agent:=false`, nothing).

## Notes for reviewers

- Stacked on #1372 — review/merge that first. The diff here is the performance-manager half plus the shared `DAEMON_FORCE_TTL` relocation.
- The force-lease / teardown-shield behavior mirrors standard mode and is exercised by the manual cross-NS test above; the unit tests cover the one piece of new pure logic (QoS reconstruction), matching the test depth of #1372.
- The kmod is untouched: type names still come from the tmpfs type registry (#1356); moving that to a kmod ioctl is deferred to the planned refactor and is independent of this change.

- `need-patch-update`

### DIFF
--- a/src/agnocastlib/include/agnocast/bridge/agnocast_bridge_utils.hpp
+++ b/src/agnocastlib/include/agnocast/bridge/agnocast_bridge_utils.hpp
@@ -5,6 +5,7 @@
 
 #include <rclcpp/rclcpp.hpp>
 
+#include <chrono>
 #include <cstring>
 #include <memory>
 #include <optional>
@@ -24,6 +25,12 @@ inline constexpr std::string_view SUFFIX_SERVICE_A2R = "_S_A2R";
 
 inline constexpr size_t SUFFIX_LEN = 6;
 static_assert(SUFFIX_PUBSUB_R2A.length() == SUFFIX_LEN);
+
+// How long a daemon cross-NS bridge request keeps a bridge forced after the last
+// request. Must exceed the daemon's publish interval (1 s) so a few missed ticks
+// don't tear down a healthy cross-NS bridge, yet stay short so a vanished remote
+// endpoint stops being forced promptly. Shared by both bridge manager modes.
+inline constexpr std::chrono::seconds DAEMON_FORCE_TTL{5};
 static_assert(SUFFIX_PUBSUB_A2R.length() == SUFFIX_LEN);
 static_assert(SUFFIX_SERVICE_R2A.length() == SUFFIX_LEN);
 static_assert(SUFFIX_SERVICE_A2R.length() == SUFFIX_LEN);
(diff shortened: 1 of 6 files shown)


k1832:
Superseded by #1390 (rebased onto the restructured stack #1388 → #1389 → #1390). Content is unchanged apart from dropping the now-removed standard-manager hunk.

Issue #1386, opened 2026-06-10
Performance bridge mode: e2e tests are not runnable out of the box (missing plugin generation + mqueue msg_max)


k1832:
## Summary

Running the e2e suites in **performance** bridge mode (`AGNOCAST_BRIDGE_MODE=performance bash scripts/test/e2e_test_1to1.bash`) fails on every bridge case unless two (three under parallel runs) manual setup steps are performed first. Neither is done by `build_all.bash` or the e2e scripts, and they are not documented, so performance-mode bridges effectively have **no runnable e2e coverage**.

Standard mode (the default) works out of the box; only performance mode needs the extra setup.

## Background

- `e2e_test_1to1` covers the bridge path via `test_ros2_sub` — a plain ROS 2 / DDS subscriber that can only receive through an A2R bridge.
- The suite runs once in whatever `AGNOCAST_BRIDGE_MODE` is set (default `standard`). So performance-mode bridges are only exercised by running it with `AGNOCAST_BRIDGE_MODE=performance`.

## Required (but un-automated) setup & root causes

1. **Bridge plugins are not generated.** Performance-mode bridges `dlopen` a per-type `libagnocast_bridge_plugins.so`, produced by `ros2 agnocast generate-bridge-plugins`. The e2e flow never generates it, so the loader finds nothing → no bridge → the bridged subscriber receives 0.
   - Workaround: `ros2 agnocast generate-bridge-plugins --message-types <type>` (e.g. `std_msgs/msg/Int64` for `e2e_test_1to1`, `agnocast_sample_interfaces/msg/DynamicSizeArray` for the sample app), then `colcon build --packages-select agnocast_bridge_plugins`.

2. **`fs.mqueue.msg_max` (default 10) is too small.** The performance manager is one-per-IPC-namespace and its request MQ uses `PERFORMANCE_BRIDGE_MQ_MAX_MESSAGES = 256` (vs standard's per-process `BRIDGE_MQ_MAX_MESSAGES = 2`). `mq_open(O_CREAT, mq_maxmsg=256)` fails with `EINVAL` when `fs.mqueue.msg_max < 256`, so the bridge request is never delivered:
   ```
   mq_open failed for name '/agnocast_bridge_manager@-1': Invalid argument (errno: 22)
   ```
   - Workaround: `sudo sysctl -w fs.mqueue.msg_max=1024`.

[...]

PR #1390, opened 2026-06-11
feat(agnocastlib): handle daemon cross-NS bridge requests in performance mode

k1832:
## Description

Make the `bridge_manager` (one per IPC namespace) act on the cross-NS bridge requests dispatched by the discovery agent. With this, cross-NS (and cross-ECU) bridges are auto-generated.

Changes:

- **agnocastlib**: the `bridge_manager` now also listens on its per-NS daemon MQ (`/agnocast_daemon_bridge_perf`) for `MqMsgDaemonBridge` requests.
  - **What `bridge_manager` already does:** it builds and keeps a bridge for a topic only when it sees a *local* DDS counterpart — i.e. demand-driven from what's visible in its own namespace.
  - **When the agent sends a request to `bridge_manager`:** the agent does so when it sees the counterpart outside its namespace — e.g. NS-A (namespace-A) has an Agnocast subscriber and NS-B an Agnocast publisher. (Without this PR, no bridge will be created in this case.) The local counterpart can't exist until the peer namespace's bridge is up too (chicken-and-egg), so while a request keeps coming the `bridge_manager` creates and keeps the bridge *without* one. It keeps the bridge as long as the agent keeps sending the request, and drops it shortly after the agent stops (after `DAEMON_FORCE_TTL`, which must outlast the agent's publish interval).
  - The `bridge_manager` has no process-local factory, so it resolves the bridge from the request's **message type** (plugin) and rebuilds its QoS from the request's explicit fields (`daemon_request_qos`).
  - Sizes the daemon request MQ (`DAEMON_BRIDGE_MQ_MAX_MESSAGES = 256`, matching the performance MQ) so a startup burst of requests is absorbed in one tick instead of trickling in at the queue depth per second.

Additive only; no existing ABI changes, kmod untouched.

## Related links

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

In addition:

- **Unit tests** (`agnocastlib` gtest, incl. the `MqMsgDaemonBridge` wire-layout cases plus two new cases pinning `daemon_request_qos` — the QoS rebuilt from the request's explicit depth / reliability / durability fields). All pass.
- **Manual cross-NS test** — **confirmed PASS**: listener in IPC ns B received the talker (IPC ns A) via the auto-generated bridge. The bridge needs a one-time setup step first — it does not work by default (see [#1386](https://github.com/autowarefoundation/agnocast/issues/1386)): generate the bridge plugin for the topic's type, then run talker/listener in separate IPC namespaces:

  ```bash
  # source this workspace first; local_setup.bash avoids pulling in other overlays (e.g. autoware)
  source /opt/ros/humble/setup.bash && source install/local_setup.bash
  ros2 agnocast generate-bridge-plugins --message-types agnocast_sample_interfaces/msg/DynamicSizeArray
  colcon build --packages-select agnocast_bridge_plugins && source install/local_setup.bash

  # talker (ns A) and listener (ns B), each in its own shell:
  sudo unshare --ipc bash -c 'sysctl -w fs.mqueue.msg_max=1024 >/dev/null; source install/setup.bash; \
    ros2 launch agnocast_sample_application talker.launch.xml discovery_agent:=true'
  sudo unshare --ipc bash -c 'sysctl -w fs.mqueue.msg_max=1024 >/dev/null; source install/setup.bash; \
    ros2 launch agnocast_sample_application listener.launch.xml discovery_agent:=true'
  ```

  Expected: within a few seconds the listener prints `subscribe message: id=N` (with `discovery_agent:=false`, nothing).

- `need-patch-update`

### DIFF
--- a/src/agnocastlib/src/bridge/performance/agnocast_performance_bridge_manager.cpp
+++ b/src/agnocastlib/src/bridge/performance/agnocast_performance_bridge_manager.cpp
@@ -58,6 +58,10 @@ void PerformanceBridgeManager::run()
   event_loop_.set_mq_handler([this](int fd) { this->on_mq_request(fd); });
   event_loop_.set_signal_handler([this]() { this->on_signal(); });
   event_loop_.set_socket_handler([this]() { return this->on_socket_request(); });
+  // One bridge manager runs per IPC namespace, so its daemon request MQ is per-NS too.
+  event_loop_.register_aux_mq(
+    create_mq_name_for_daemon_bridge(PERFORMANCE_BRIDGE_VIRTUAL_PID), DAEMON_BRIDGE_MQ_MAX_MESSAGES,
+    DAEMON_BRIDGE_MQ_MESSAGE_SIZE, [this](int fd) { this->on_daemon_mq_request(fd); });
 
   while (!shutdown_requested_) {
     if (!event_loop_.spin_once(EVENT_LOOP_TIMEOUT_MS)) {
@@ -66,6 +70,7 @@ void PerformanceBridgeManager::run()
     }
 
     check_and_create_pubsub_bridges();
+    create_daemon_forced_bridges();
     check_and_remove_pubsub_bridges();
     check_and_remove_service_bridges();
     check_and_remove_request_cache();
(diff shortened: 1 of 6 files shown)


src/agnocastlib/src/bridge/performance/agnocast_performance_bridge_manager.cpp
```diff
@@ -125,6 +131,100 @@ void PerformanceBridgeManager::on_mq_request(int fd)
...
+void PerformanceBridgeManager::on_daemon_mq_request(int fd)
+{
+  MqMsgDaemonBridge req{};
+  while (mq_receive(fd, reinterpret_cast<char *>(&req), sizeof(req), nullptr) > 0) {
+    if (shutdown_requested_) {
+      break;
+    }
+    register_daemon_pubsub_request(req);
+  }
+}
```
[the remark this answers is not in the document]
k1832: Fixed — `on_daemon_mq_request` now checks the `mq_receive` return value like `on_mq_request`: it breaks on EAGAIN (queue drained) and logs unexpected errno values instead of silently dropping daemon bridge requests.

src/agnocastlib/src/bridge/performance/agnocast_performance_bridge_manager.cpp
```diff
@@ -58,6 +58,11 @@ void PerformanceBridgeManager::run()
   event_loop_.set_mq_handler([this](int fd) { this->on_mq_request(fd); });
   event_loop_.set_signal_handler([this]() { this->on_signal(); });
   event_loop_.set_socket_handler([this]() { return this->on_socket_request(); });
+  // Performance mode runs one bridge manager per IPC namespace, so the daemon
+  // request MQ is per-NS too (not keyed by pid like standard mode).
+  event_loop_.register_aux_mq(
+    create_mq_name_for_daemon_bridge(PERFORMANCE_BRIDGE_VIRTUAL_PID), DAEMON_BRIDGE_MQ_MAX_MESSAGES,
+    DAEMON_BRIDGE_MQ_MESSAGE_SIZE, [this](int fd) { this->on_daemon_mq_request(fd); });
```
Koichi98, line 64: Seems like `DAEMON_BRIDGE_MQ_MESSAGE_SIZE` is two. Isn't it a bit too small?

k1832: (Assuming you meant `DAEMON_BRIDGE_MQ_MAX_MESSAGES`.) yes, 2 was too small.... I bumped it up to 256 to match with `PERFORMANCE_BRIDGE_MQ_MAX_MESSAGES`

Koichi98 (approved):

PR #1393, opened 2026-06-12
feat(agnocastlib): add `agnocast::GenericSubscription`

ruth561:
## Description

Add `agnocast::GenericSubscription`, which mirrors `rclcpp::GenericSubscription`.

The topic type is given as a runtime string (e.g. `"std_msgs/msg/String"`) instead of a compile-time template argument. Messages are delivered to the user callback as `rclcpp::SerializedMessage` (i.e. serialized bytes), not as the message type itself.

Only the Performance bridge mode is supported; under the Standard bridge mode no bridge request is issued and a warning is logged once.

## How it works

The design tries to reuse the existing subscription machinery (`CallbackInfo`, `register_*_callback`, `receive_and_execute_message`) as much as possible. Only the message envelope and the dispatch lambda are new; the executor / epoll / mq path is unchanged.

Key points:

- **Type-erased envelope.** `CallbackInfo` already dispatches messages through an `AnyObject` envelope (`TypedMessagePtr<T>` for `BasicSubscription`). We add a sibling `RawMessagePtr` that wraps `ipc_shared_ptr<std::byte>`, so the same reference-counting and kernel-release path is reused even though the message type is unknown at compile time.
- **Dispatch.** On each message, the type-erased callback returned by `get_erased_generic_callback` serializes the data into a `rclcpp::SerializedMessage` using `rmw_serialize`. It then releases the SHM reference and invokes the user callback.
- **Registration.** The existing `register_callback` is split into a small `register_erased_callback` core. Both `register_callback` (typed) and the new `register_generic_callback` go through it.

## How was this PR tested?

- [ ] Autoware
- [x] sample application

Manual tests using the sample application were run as follows.

**Terminal 1 (listener, common to both cases):**

```bash
export AGNOCAST_BRIDGE_MODE=performance
bash scripts/sample_application/run_generic_listener.bash
```

**Terminal 2 (Case 1: A2A):**

```bash
export AGNOCAST_BRIDGE_MODE=performance
bash scripts/sample_application/run_talker.bash
```

**Terminal 2 (Case 1: R2A):**

```bash
ros2 topic pub /my_topic agnocast_sample_interfaces/msg/DynamicSizeArray "{ id: 1729, data: [] }"
```

## Notes for reviewers

A few things were intentionally left out of this PR to keep it small and
keep reviewer load low. They are written down here so the discussion is not
repeated in line comments.

- **Code duplication with `BasicSubscription` is left as-is on purpose.**
  `GenericSubscription` and `BasicSubscription` share quite a lot of logic
  (constructor body, QoS override handling, ioctl call, mq setup,
  destructor cleanup). They could be unified, but doing it in this PR
  would make the diff much harder to review. The unification is planned as
  a separate follow-up PR.

- **Different bridge-request design from `BasicSubscription` — please do
  not worry about it in this PR.**
  `BasicSubscription` takes a template type `BridgeRequestPolicy`, but
  `GenericSubscription` adopts a design that does not use
  `BridgeRequestPolicy`. Instead, it introduces a small
  `SubscriptionRole` enum (`Default` / `AgnocastOnly` / `BridgeInternal`)
  to express the same intent without extra templates.
  `BasicSubscription` is intended to adopt this same design in the
  follow-up refactor PR mentioned above, so the two will end up
  consistent. The design difference itself is therefore out of scope for
  this PR.

`need-patch-update`

### DIFF
--- a/src/agnocastlib/include/agnocast/agnocast_callback_info.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_callback_info.hpp
@@ -111,7 +111,7 @@ TypeErasedCallback get_erased_callback(Func && callback)
   };
 }
 
-// Serialize directly into an existing stack-allocated SerializedMessage to avoid heap allocation.
+// Serialize raw message data into a SerializedMessage.
 // Returns false (and logs an error) on rmw_serialize failure.
 bool generic_serialize_message_in_place(
   const void * raw, const rosidl_message_type_support_t * type_support,
@@ -126,7 +126,7 @@ TypeErasedCallback get_erased_generic_callback(
     std::is_invocable_v<F, std::shared_ptr<rclcpp::SerializedMessage>> ||
       std::is_invocable_v<F, std::unique_ptr<rclcpp::SerializedMessage>> ||
       std::is_invocable_v<F, rclcpp::SerializedMessage &>,
-    "This callback type cannot be handled as a GenericCallback."
+    "This callback type cannot be handled as a GenericCallback. "
     "Callback must be invocable with one of the following arguments "
     "(or any types implicitly convertible from them, e.g., const variants): "
     "std::unique_ptr<rclcpp::SerializedMessage>, "
@@ -161,7 +161,6 @@ TypeErasedCallback get_erased_generic_callback(
       }
       callback(std::move(serialized));
     } else {
-      // Use a stack-allocated SerializedMessage to avoid heap allocation.
       rclcpp::SerializedMessage serialized;
       if (!generic_serialize_message_in_place(raw_ptr.get(), type_support, serialized)) {
         return;
--- a/src/agnocastlib/include/agnocast/agnocast_callback_info.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_callback_info.hpp
@@ -4,7 +4,11 @@
 #include "agnocast/agnocast_epoll_update_dispatcher.hpp"
 #include "agnocast/agnocast_smart_pointer.hpp"
 
+#include <rclcpp/serialized_message.hpp>
+
+#include <cstddef>
 #include <cstdint>
+#include <memory>
 #include <mutex>
 #include <type_traits>
 
@@ -40,8 +44,33 @@ public:
   agnocast::ipc_shared_ptr<T> && get() && { return std::move(ptr_); }
 };
 
+// Class for a generic message type used by GenericSubscription.
+class RawMessagePtr : public AnyObject
+{
+  // We use std::byte for ipc_shared_ptr because the exact message type is unknown
+  // at compile time in GenericSubscription.
+  //
+  // This is safe because this pointer purely references a subscribed message.
+  // It does not handle memory allocation or deallocation (no delete is called);
+  // it strictly manages the reference count and releases the reference to the
+  // kernel module. (See ipc_shared_ptr implementation for details).
+  agnocast::ipc_shared_ptr<std::byte> ptr_;
+
+public:
+  explicit RawMessagePtr(agnocast::ipc_shared_ptr<std::byte> p) : ptr_(std::move(p)) {}
+
+  const std::type_info & type() const override { return typeid(RawMessagePtr); }
+
+  agnocast::ipc_shared_ptr<std::byte> && get() && { return std::move(ptr_); }
+};
+
 // Type for type-erased callback function
 using TypeErasedCallback = std::function<void(AnyObject &&)>;
+
+// Type for message creator function that constructs a type-erased message
+// envelope (AnyObject) from a raw shared-memory pointer and its metadata.
+using MessageCreator = std::function<std::unique_ptr<AnyObject>(
+  const void *, const std::string &, const topic_local_id_t, const int64_t)>;
 
 struct CallbackInfo
 {
@@ -51,9 +80,7 @@ struct CallbackInfo
   mqd_t mqdes;
   rclcpp::CallbackGroup::SharedPtr callback_group;
   TypeErasedCallback callback;
-  std::function<std::unique_ptr<AnyObject>(
-    const void *, const std::string &, const topic_local_id_t, const uint64_t)>
-    message_creator;
+  MessageCreator message_creator;
   bool need_epoll_update = true;
 };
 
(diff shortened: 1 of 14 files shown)


src/agnocastlib/test/unit/test_agnocast_generic_callback.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+#include <gtest/gtest.h>
+
+#include <memory>
+#include <string>
+
+using agnocast::get_erased_generic_callback;
+using agnocast::RawMessagePtr;
+using agnocast::TypedMessagePtr;
+using agnocast::TypeErasedCallback;
```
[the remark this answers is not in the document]
ruth561: Fixed in [14ec6ef](https://github.com/autowarefoundation/agnocast/pull/1393/commits/14ec6ef98e8a19cf79db3ec0ab5dabd68b73796a).

src/agnocastlib/include/agnocast/agnocast_callback_info.hpp
```diff
@@ -4,7 +4,11 @@
 #include "agnocast/agnocast_epoll_update_dispatcher.hpp"
 #include "agnocast/agnocast_smart_pointer.hpp"
 
+#include <rclcpp/serialized_message.hpp>
+
+#include <cstddef>
 #include <cstdint>
+#include <memory>
```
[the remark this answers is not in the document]
ruth561: Fixed in [14ec6ef](https://github.com/autowarefoundation/agnocast/pull/1393/commits/14ec6ef98e8a19cf79db3ec0ab5dabd68b73796a).

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
```diff
@@ -419,6 +450,108 @@ class BasicPollingSubscriber
...
+    agnocast::SubscriptionOptions options = agnocast::SubscriptionOptions(),
+    SubscriptionRole role = SubscriptionRole::Default)
+  : SubscriptionBase(node, topic_name)
+  {
+    rclcpp::CallbackGroup::SharedPtr callback_group = get_valid_callback_group(node, options);
+
+    const void * callback_addr = static_cast<const void *>(&callback);
+    const char * callback_symbol = tracetools::get_symbol(callback);
+
+    load_typesupport_impl(topic_type);
```
Koichi98: nit: this reads a bit asymmetrically — `load_typesupport_impl` writes to `type_support_handle_` as a side effect, and the next line passes that same member back into a free function. Could `load_typesupport_impl` return its result instead, so the constructor does the member assignment explicitly?

src/agnocastlib/src/agnocast_callback_info.cpp
```diff
@@ -28,6 +32,59 @@ uint32_t allocate_callback_info_id()
   return callback_info_id;
 }
 
+bool generic_serialize_message_in_place(
```
Koichi98: nit: `_in_place` feels misleading here — `raw` (input) and `out` (output) live in different memory, so this isn't really in-place. Could we rename it to something like `serialize_message` or `fill_serialized_message`?

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
```diff
@@ -419,6 +450,108 @@ class BasicPollingSubscriber
...
```
Koichi98: It seems like `load_typesupport_impl()` can throw. When it does, the destructor erases a garbage id from `id2_callback_info` and `mq_closes`  an arbitrary fd.

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_node.hpp
```diff
@@ -407,9 +427,16 @@ template <typename MessageT>
...
   const std::string message_type_name = rosidl_generator_traits::name<MessageT>();
+  send_performance_pubsub_bridge_registration_by_type_name(
+    topic_name, id, message_type_name, direction);
+}
+
+inline void send_performance_pubsub_bridge_registration_by_type_name(
+  const std::string & topic_name, topic_local_id_t id, const std::string & message_type_name,
+  BridgeDirection direction)
+{
+  static const auto logger = rclcpp::get_logger("agnocast_performance_service_bridge_registrar");
```
Koichi98:
```suggestion
  static const auto logger = rclcpp::get_logger("agnocast_performance_bridge_registrar");
```
?

src/agnocastlib/include/agnocast/agnocast_callback_info.hpp
```diff
@@ -84,10 +113,77 @@ TypeErasedCallback get_erased_callback(Func && callback)
...
+      std::is_invocable_v<F, std::unique_ptr<rclcpp::SerializedMessage>> ||
+      std::is_invocable_v<F, rclcpp::SerializedMessage &>,
+    "This callback type cannot be handled as a GenericCallback. "
+    "Callback must be invocable with one of the following arguments "
+    "(or any types implicitly convertible from them, e.g., const variants): "
+    "std::unique_ptr<rclcpp::SerializedMessage>, "
+    "std::shared_ptr<rclcpp::SerializedMessage>, or "
+    "rclcpp::SerializedMessage &.");
+
+  return [callback = std::forward<Func>(callback), type_support](AnyObject && arg) mutable {
```
Koichi98: The lambda holds `type_support` as a raw pointer, but doesn't hold the `shared_ptr<SharedLibrary>` that keeps it alive. Since `CallbackInfo` is copied out of `id2_callback_info` and queued before it runs, if the destructor of  `GenericSubscription` runs while a copy is in flight, it will `dlclose` the typesupport and dangle the pointer.

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
ruth561: Thank you for the careful review! Just to clarify: in C++, if a constructor throws, the object's destructor is never called — only the destructors of already-constructed members are. So if `load_typesupport_impl()` throws, `~GenericSubscription()` won't run, meaning there's no risk of closing a garbage fd or erasing a spurious map entry. This matches `BasicSubscription`'s design, so I kept them consistent.

By the way, there is a similar risk in `SubscriptionBase`. Its destructor always calls `ioctl(AGNOCAST_REMOVE_SUBSCRIBER_CMD)`, but `id_` is only assigned inside `initialize()`, not in the constructor itself. So if an exception is thrown in a derived class constructor before `initialize()` is called, the destructor would issue an ioctl with `id_==0`. Since id=0 is not a reserved sentinel — it is the actual ID assigned to the very first subscriber on any given topic — this would silently remove another subscriber's kernel entry. I've fixed this bug in [290db05](https://github.com/autowarefoundation/agnocast/pull/1393/commits/290db05f7121eae728f090015480e0826b5ee3bb) by initializing `id_` to `-1` and guarding the destructor with `if (id_ >= 0) { [...] }`.

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_node.hpp
ruth561: Thanks! Fixed in [098ffe4](https://github.com/autowarefoundation/agnocast/pull/1393/commits/098ffe4776bbb1e86e4a3a9bedd01b2d659afa91).

src/agnocastlib/src/agnocast_callback_info.cpp
ruth561: Renamed this function to `serialize_message` in [1bbfe67](https://github.com/autowarefoundation/agnocast/pull/1393/commits/1bbfe67daaa8727d0c9150c785a2f611f5c2480d).

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
ruth561: In 02cea6e, I changed the signature of load_typesupport_impl() to return the value explicitly.

src/agnocastlib/include/agnocast/agnocast_callback_info.hpp
ruth561: Good catch, thanks! Fixed in this [02cea6e](https://github.com/autowarefoundation/agnocast/pull/1393/commits/02cea6e44c08adc00371b6ba0d642174a58daf3c).

Introduced a `TypeSupportBundle` struct that holds both the `shared_ptr<rcpputils::SharedLibrary>` and the raw `rosidl_message_type_support_t*` handle together. `get_erased_generic_callback()` now accepts a `TypeSupportBundle` by value, and the lambda captures it. This means every copy of the `TypeErasedCallback` co-owns the `shared_ptr` to the type-support library, so the library cannot be `dlclosed` until all copies are gone—even if the `GenericSubscription` destructor runs in the meantime.

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
Koichi98: Thank you.

Koichi98 (dismissed):

Koichi98 (approved):

PR #1395, opened 2026-06-15
feat(agnocastlib): add `agnocast::GenericPublisher`

ruth561:
## Description

Add `agnocast::GenericPublisher`, which mirrors `rclcpp::GenericPublisher`.

The topic type is given as a runtime string (e.g. `"std_msgs/msg/String"`) instead of a compile-time template argument. Messages are passed to `publish()` as `rclcpp::SerializedMessage` objects and are deserialized into Agnocast shared memory inside the `publish()` call.

Only the Performance bridge mode is supported; under the Standard bridge mode no bridge request is issued and a warning is logged once.

## How it works

The design keeps all deserialization work inside `GenericPublisher::publish()` and reuses the existing publisher machinery (`PublisherBase`, `publish_core`, the kernel ioctl path) unchanged.

Key points:

- **Dynamic typesupport loading.** Two shared libraries (`rosidl_typesupport_cpp` and `rosidl_typesupport_introspection_cpp`) are loaded eagerly in the constructor and held for the publisher's lifetime.
- **Publish path.** On each `publish()` call, a buffer is heap-allocated, populated by `rmw_deserialize`, and passed to `publish_core` as a virtual address — exactly as a typed `Publisher<T>` would.
- **`PublisherRole` enum.** A small `PublisherRole` enum (`Default` / `AgnocastOnly` / `BridgeInternal`) is introduced to encode bridge-registration behaviour, replacing the template `BridgeRequestPolicy` pattern used by `BasicPublisher`.
- **Free functions.** `create_generic_publisher()` overloads (QoS and history-depth) are added to `agnocast.hpp`, mirroring the existing `create_publisher()` overloads.

## How was this PR tested?

- [ ] Autoware
- [x] sample application

Manual tests using the sample application were run as follows.

**Terminal 1 (talker):**

```bash
export AGNOCAST_BRIDGE_MODE=performance
bash scripts/sample_application/run_generic_talker.bash
```

**Terminal 2 (Case 1: A2A — listener using `GenericSubscription`):**

```bash
export AGNOCAST_BRIDGE_MODE=performance
bash scripts/sample_application/run_generic_listener.bash
```

**Terminal 2 (Case 2: A2R — ROS 2 subscriber):**

```bash
ros2 topic echo_agnocast /my_topic
```

## Notes for reviewers

A few things were intentionally left out of this PR to keep it small and keep reviewer load low. They are written down here so the discussion is not repeated in line comments.

- **`PublisherRole` design difference from `BasicPublisher`.** `BasicPublisher` uses a template `BridgeRequestPolicy`, while `GenericPublisher` uses the new `PublisherRole` enum. The intent is that `BasicPublisher` will adopt the same design in a follow-up refactor PR, so the two will end up consistent. This design difference is therefore out of scope for this PR.

`need-patch-update`

### DIFF
--- a/src/agnocastlib/src/agnocast_publisher.cpp
+++ b/src/agnocastlib/src/agnocast_publisher.cpp
@@ -300,7 +300,7 @@ rclcpp::QoS GenericPublisher::constructor_impl(
 
 GenericPublisher::GenericPublisher(
   rclcpp::Node * node, const std::string & topic_name, const std::string & topic_type,
-  const rclcpp::QoS & qos, agnocast::PublisherOptions options, PublisherRole role)
+  const rclcpp::QoS & qos, const PublisherOptions & options, PublisherRole role)
 {
   const rclcpp::QoS actual_qos = constructor_impl(node, topic_name, topic_type, qos, options, role);
 
@@ -312,7 +312,7 @@ GenericPublisher::GenericPublisher(
 
 GenericPublisher::GenericPublisher(
   agnocast::Node * node, const std::string & topic_name, const std::string & topic_type,
-  const rclcpp::QoS & qos, agnocast::PublisherOptions options, PublisherRole role)
+  const rclcpp::QoS & qos, const PublisherOptions & options, PublisherRole role)
 {
   const rclcpp::QoS actual_qos = constructor_impl(node, topic_name, topic_type, qos, options, role);
 
--- a/src/agnocastlib/src/agnocast_publisher.cpp
+++ b/src/agnocastlib/src/agnocast_publisher.cpp
@@ -1,11 +1,19 @@
 #include "agnocast/agnocast_publisher.hpp"
 
+#include "agnocast/bridge/agnocast_bridge_node.hpp"
 #include "agnocast/internal/type_registry_writer.hpp"
 #include "agnocast/node/agnocast_node.hpp"
 
+#include <rclcpp/typesupport_helpers.hpp>
+#include <rosidl_runtime_cpp/message_initialization.hpp>
+
+#include <rcutils/allocator.h>
+#include <rmw/rmw.h>
+#include <rmw/serialized_message.h>
 #include <sys/types.h>
 
 #include <array>
+#include <new>
 
 namespace agnocast
 {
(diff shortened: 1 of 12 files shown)


src/agnocastlib/src/agnocast_publisher.cpp
```diff
@@ -251,4 +259,126 @@ PublisherBase::~PublisherBase()
...
+
+  const auto va = reinterpret_cast<uint64_t>(ptr);
+  union ioctl_publish_msg_args publish_msg_args {
+  };
+  {
+    std::lock_guard<std::mutex> lock(opened_mqs_mtx_);
+    publish_msg_args = publish_core(this, topic_name_, id_, va, opened_mqs_);
+  }
+
+  decrement_borrowed_publisher_num();
```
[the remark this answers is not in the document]
ruth561: I moved `decrement_borrowed_publisher_num()` right after the `rmw_deserialize()` call.
Please take a look at [f0f879d](https://github.com/autowarefoundation/agnocast/pull/1395/commits/f0f879d3229c56248d05c88b3ac0766bfc86dc25).

src/agnocastlib/src/agnocast_publisher.cpp
```diff
@@ -251,4 +259,125 @@ PublisherBase::~PublisherBase()
...
+  void * ptr = ::operator new(members_->size_of_);
+
+  // Invoke the constructor of the message type at ptr.
+  // Type-specific initialization is unnecessary because the message object
+  // will be immediately populated with data by rmw_deserialize.
+  // Perform minimal initialization only.
+  members_->init_function(ptr, rosidl_runtime_cpp::MessageInitialization::SKIP);
+
+  const rmw_ret_t ret =
+    rmw_deserialize(&serialized_msg.get_rcl_serialized_message(), type_support_handle_, ptr);
```
Koichi98, line 351: Could you check whether this throws or not in CycloneDDS?

ruth561: I checked that this function catches all exceptions derived from `std::exception` [here](https://github.com/ros2/rmw_cyclonedds/blob/6b51c41d60fdbf87af6f490f322b1a5baa7bf1f3/rmw_cyclonedds_cpp/src/rmw_node.cpp#L1922-L1933). Exceptions that don't inherit from `std::exception` are not caught and will pass through to the caller. However, I think we can safely ignore this case because throwing non-standard exceptions is generally considered bad practice.

Since `rmw_deserialize` has an `extern "C"` signature and might be called from C, it should avoid throwing any exceptions (as specified in [the ROS 2 style guide](https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Code-Style-Language-Versions.html#exceptions)). If an exception is thrown to the caller, it would be an issue on the Cyclone DDS side rather than ours. I believe the original intent of this function is to ensure no exceptions escape to the caller.

Koichi98: Cool, thanks.

Koichi98 (approved):

Issue #943, opened 2026-01-20
Optimize receive_msg ioctl to copy only actual received entries


Koichi98:
Increasing `MAX_RECEIVE_NUM` should improve performance by allowing more messages to be received in a single ioctl call. However, the current implementation copies the entire array from kernel to user space, so we cannot increase `MAX_RECEIVE_NUM` significantly due to the copy overhead.                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  Modify the implementation to copy only the actually received entries (`ret_entry_num` elements) instead of the full array.                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                                                                                         
Note: When handling larger arrays, `copy_to_user` may not work correctly if the data crosses page boundaries. This needs to be considered in the implementation.  

#### Related Link
#937

scoot-skoot:
Hello! I am a Tech Fellow taking CodePath's AI301 course this summer, and I would love to take a stab at this issue. Could this issue be assigned to me?

vinuthnaDepala:
Hi! I am a student in AI 301 and am willing to take this issue on! I want to confirm if this issue is still open!

PR #1396, opened 2026-06-15
receive msg optimization

scoot-skoot:
## Description
After this change, instead of copying the full fixed-size buffer (`MAX_RECEIVE_NUM`),
we only copy the valid prefix (`ret_entry_num`) of `ret_entry_ids` and `ret_entry_addrs`.

This avoids unnecessary kernel-to-userspace copies while preserving the existing contract
that only indices in the range `[0, ret_entry_num - 1]` are valid.

This change does not affect ABI semantics, as userspace already relies on `ret_entry_num`
to determine the number of valid entries.

## Related links
Related to #943 
## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Notes for reviewers
This change does not modify ioctl fields or semantics.

`ret_entry_num` continues to define the valid range for `ret_entry_ids` and `ret_entry_addrs`.

The unused array elements are intentionally not copied to userspace.

Existing userspace already iterates only up to `ret_entry_num`.

`need-patch-update`

### DIFF
--- a/agnocast_kmod/agnocast_ioctl.c
+++ b/agnocast_kmod/agnocast_ioctl.c
@@ -2270,14 +2270,6 @@ static long receive_msg_cmd(union ioctl_receive_msg_args __user * arg)
   }
   kfree(pub_shm_infos);
 
-  // Function Replacement Target
-  /*
-  if (ret == 0) {
-    if (copy_to_user(arg, &receive_msg_args, sizeof(receive_msg_args))) return -EFAULT;
-  }
-  return ret;
-  */
-
   size_t n = receive_msg_args.ret_entry_num;
 
   if (n > MAX_RECEIVE_NUM) n = MAX_RECEIVE_NUM;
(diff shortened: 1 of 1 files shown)


Koichi98:
@scoot-skoot 
Thanks so much for your interest, but I'd like to close this PR, mainly for the following three reasons:
  
  1. Previously I thought we'd need to increase `MAX_RECEIVE_NUM`, which gave us the motivation to keep this data copy to a bare minimum and reduce copies. However, with PR https://github.com/autowarefoundation/agnocast/pull/1011, message reception and the execution of its callback are now done together in the executor (without allowing other callbacks to be interleaved by the executor), so I don't think we can make it very large anymore. We plan to revisit this in the future.
  2. Previously the lock in the kernel module was global in scope, but it's now a per-hashtable rwlock, and this section is no longer within the lock scope, so a bit of latency here won't affect other threads.
  3. If `MAX_RECEIVE_NUM` is around 10, the copy cost is essentially negligible — in fact, adding an extra if branch could well cost more, both at runtime and in terms of readability. At that scale, I'd rather prioritize readability.

Sorry I never got back to you even though you left a comment on https://github.com/autowarefoundation/agnocast/issues/943. I'll take a quick look for other up-to-date good-first-issues, so if you're still keen on contributing to Agnocast, could you give me a day?

scoot-skoot:
Thank you for the detailed explanation and for taking the time to review this. I'd definitely be interested in tackling another issue if you have any suggestions!

PR #1403, opened 2026-06-17
feat(agnocast_kmod, agnocast_ioctl_wrapper)[needs minor version update]: make topic list / info ioctls domain-aware

k1832:
## Description

After topic isolation by `ROS_DOMAIN_ID` (#1398), the same topic name can exist in multiple domains within one IPC namespace. The ros2cli-facing ioctls could not distinguish them: `get_topic_list` returned bare names, and `get_topic_subscriber_info` / `get_topic_publisher_info` resolved only the caller's domain. This makes those ioctls domain-aware so a per-namespace consumer (the discovery agent) can enumerate every domain.

Changes:

- **agnocast_kmod**: `get_topic_list` optionally fills a parallel `domain_id` array (one per topic name); `get_topic_subscriber_info` / `get_topic_publisher_info` take an explicit `domain_id` input and select the wrapper via `find_topic(name, ipc_ns, domain_id)`.
- **agnocast_ioctl_wrapper**: mirror the two arg structs so `_IOWR` encodes the same size. The wrapper passes `0` (no domain output / default domain), so userspace behavior is unchanged in this PR.

Scope: kmod / ioctl plumbing only. Consuming the domain (discovery agent stamps the real `domain_id`, decider keys bridge decisions on domain) lands in follow-up PRs.

## Related links

- Jira (TIER IV internal link): https://tier4.atlassian.net/browse/T4DEV-51894
- Design doc (TIER IV internal link): https://tier4.atlassian.net/wiki/x/4ABEPAE
- Stacked on #1398

## How was this PR tested?

- [ ] Autoware
- [ ] `bash scripts/test/e2e_test_1to1.bash` (required)
- [ ] `bash scripts/test/e2e_test_2to2.bash` (required)
- [x] kunit tests (required when modifying the kernel module)
- [ ] `bash scripts/test/run_requires_kernel_module_tests.bash` (required)
- [ ] sample application

In addition:

- **kunit (added)**: `get_topic_subscriber_info` selects the wrapper by `domain_id` — a subscriber registered in domain 1 is visible only for `domain_id == 1`, and domain 0 returns nothing.
- Built `agnocast_kmod` and `agnocast_ioctl_wrapper`; the wrapper ABI sync keeps the `ros2 agnocast` CLI working unchanged (e2e not run yet — draft).

## Version Update Label (Required)

- `need-minor-update` (kmod ioctl ABI: `ioctl_topic_list_args` / `ioctl_topic_info_args` layout change)

PR #1404, opened 2026-06-17
fix(agnocast_kmod): gate the performance bridge manager per (ipc_ns, domain)

k1832:
## Description

A performance bridge manager's MQ name carries the `ROS_DOMAIN_ID` suffix (`_d<domain>`), so each domain needs its own manager. The spawn gate `has_alive_performance_bridge_manager` matched only the IPC namespace, so a manager already running in one domain suppressed spawning a manager in another domain of the same namespace — the same-namespace / different-domain case (Case 2) never got its second manager.

Match the domain too: a process registering as a performance bridge manager is suppressed only when an alive manager already exists in the same `(ipc_ns, domain)`.

No ABI change — `add_process` already carries `domain_id` (#1398); this only refines the gate logic.

## Related links

- Jira (TIER IV internal link): https://tier4.atlassian.net/browse/T4DEV-51895
- Stacked on #1403

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

In addition:

- **kunit (added)**: `test_case_add_process_perf_manager_per_domain` — a manager in domain 0 does not suppress a manager in domain 1 (both registered), while a second manager in domain 0 is suppressed.
- Built `agnocast_kmod`.

- `need-patch-update` (no ABI change; backward-compatible spawn-gate fix)

### DIFF
--- a/agnocast_kmod/agnocast_ioctl.c
+++ b/agnocast_kmod/agnocast_ioctl.c
@@ -553,15 +553,19 @@ int agnocast_ioctl_get_version(struct ioctl_get_version_args * ioctl_ret)
   return 0;
 }
 
-static bool has_alive_performance_bridge_manager(const struct ipc_namespace * ipc_ns)
+// A performance bridge manager is per-(ipc_ns, domain): its MQ name carries the
+// domain suffix, so each domain needs its own manager. Gate on the domain too,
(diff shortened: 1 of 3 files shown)


k1832:
Superseded: this commit now lives in the combined kmod PR #1409 (C1+C2).

PR #1405, opened 2026-06-17
feat(agnocast_ioctl_wrapper, ros2agnocast_discovery_agent): stamp the real domain_id on discovered topics

k1832:
## Description

The discovery agent hardcoded `domain_id = 0` on every gossiped topic, so cross-namespace bridge decisions could not respect `ROS_DOMAIN_ID`. This threads the real domain through the ioctl wrapper so the per-namespace agent reports each topic's true domain.

Changes:

- **agnocast_ioctl_wrapper**: `get_agnocast_topics` fills a parallel `domain_id` array (one per topic name); `get_agnocast_sub_nodes` / `get_agnocast_pub_nodes` take a `domain_id` input that selects the wrapper.
- **ros2agnocast_discovery_agent**: pairs `name[i]` with `domain[i]`, builds one `AgnocastTopic` per `(name, domain)`, stamps the real `domain_id`, and forwards it to the endpoint queries.
- **ros2agnocast** (CLI): binding updated to the new wrapper signatures; it stays on the default domain (0). Full per-domain CLI output is a follow-up.

No kmod ABI change — this builds on the domain-aware ioctls from #1403.

## Related links

- Jira (TIER IV internal link): https://tier4.atlassian.net/browse/T4DEV-51894
- Design doc (TIER IV internal link): https://tier4.atlassian.net/wiki/x/4ABEPAE
- Stacked on #1404

## How was this PR tested?

- [ ] Autoware
- [ ] `bash scripts/test/e2e_test_1to1.bash` (required)
- [ ] `bash scripts/test/e2e_test_2to2.bash` (required)
- [ ] kunit tests (required when modifying the kernel module)
- [ ] `bash scripts/test/run_requires_kernel_module_tests.bash` (required)
- [ ] sample application

In addition:

- **pytest**: `ros2agnocast_discovery_agent` (44 passed, incl. new `test_read_local_topics_stamps_domain_id` asserting the real domain is stamped and forwarded) and `ros2agnocast` CLI (83 passed) with the updated bindings.
- Built `agnocast_ioctl_wrapper`.

## Version Update Label (Required)

- `need-patch-update` (userspace wrapper / agent only; no kmod ABI change)

PR #1408, opened 2026-06-17
feat(ros2agnocast_discovery_agent): make cross-NS bridge decisions domain-aware

k1832:
## Description

With the discovery agent now reporting each topic's real domain (#1405), the bridge decider must respect it. Otherwise the same topic name in two domains would be matched and bridged, breaking `ROS_DOMAIN_ID` isolation.

Changes (`bridge_decider.py`):

- Match local and remote topics by `(topic, domain)` — a cross-NS bridge is created only **within the same domain**. Cross-domain relaying is the external `domain_bridge`'s job.
- Dedup requests per `(topic, domain, direction)`.
- Dispatch each request to the bridge_manager that owns its domain (one manager per domain, addressed by the `_d<domain>` MQ suffix) instead of the agent's own `ROS_DOMAIN_ID`.

The wire payload is unchanged: the per-domain manager already runs in the target domain, so `domain_id` only routes the request and is not serialized (`sizeof(MqMsgDaemonBridge)` stays 524).

## Related links

- Jira (TIER IV internal link): https://tier4.atlassian.net/browse/T4DEV-51894
- Stacked on #1405

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

In addition:

- **pytest** (`ros2agnocast_discovery_agent`, 47 passed incl. flake8/pep257), with new cases: cross-domain match is skipped, same non-zero domain is bridged, and dispatch routes to the per-domain `_d<domain>` MQ.

- `need-patch-update` (discovery agent logic only; no ABI / wire-format change)

### DIFF
--- a/src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/bridge_decider.py
+++ b/src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/bridge_decider.py
@@ -62,6 +62,9 @@ class BridgeRequest:
     qos_depth: int
     qos_is_transient_local: bool
     qos_is_reliable: bool
+    # Selects the target bridge_manager's MQ (one manager per domain); not part
+    # of the wire payload, since that manager already runs in this domain.
+    domain_id: int = 0
 
 
 def serialize_request(req: BridgeRequest) -> bytes:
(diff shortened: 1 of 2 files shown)


k1832:
Superseded: this commit now lives in the combined discovery PR #1410 (C3+C4).

PR #1410, opened 2026-06-17
feat(agnocast_ioctl_wrapper, ros2agnocast_discovery_agent): make cross-NS discovery and bridge decisions domain-aware

k1832:
## Description

Thread the real domain through the discovery pipeline so cross-NS bridge decisions respect `ROS_DOMAIN_ID`, building on the domain-aware kmod ioctls in the base PR (#1409).

Changes:

- **Discovery agent stamps the real `domain_id`** (it was hardcoded to 0): `get_agnocast_topics` returns a parallel domain array and `get_agnocast_sub_nodes` / `get_agnocast_pub_nodes` take a `domain_id` input (ioctl wrapper); the per-NS agent pairs `name[i]` with `domain[i]` and builds one `AgnocastTopic` per `(name, domain)`. The `ros2agnocast` CLI binding is updated to the new signatures but stays on the default domain (0) — full per-domain CLI output is a follow-up.
- **Bridge decider is domain-aware**: matches local and remote topics by `(topic, domain)` so a bridge is created only within the same domain (cross-domain relaying is the external `domain_bridge`'s job, and bridging across domains would break isolation), dedups per `(topic, domain, direction)`, and dispatches each request to the bridge_manager that owns its domain (`_d<domain>` MQ).

No kmod ABI / wire-format change — the per-domain manager already runs in the target domain, so `domain_id` only routes the request and is not serialized.

## Related links

- Jira (TIER IV internal link): https://tier4.atlassian.net/browse/T4DEV-51894
- Builds on #1409 (merged)

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

In addition:

- **pytest**: `ros2agnocast_discovery_agent` 47 passed (incl. domain stamping + forwarding, cross-domain match skipped, same-domain bridged, per-domain `_d<domain>` MQ routing) and `ros2agnocast` CLI 83 passed with the updated bindings.
- Built `agnocast_ioctl_wrapper`.

- `need-patch-update` (userspace wrapper / agent / CLI only; no kmod ABI or wire-format change)

### DIFF
--- a/src/agnocast_ioctl_wrapper/src/topic_list.cpp
+++ b/src/agnocast_ioctl_wrapper/src/topic_list.cpp
@@ -11,9 +11,17 @@
 
 extern "C" {
 
-char ** get_agnocast_topics(int * topic_count)
+// When domain_ids_out is non-null, on success *domain_ids_out receives a malloc'd
+// array of *topic_count domain_ids parallel to the returned topic names; the caller
+// frees it with free_agnocast_topic_domains. The domain buffer is allocated here so
+// it always matches the name buffer's capacity — the caller never sizes it. Pass
+// nullptr to skip domain output.
+char ** get_agnocast_topics(int * topic_count, uint32_t ** domain_ids_out)
 {
   *topic_count = 0;
+  if (domain_ids_out != nullptr) {
+    *domain_ids_out = nullptr;
+  }
 
   int fd = open("/dev/agnocast", O_RDONLY);
   if (fd < 0) {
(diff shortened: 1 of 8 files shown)


Koichi98 (dismissed):

src/agnocast_ioctl_wrapper/src/topic_list.cpp
```diff
@@ -33,17 +44,30 @@ char ** get_agnocast_topics(int * topic_count)
     return nullptr;
   }
 
+  uint32_t * domain_buffer = nullptr;
```
Koichi98, line 47: (Really a #1409 point, I should have asked in the earlier PR)

The current implementation of ioctl returns all domains + a parallel `domain_id` array instead of filtering to the caller's own domain. It seems to me like, every caller only wants its own `ROS_DOMAIN_ID`, and data-path ioctls already scope to it via `find_topic_for_current`. Why can't we filter-out the other domains info inside the kernel module?

k1832: Really a good point, but this is intentional. 

The data-path ioctls (publish/receive/take) do scope to the caller's domain via find_topic_for_current. But get_topic_list is the discovery call and is
deliberately the wide one — it returns the whole namespace stamped by domain, while get_topic_{subscriber,publisher}_info takes a domain_id input and filters in-kernel. So
we do filter kernel-side, just on *_info, not list.

list stays wide for extensibility: today's only consumer (the discovery agent) filters to its own ROS_DOMAIN_ID, but returning all domains keeps the door open for a
cross-domain view (e.g. a future domain-aware ros2 topic list_agnocast). The filter is cheap.

Design discussion (TIER IV internal): https://tier4.atlassian.net/wiki/spaces/CRL/pages/5306056928/domain+bridge?focusedCommentId=5315395772

If you have a strong opinion on this, we can revisit this now, or discuss again in the near future.

Koichi98: Thanks. There shouldn't be a big problem so let's proceed and revisit later as needed.

k1832: @Koichi98 
Added a comment to add the intention in the code https://github.com/autowarefoundation/agnocast/pull/1410/changes/744f8986e741a527eaba3d5a0b96ae22d1adb055

Koichi98 (approved):

PR #1412, opened 2026-06-18
refactor(agnocastlib): remove the BridgeRegistrationPolicy template parameter from pub/sub classes

ruth561:
## Description

Removes the `BridgeRegistrationPolicy` template parameter from the pub/sub types (`Publisher`, `Subscription`, `TakeSubscription`, `PollingSubscriber`).

### Background

Previously, bridge registration behavior was controlled by a policy type as a second template argument (e.g., `BasicPublisher<MessageT, AgnocastToRosPubsubRegistrationPolicy>`). The user-facing types (like `Publisher<MessageT>`) were type aliases for these `Basic*` template classes. 

This design had the following drawbacks:

- **Code duplication**: The bridge registration call was duplicated in every `constructor_impl`. Each policy struct was just a basic wrapper doing the same thing.
- **Unnecessary template complexity**: The `BridgeRegistrationPolicy` parameter was originally added to fix a circular header dependency (#750). Since that issue is now resolved, we no longer need it. We can simply use an enum (`PublisherRole` / `SubscriptionRole`) as a constructor argument to control bridge requests.

> **Note:** Using a policy pattern does not significantly reduce runtime overhead here, because it is only called once during construction. Instead, removing it improves build times and reduces header dependencies.

### Changes

- **Replaced template classes with concrete classes**: `BasicPublisher<MessageT, Policy>`, `BasicSubscription<MessageT, Policy>`, etc., are now replaced by `Publisher<MessageT>`, `Subscription<MessageT>`, etc. These new classes take an enum (`PublisherRole` / `SubscriptionRole`) instead of a template policy.
- **Consolidated registration logic**: Bridge registration logic is merged into `PublisherBase::init_base` and `SubscriptionBase::initialize`. This removes duplicated calls from each class.
- **Moved implementation to `.cpp` file**: Moved `PublisherBase::init_base` from the header file to `agnocast_publisher.cpp` (with explicit instantiations for `rclcpp::Node` and `agnocast::Node`). This reduces header dependencies and speeds up incremental builds.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Notes for reviewers

This PR does not change `BasicService<ServiceT, BridgeRegistrationPolicy>`. Service-level bridge registration uses a different policy mechanism. We will update it in a future PR (follow-up).

## Post-merge checklist

`need-patch-update` because the public APIs marked as `AGNOCAST_PUBLIC` are not changed.

### DIFF
--- a/src/agnocastlib/include/agnocast/agnocast_subscription.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_subscription.hpp
@@ -10,21 +10,14 @@
 #include "rclcpp/detail/qos_parameters.hpp"
 #include "rclcpp/rclcpp.hpp"
 
-#include <fcntl.h>
 #include <mqueue.h>
-#include <string.h>
-#include <sys/ioctl.h>
-#include <sys/mman.h>
-#include <sys/stat.h>
 #include <unistd.h>
 
 #include <atomic>
 #include <cstdint>
-#include <cstring>
 #include <functional>
 #include <memory>
 #include <string>
-#include <thread>
 #include <vector>
 
 namespace agnocast
@@ -111,9 +104,9 @@ class SubscriptionBase
 protected:
   topic_local_id_t id_{-1};
   const std::string topic_name_;
-  union ioctl_add_subscriber_args initialize(
+  void initialize(
     const rclcpp::QoS & qos, const bool is_take_sub, const bool ignore_local_publications,
-    const bool is_bridge, const std::string & node_name, const std::string & type_name);
+    SubscriptionRole role, const std::string & node_name, const std::string & type_name);
 
 public:
   SubscriptionBase(rclcpp::Node * node, const std::string & topic_name);
@@ -140,9 +133,18 @@ public:
   }
 };
 
-// Internal implementation — users should use agnocast::Subscription<MessageT> instead.
-template <typename MessageT, typename BridgeRegistrationPolicy>
-class BasicSubscription : public SubscriptionBase
+/**
+ * @brief Agnocast subscription for a compile-time known message type.
+ *
+ * Delivers messages via a callback that is invoked each time a publisher
+ * writes to the topic. Allocate instances with
+ * `agnocast::create_subscription<MessageT>()` or construct directly.
+ *
+ * @tparam MessageT  ROS message type.
+ */
+AGNOCAST_PUBLIC
+template <typename MessageT>
+class Subscription : public SubscriptionBase
 {
   std::pair<mqd_t, std::string> mq_subscription_;
   uint32_t callback_info_id_;
@@ -151,7 +153,7 @@ class BasicSubscription : public SubscriptionBase
   rclcpp::QoS constructor_impl(
     NodeT * node, const rclcpp::QoS & qos, Func && callback,
     rclcpp::CallbackGroup::SharedPtr callback_group, agnocast::SubscriptionOptions options,
-    const bool is_bridge)
+    SubscriptionRole role)
   {
     const bool override_qos = options.qos_overriding_options.get_policy_kinds().size() > 0;
     rclcpp::node_interfaces::NodeParametersInterface::SharedPtr node_parameters =
@@ -172,11 +174,7 @@ class BasicSubscription : public SubscriptionBase
     if constexpr (rosidl_generator_traits::is_message<MessageT>::value) {
       type_name = rosidl_generator_traits::name<MessageT>();
     }
-    union ioctl_add_subscriber_args add_subscriber_args = initialize(
-      actual_qos, false, options.ignore_local_publications, is_bridge, node_name, type_name);
-
-    id_ = add_subscriber_args.ret_id;
-    BridgeRegistrationPolicy::template register_bridge<MessageT>(topic_name_, id_);
+    initialize(actual_qos, false, options.ignore_local_publications, role, node_name, type_name);
 
     mqd_t mq = open_mq_for_subscription(topic_name_, id_, mq_subscription_);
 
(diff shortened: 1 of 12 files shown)


src/agnocastlib/src/agnocast_subscription.cpp
```diff
@@ -42,14 +43,19 @@ union ioctl_add_subscriber_args SubscriptionBase::initialize(
...
     exit(EXIT_FAILURE);
   }
 
-  return add_subscriber_args;
+  id_ = add_subscriber_args.ret_id;
+
+  if (role == SubscriptionRole::Default) {
+    register_pubsub_bridge_by_type_name(
+      topic_name_, id_, type_name, BridgeDirection::ROS2_TO_AGNOCAST);
+  }
```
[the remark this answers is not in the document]
ruth561: Fixed in [382e959](https://github.com/autowarefoundation/agnocast/pull/1412/commits/382e9591819eca850a658f2196f836c446b1e2e9). Also, I added a warning log here for when `type_name` is empty.

src/agnocastlib/src/agnocast_publisher.cpp
```diff
@@ -201,6 +202,49 @@ uint32_t get_intra_subscription_count_core(const std::string & topic_name)
...
+
+  const bool is_bridge = (role == PublisherRole::BridgeInternal);
+  const std::string node_name = node->get_fully_qualified_name();
+  id_ = initialize_publisher(topic_name_, node_name, actual_qos, is_bridge, type_name);
+  generate_gid();
+
+  if (role == PublisherRole::Default) {
+    register_pubsub_bridge_by_type_name(
+      topic_name_, id_, type_name, BridgeDirection::AGNOCAST_TO_ROS2);
+  }
```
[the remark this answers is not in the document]
ruth561: Fixed in [382e959](https://github.com/autowarefoundation/agnocast/pull/1412/commits/382e9591819eca850a658f2196f836c446b1e2e9). Also, I added a warning log here for when `type_name` is empty.

src/agnocastlib/src/agnocast_publisher.cpp
```diff
@@ -201,6 +202,57 @@ uint32_t get_intra_subscription_count_core(const std::string & topic_name)
...
+  if (role == PublisherRole::Default) {
+    if (!type_name.empty()) {
+      register_pubsub_bridge_by_type_name(
+        topic_name_, id_, type_name, BridgeDirection::AGNOCAST_TO_ROS2);
+    } else {
+      RCLCPP_WARN(
+        logger,
+        "Bridge registration is skipped because the type_name is empty (topic: '%s'). "
+        "Please make sure to specify the valid message type in normal use case.",
+        topic_name_.c_str());
```
kobayu858: nit: This code path should normally be unreachable (since service types are always constructed with AgnocastOnly), but just in case, it would be helpful to distinguish whether the log came from the publisher or the subscription side.

src/agnocastlib/src/agnocast_subscription.cpp
```diff
@@ -42,14 +43,27 @@ union ioctl_add_subscriber_args SubscriptionBase::initialize(
...
+    if (!type_name.empty()) {
+      register_pubsub_bridge_by_type_name(
+        topic_name_, id_, type_name, BridgeDirection::ROS2_TO_AGNOCAST);
+    } else {
+      RCLCPP_WARN(
+        logger,
+        "Bridge registration is skipped because the type_name is empty (topic: '%s'). "
+        "Please make sure to specify the valid message type in normal use case.",
+        topic_name_.c_str());
+    }
```
kobayu858: nit: This code path should normally be unreachable (since service types are always constructed with AgnocastOnly), but just in case, it would be helpful to distinguish whether the log came from the publisher or the subscription side.

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
```diff
@@ -140,9 +133,18 @@ class SubscriptionBase
...
+/**
+ * @brief Agnocast subscription for a compile-time known message type.
+ *
+ * Delivers messages via a callback that is invoked each time a publisher
+ * writes to the topic. Allocate instances with
+ * `agnocast::create_subscription<MessageT>()` or construct directly.
+ *
+ * @tparam MessageT  ROS message type.
+ */
+AGNOCAST_PUBLIC
```
Koichi98, line 145: This Subscription class shouldn't be `AGNOCAST_PUBLIC` since this is not an exposed API and just an internal implementation. Same for `TakeSubscription` and `PollingSubscription`.

Koichi98: Seems like `Subscription` / `Publisher` and some other classes were already `AGNOCAST_PUBLIC` before this PR. My mistake.

src/agnocastlib/src/agnocast_publisher.cpp
ruth561: Thank you for pointing that out! I fixed it by using R2A and A2R representations.
Also, I changed the log level from WARN to ERROR, since an empty type_name is incorrect API usage. Please take a look at [2dda30f](https://github.com/autowarefoundation/agnocast/pull/1412/commits/2dda30f13c994266bc477c1150b03004b0513a48).

src/agnocastlib/src/agnocast_subscription.cpp
ruth561: Same as above.

Koichi98 (approved):

kobayu858 (approved):

PR #1416, opened 2026-06-19
refactor(agnocast_kmod): heap-allocate topic_struct behind a pointer

k1832:
> **Stacked on #1410.** Domain-bridge (51895) stack: **#1416 (this)** → #1417 → #1418.

## Description

Hold a `struct topic_struct *` in `topic_wrapper` instead of an inline struct, and move the per-topic `rwsem` into `topic_struct`, so a `topic_struct`'s lifetime can be managed independently of its wrapper. The next PR shares one `topic_struct` between two wrappers to deliver across ROS domains within one IPC namespace (zero-copy). **Pure refactor — no behavior change.**

First PR of the [51895] domain-bridge stack; based on #1410.

## Related links

- Jira (TIER IV internal link): https://tier4.atlassian.net/browse/T4DEV-51895
- Design doc (TIER IV internal link): https://tier4.atlassian.net/wiki/x/4ABEPAE

## How was this PR tested?

- [ ] Autoware
- [x] `bash scripts/test/e2e_test_1to1.bash` (required) — pass (run on the full stack tip; no regression)
- [x] `bash scripts/test/e2e_test_2to2.bash` (required) — pass
- [ ] kunit tests (required when modifying the kernel module) — runs in CI (kunit.py/QEMU); existing cases unchanged
- [x] `bash scripts/test/run_requires_kernel_module_tests.bash` (required) — pass
- [ ] sample application

Also: kmod builds with `-Werror`; clang-format clean.

## Version Update Label

- `need-patch-update` (internal kmod refactor; no ABI/behavior change)

PR #1418, opened 2026-06-19
feat(agnocast_ioctl_wrapper, ros2agnocast_discovery_agent): register domain bridge rules from YAML

k1832:
> **Stacked on #1417.** Domain-bridge (51895) stack: #1416 → #1417 → **#1418 (this)**.

## Description

Userspace path to register the kmod domain-bridge rules from #1417, plus the cross-domain e2e test.

- `agnocast_ioctl_wrapper`: exports `add_agnocast_domain_bridge_rule(topic, from_domain, to_domain)` + the matching ioctl arg struct / cmd.
- `ros2agnocast_discovery_agent`: a standalone `register_domain_bridge` console script reads a ROS 2 `domain_bridge` YAML and registers each `(topic, from, to)` rule. It is idempotent and exits non-zero if a rule is rejected. The discovery agent does **not** register rules.
- A reference `systemd` one-shot + README (`systemd/agnocast-domain-bridge.service.example`) document the required ordering.
- e2e `scripts/test/e2e_test_domain_bridge.bash`: registers a rule, runs the sample talker/listener in two domains in one IPC namespace, and checks the message crosses domains with no DDS / bridge node (zero-copy).

## Related links

- Jira (TIER IV internal link): https://tier4.atlassian.net/browse/T4DEV-51895

## How was this PR tested?

- [ ] Autoware
- [ ] kunit tests (required when modifying the kernel module) — N/A (kmod untouched)
- [ ] sample application

`colcon test` (`ros2agnocast_discovery_agent`) passes — `domain_bridge` YAML parsing (`parse_domain_bridge_config`, incl. malformed / out-of-range) and the `register_domain_bridge` tool (register / fail-loud / env fallback). Cross-domain e2e (`scripts/test/e2e_test_domain_bridge.bash`) on a freshly loaded kmod: talker (d1) reaches listener (d2), no DDS / bridge node.

## Notes for reviewers

A rule must be registered **before any publisher/subscriber for that topic exists in either domain** — otherwise the kmod returns `-EBUSY` (grouping merges the two domains' id spaces, safe only before anything is created; see #1417). The discovery agent comes up alongside the first node, so it cannot reliably precede those endpoints; that is why registration is a separate, one-time step run before applications — e.g. a boot step ordered after the kmod load and before the app launches (the shipped `.service.example`).

## Post-merge checklist

- [ ] Reflect in [`agnocast_doc`](https://github.com/autowarefoundation/agnocast_doc) — domain-bridge guide (autowarefoundation/agnocast_doc#41).

## Version Update Label

- `need-patch-update` (userspace + test only; no kmod / ABI change)

### DIFF
--- a/src/agnocast_ioctl_wrapper/src/domain_bridge.cpp
+++ b/src/agnocast_ioctl_wrapper/src/domain_bridge.cpp
@@ -16,6 +16,13 @@ extern "C" {
 int add_agnocast_domain_bridge_rule(
   const char * topic_name, uint32_t from_domain, uint32_t to_domain)
 {
+  // Exported C symbol: guard the pointer so a null from a caller returns an
+  // error instead of segfaulting in strlen() below.
+  if (topic_name == nullptr) {
+    errno = EINVAL;
+    return -1;
+  }
+
   int fd = open("/dev/agnocast", O_RDONLY);
   if (fd < 0) {
     if (errno == ENOENT) {
--- a/src/agnocast_ioctl_wrapper/src/domain_bridge.cpp
+++ b/src/agnocast_ioctl_wrapper/src/domain_bridge.cpp
@@ -1,0 +1,43 @@
+#include "agnocast_ioctl.hpp"
+
+#include <fcntl.h>
+#include <sys/ioctl.h>
+#include <unistd.h>
+
+#include <cerrno>
+#include <cstdio>
+#include <cstring>
+
+extern "C" {
+
+// Register a domain bridge rule with the kmod so it relays the topic between
+// from_domain and to_domain within this process's IPC namespace. Returns 0 on
+// success, -1 on failure.
+int add_agnocast_domain_bridge_rule(
+  const char * topic_name, uint32_t from_domain, uint32_t to_domain)
+{
+  int fd = open("/dev/agnocast", O_RDONLY);
+  if (fd < 0) {
+    if (errno == ENOENT) {
+      fprintf(stderr, "%s", AGNOCAST_DEVICE_NOT_FOUND_MSG);
+    } else {
+      perror("Failed to open /dev/agnocast");
+    }
+    return -1;
+  }
+
+  ioctl_add_domain_bridge_args args = {};
+  args.topic_name = {topic_name, strlen(topic_name)};
+  args.from_domain = from_domain;
+  args.to_domain = to_domain;
+
+  if (ioctl(fd, AGNOCAST_ADD_DOMAIN_BRIDGE_CMD, &args) < 0) {
+    perror("AGNOCAST_ADD_DOMAIN_BRIDGE_CMD failed");
+    close(fd);
+    return -1;
+  }
+
+  close(fd);
+  return 0;
+}
+}
--- a/src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/domain_bridge_config.py
+++ b/src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/domain_bridge_config.py
@@ -1,9 +1,9 @@
 """Parse a ROS 2 ``domain_bridge`` YAML into the (topic, from_domain, to_domain)
 rules the Agnocast daemon registers with the kmod.
 
-The same YAML drives both the external ``domain_bridge`` node (cross-ECU, Case 8)
-and the daemon's kmod rule injection that opens same-namespace zero-copy
-cross-domain delivery (Case 2). Only the topic name and domain pair matter here;
+The same YAML drives both the external ``domain_bridge`` node (cross-ECU, via DDS)
+and the daemon's kmod rule injection that opens same-IPC-namespace zero-copy
+cross-domain delivery. Only the topic name and domain pair matter here;
 ``type`` and other fields are ignored.
 """
 import yaml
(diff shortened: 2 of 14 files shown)


src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/domain_bridge_config.py
```diff
@@ -0,0 +1,39 @@
...
+
+    rules = []
+    for topic_name, spec in (doc.get('topics') or {}).items():
+        spec = spec or {}
+        from_domain = spec.get('from_domain', default_from)
+        to_domain = spec.get('to_domain', default_to)
+        if from_domain is None or to_domain is None:
+            continue
+        rules.append((str(topic_name), int(from_domain), int(to_domain)))
+    return rules
```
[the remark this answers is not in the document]
k1832: Done in 59a18a40. `parse_domain_bridge_config` now validates structure with `isinstance` (root, `topics`, and each topic `spec` must be mappings) and the domain range, raising `ValueError`/`TypeError` instead of `AttributeError` or a silent `uint32` wrap. The agent's `except (OSError, yaml.YAMLError, ValueError, TypeError)` already covers these, so a malformed config now leaves the daemon running without rules instead of crashing at startup. Added tests for non-mapping root/topics/spec and out-of-range/negative domains.

src/ros2agnocast_discovery_agent/setup.py
```diff
@@ -13,7 +13,7 @@
         ('share/' + package_name + '/launch', ['launch/discovery_agent.launch.xml']),
         ('lib/' + package_name, ['scripts/discovery_agent']),
     ],
-    install_requires=['setuptools'],
+    install_requires=['setuptools', 'pyyaml'],
     zip_safe=True,
```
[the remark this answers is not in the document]
k1832: Done in 59a18a40 — added `<exec_depend>python3-yaml</exec_depend>` to `ros2agnocast_discovery_agent/package.xml`.

src/agnocast_ioctl_wrapper/src/domain_bridge.cpp
```diff
@@ -0,0 +1,43 @@
...
+extern "C" {
+
+// Register a domain bridge rule with the kmod so it relays the topic between
+// from_domain and to_domain within this process's IPC namespace. Returns 0 on
+// success, -1 on failure.
+int add_agnocast_domain_bridge_rule(
+  const char * topic_name, uint32_t from_domain, uint32_t to_domain)
+{
+  int fd = open("/dev/agnocast", O_RDONLY);
+  if (fd < 0) {
```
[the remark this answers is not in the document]
k1832: Done in 59a18a40 — `add_agnocast_domain_bridge_rule` now returns `-1` with `errno=EINVAL` on a null `topic_name`, before the `strlen()`.

k1832:
I made the domain-bridge-rule registration a standalone executable. Let me carefully review and update the PR body before making this review-ready again

k1832:
Ready for review

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/register_domain_bridge.py
```diff
@@ -0,0 +1,72 @@
...
+    failures = 0
+    for topic, from_domain, to_domain in rules:
+        if add_rule(topic.encode('utf-8'), from_domain, to_domain) == 0:
+            print(f'registered: {topic} {from_domain}->{to_domain}')
+            continue
+        # A non-zero return most likely means an endpoint for the topic already
+        # exists in one of the domains; the rule must precede every node.
+        failures += 1
+        print(
+            f'error: failed to register {topic} {from_domain}->{to_domain}; '
```
Koichi98: If there are some other causes that `add_rule` could fail, this error message might be misleading.

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/domain_bridge_config.py
```diff
@@ -0,0 +1,67 @@
...
+    rules = []
+    for topic_name, spec in topics.items():
+        if spec is None:
+            spec = {}
+        elif not isinstance(spec, dict):
+            raise ValueError(f'spec for topic {topic_name!r} must be a mapping')
+        from_domain = spec.get('from_domain', default_from)
+        to_domain = spec.get('to_domain', default_to)
+        if from_domain is None or to_domain is None:
+            continue
```
Koichi98, line 63: This silently drops the topic. Could you consider surfacing it? Might be better in the caller rather than printing from this parser

k1832: Done. The functions now returns the skipped topic names as well instead of showing the warning msgs within the function, and the caller is responsible for showing the warning msgs.

d2a4ebc3

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/register_domain_bridge.py
k1832: Done. Removed the assumed cause. The real `errno` is reported in `add_rule` (-> `add_agnocast_domain_bridge_rule`) already.

Koichi98 (approved):

PR #1419, opened 2026-06-19
test(agnocast_e2e_test): add Case 2 cross-domain zero-copy e2e

k1832:
> **Stacked on #1418.** Domain-bridge (51895) stack: #1416 → #1417 → #1418 → **#1419 (this)**.

## Description

End-to-end test for Case 2 cross-domain zero-copy: registers a domain bridge rule, runs an Agnocast publisher in one domain and a subscriber in another within the same IPC namespace, and asserts the subscriber receives every message — exercising the kmod cross-domain matching end to end with no DDS / bridge node. Run via `scripts/test/e2e_test_domain_bridge.bash`. **Test-only.**

## How was this PR tested?

- [ ] Autoware
- [x] sample application — manual Case 2 (talker domain 1 / listener domain 2, same IPC NS) passes; the new `e2e_test_domain_bridge.bash` automates it

## Version Update Label

- `need-patch-update` (tests only)

### DIFF
--- a/scripts/test/e2e_test_domain_bridge.bash
+++ b/scripts/test/e2e_test_domain_bridge.bash
@@ -1,0 +1,42 @@
+#!/bin/bash
+# Case 2 (same IPC namespace, cross-domain) zero-copy delivery e2e.
+#
+# Registers a domain bridge rule with the kmod, then runs an Agnocast publisher
+# in FROM_DOMAIN and subscriber in TO_DOMAIN (same IPC namespace) and asserts the
+# subscriber receives every message — proving kmod cross-domain matching without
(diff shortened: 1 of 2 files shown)


k1832:
Folded into #1418 — the Case 2 e2e depends on the wrapper symbol introduced there and ships with the userspace path it validates.

k1832:
Folded into #1418: the Case 2 e2e (`scripts/test/e2e_test_domain_bridge.bash` + launch_test) is now a commit on the #1418 branch, since it depends on the `add_agnocast_domain_bridge_rule` wrapper symbol introduced there and a test belongs with the code it validates. Auto-marked merged because its commit landed in the base branch.

PR #1420, opened 2026-06-21
perf(agnocast_kmod): take read lock in receive/take ioctls

sykwer:
## Description

receive/take held the per-topic rwsem in write mode, serializing all subscribers on a topic. They only mutate disjoint per-subscriber state, so a read lock is sufficient and lets subscribers receive concurrently, removing the O(S) lock-wait that dominated tail latency under fan-out.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Post-merge checklist

**Important notes:**


sykwer:
## What
  `agnocast_ioctl_receive_msg` / `agnocast_ioctl_take_msg` now take the per-topic
  `topic_rwsem` in **read** mode instead of write. Publish keeps the write lock.

  ## Why
  The write lock serialized *all* subscribers on a topic, even though receive/take
  only mutate disjoint per-subscriber state. A read lock lets subscribers receive
  concurrently, eliminating the O(S) lock-wait that dominated tail latency under
  high subscriber fan-out. (This is the same design the `2.3.0-isorc26` evaluation
  branch reached; the timing instrumentation from that branch is intentionally not
  included here.)

  ## Why it's safe
  The read path still respects the two locks that guard everything it touches:

  - **Entries rbtree** is structurally mutated only under `topic_rwsem` **write**
    (publish: insert / depth-release) or `global_htables_rwsem` **write**
    (subscriber/publisher removal, process-exit, module unload). Receive holds
    `topic_rwsem` read **and** `global_htables_rwsem` read for its entire duration,
    so both write classes are mutually exclusive with it — the tree can never be
    mutated or freed underneath a receiver.
  - **Per-subscriber fields** (`latest_received_entry_id`, `need_mmap_update`) live
    in disjoint `sub_info` structs (subscriber ids are unique per topic), so
    cross-process receivers never collide. Same-subscriber receives are serialized
    in userspace by agnocastlib's process-global `mmap_mtx` (held across every
    receive/take ioctl, regardless of callback-group type). The kernel doesn't rely
    on that mutex for memory safety — a duplicate same-subscriber receive merely
    returns `-EALREADY` and never touches the tree.
  - **Reference bitmap** updates are atomic (`test_and_set_bit` / `test_and_clear_bit`).
  - **`reference_memory`** (mmap bookkeeping) is self-synchronized via `mempool_lock`;
    `proc_info` is freed only under `global_htables_rwsem` write (RCU).

  `release_message_entry_reference` already used this exact topic-read + atomic-bitop
  pattern, so this change extends an established, proven-safe discipline to receive/take.

  ABI unchanged (no ioctl struct / constant change) → **`need-patch-update`**.

Koichi98 (approved):

PR #1426, opened 2026-06-26
feat(agnocast_kmod)[needs minor version update]: manage the discovery agent in a kmod registry

k1832:
## Description

First PR of the kmod-managed discovery-agent (auto-fork) stack. The kmod owns the agent's **singleton and liveness in a dedicated registry**, so the fork gate and the singleton claim share one source of truth — no userspace lock file. This is the redesign requested in review (see the #1428 discussion, "manage the agent fully in the kmod"); it closes the orphan race the earlier flock approach had.

The agent runs **one per (IPC namespace, ROS_DOMAIN_ID)**, started by the first Agnocast node in that scope and self-exiting when the last one leaves.

- **Registry**: `discovery_agent_htable` (RCU, keyed by pid), one entry per live agent. Unlike `proc_info` there is no `exited` flag — the entry is removed the moment the agent exits, so "registered" always means "alive".
- **`AGNOCAST_ADD_DISCOVERY_AGENT` (30)** — atomic singleton claim: the first caller for a (NS, domain) wins; a duplicate loses (`ret_already_exists`) and must exit. Replaces the userspace flock.
- **`AGNOCAST_DISCOVERY_AGENT_SHOULD_EXIT` (29)** gains a `commit` flag: `false` = read-only idle poll (userspace counts consecutive idle polls before exiting); `true` = the atomic exit gate — deregisters the agent iff the (NS, domain) is truly empty, under the same write lock `add_process` takes, so a process starting during the grace period keeps the agent running instead of orphaning the scope.
- **`AGNOCAST_DISCOVERY_AGENT_EXISTS` (31)** — read-only liveness query for the CLI status verb (#1427).
- **`add_process`** returns `ret_discovery_agent_exist = (find_discovery_agent(ns, domain) != NULL)` — the first process sees `false` and forks the agent; later ones see `true`.
- The agent has no `proc_info`, so its liveness is wired into the exit pipeline (`is_agnocast_pid` + `agnocast_process_exit_cleanup`) and module teardown.

cmd numbers 30/31 follow 28 (`ADD_DOMAIN_BRIDGE`, now merged) and 29.

## Related links

- Discovery agent lifecycle discussion (TIER IV internal link): https://tier4.atlassian.net/wiki/x/4ABEPAE

## How was this PR tested?

- [ ] Autoware
- [x] `bash scripts/test/e2e_test_1to1.bash` (required) - tested on `#1428`
- [x] `bash scripts/test/e2e_test_2to2.bash` (required) - tested on `#1428`
- [x] kunit tests (required) — 190/190 via the `kunit` CI path (kunit.py + QEMU, KASAN + lockdep), incl. 9 discovery-agent cases: register first-wins / duplicate-loses / exist-reflects-registration / commit-exit-when-idle / commit-vetoed-when-busy / reaped-on-exit / **orphan-race** / exists-ioctl / commit-ignores-exited-process.
- [x] `bash scripts/test/run_requires_kernel_module_tests.bash` (required) - tested on `#1428`
- [ ] sample application

## Notes for reviewers

This is the redesign from the #1428 review discussion. The `orphan_race` KUnit case is the one to sanity-check: agent A commits and deregisters, a new process then sees no agent and forks a replacement that wins the claim — exactly one agent ends up registered, no orphan.

## Post-merge checklist

- `need-minor-update` (`add_process` ABI + new ioctls 30/31 + `commit` field on 29)

### DIFF
--- a/agnocast_kmod/agnocast.h
+++ b/agnocast_kmod/agnocast.h
@@ -47,6 +47,7 @@ union ioctl_add_process_args {
     uint64_t ret_shm_size;
     bool ret_unlink_daemon_exist;
     bool ret_performance_bridge_daemon_exist;
+    bool ret_discovery_agent_exist;
   };
 };
 
@@ -266,6 +267,12 @@ struct ioctl_check_and_request_bridge_shutdown_args
   bool ret_should_shutdown;
 };
 
+struct ioctl_discovery_agent_should_exit_args
+{
+  uint32_t domain_id;
+  bool ret_should_exit;
+};
+
 struct ioctl_set_ros2_subscriber_num_args
 {
   struct name_info topic_name;
@@ -309,6 +316,8 @@ struct ioctl_add_domain_bridge_args
 #define AGNOCAST_SET_ROS2_PUBLISHER_NUM_CMD _IOW(0xA6, 26, struct ioctl_set_ros2_publisher_num_args)
 #define AGNOCAST_NOTIFY_BRIDGE_SHUTDOWN_CMD _IO(0xA6, 27)
 #define AGNOCAST_ADD_DOMAIN_BRIDGE_CMD _IOW(0xA6, 28, struct ioctl_add_domain_bridge_args)
+#define AGNOCAST_DISCOVERY_AGENT_SHOULD_EXIT_CMD \
+  _IOWR(0xA6, 29, struct ioctl_discovery_agent_should_exit_args)
 
 // ================================================
 // ros2cli ioctls
@@ -483,6 +492,9 @@ int agnocast_ioctl_set_ros2_publisher_num(
   const char * topic_name, const struct ipc_namespace * ipc_ns, uint32_t count);
 
 int agnocast_ioctl_notify_bridge_shutdown(const pid_t pid);
+
+int agnocast_ioctl_discovery_agent_should_exit(
+  const struct ipc_namespace * ipc_ns, const uint32_t domain_id, bool * ret_should_exit);
 
 int agnocast_ioctl_get_exit_process(
   const struct ipc_namespace * ipc_ns, struct ioctl_get_exit_process_args * ioctl_ret,
--- a/agnocast_kmod/agnocast.h
+++ b/agnocast_kmod/agnocast.h
@@ -267,10 +267,33 @@ struct ioctl_check_and_request_bridge_shutdown_args
   bool ret_should_shutdown;
 };
 
+// The caller's IPC namespace and pid come from `current` in the kmod, so they are
+// not passed here; only the domain is (the kernel cannot infer a ROS_DOMAIN_ID).
 struct ioctl_discovery_agent_should_exit_args
 {
   uint32_t domain_id;
+  // false: read-only idle poll (userspace counts consecutive idle polls before exiting).
+  // true: atomic exit gate -- deregister iff the domain is truly empty, serialized against
+  // add_process so a process racing in vetoes the exit instead of being orphaned.
+  bool commit;
   bool ret_should_exit;
+};
+
+// The caller's IPC namespace and pid come from `current` in the kmod, so they are
+// not passed here; only the domain is (the kernel cannot infer a ROS_DOMAIN_ID).
+struct ioctl_add_discovery_agent_args
+{
+  uint32_t domain_id;
+  bool ret_already_exists;  // true: another agent already owns this (ns, domain); caller must exit
+};
+
+// The caller's IPC namespace and pid come from `current` in the kmod, so they are
+// not passed here; only the domain is (the kernel cannot infer a ROS_DOMAIN_ID).
+struct ioctl_discovery_agent_exists_args
+{
+  uint32_t domain_id;
+  bool
+    ret_exists;  // whether an agent is registered for this (ns, domain); read-only liveness query
 };
 
 struct ioctl_set_ros2_subscriber_num_args
@@ -318,6 +341,9 @@ struct ioctl_add_domain_bridge_args
 #define AGNOCAST_ADD_DOMAIN_BRIDGE_CMD _IOW(0xA6, 28, struct ioctl_add_domain_bridge_args)
 #define AGNOCAST_DISCOVERY_AGENT_SHOULD_EXIT_CMD \
   _IOWR(0xA6, 29, struct ioctl_discovery_agent_should_exit_args)
+#define AGNOCAST_ADD_DISCOVERY_AGENT_CMD _IOWR(0xA6, 30, struct ioctl_add_discovery_agent_args)
+#define AGNOCAST_DISCOVERY_AGENT_EXISTS_CMD \
+  _IOWR(0xA6, 31, struct ioctl_discovery_agent_exists_args)
 
 // ================================================
 // ros2cli ioctls
@@ -494,7 +520,15 @@ int agnocast_ioctl_set_ros2_publisher_num(
 int agnocast_ioctl_notify_bridge_shutdown(const pid_t pid);
 
 int agnocast_ioctl_discovery_agent_should_exit(
-  const struct ipc_namespace * ipc_ns, const uint32_t domain_id, bool * ret_should_exit);
+  const pid_t pid, const struct ipc_namespace * ipc_ns, const uint32_t domain_id, const bool commit,
+  bool * ret_should_exit);
+
+int agnocast_ioctl_add_discovery_agent(
+  const pid_t pid, const struct ipc_namespace * ipc_ns, const uint32_t domain_id,
+  struct ioctl_add_discovery_agent_args * ioctl_ret);
+
+int agnocast_ioctl_discovery_agent_exists(
+  const struct ipc_namespace * ipc_ns, const uint32_t domain_id, bool * ret_exists);
 
 int agnocast_ioctl_get_exit_process(
   const struct ipc_namespace * ipc_ns, struct ioctl_get_exit_process_args * ioctl_ret,
@@ -518,6 +552,7 @@ int agnocast_increment_message_entry_rc(
   const char * topic_name, const struct ipc_namespace * ipc_ns, const topic_local_id_t pubsub_id,
   const int64_t entry_id);
 int agnocast_get_alive_proc_num(void);
+int agnocast_get_discovery_agent_num(void);
 bool agnocast_is_proc_exited(const pid_t pid);
 int agnocast_get_topic_entries_num(const char * topic_name, const struct ipc_namespace * ipc_ns);
 int64_t agnocast_get_latest_received_entry_id(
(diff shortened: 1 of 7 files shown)


k1832:
Marking ready for review — base of the auto-fork stack (#1426 → #1427 → #1428).

The new discovery-agent lifecycle hooks (the `add_process` first-in-`(NS, domain)` gate and the `should_exit` ioctl) are covered by kunit under KASAN + lockdep.

The required e2e checkboxes are left **unchecked**: bridge-ON e2e currently fails on `main` itself due to a pre-existing performance-bridge problem (#1386), unrelated to this stack. Stack-level runtime validation (auto-fork / singleton / opt-out / idle-exit / lifetime / per-domain) is on the capstone #1428. I'll tick the required boxes once #1386 is resolved.

k1832:
Thanks for the review — addressed the discovery-agent gate comments:

- **Counting exited processes:** `get_process_num_in_domain()` counted `exited`-but-not-yet-drained entries, so `ret_discovery_agent_exist` / the should-exit ioctl could be wrong during the unlink daemon's drain window. Added a dedicated `get_alive_process_num_in_domain()` (skips `->exited`) and used it for both. I intentionally left the shared `get_process_num_in_domain()` — and its perf-bridge-manager caller from #1409 — unchanged; that path has the same latent pattern but is outside this PR's scope (worth a separate look).
- **Missing edge-case test:** added `test_case_discovery_agent_ignores_exited_process`, which drives a process to exited-but-not-drained (`agnocast_enqueue_exit_pid`) and asserts the gate treats the domain as empty.

Also rebased onto latest `main` (force-pushed).

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -599,6 +600,7 @@ int agnocast_ioctl_add_process(
   ioctl_ret->ret_unlink_daemon_exist = (get_process_num(ipc_ns) > 0);
   ioctl_ret->ret_performance_bridge_daemon_exist =
     has_alive_performance_bridge_manager(ipc_ns, domain_id);
+  ioctl_ret->ret_discovery_agent_exist = (get_process_num_in_domain(ipc_ns, domain_id) > 0);
```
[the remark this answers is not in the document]
k1832: Addressed by the kmod-registry rework: `ret_discovery_agent_exist` no longer derives from a process count — the fork gate now reads `find_discovery_agent()` on the new discovery-agent registry, so exited-but-not-drained `proc_info` entries cannot affect it. (84e43bbc)

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -2192,6 +2194,16 @@ int agnocast_ioctl_notify_bridge_shutdown(const pid_t pid)
...
 
+// The agent is not a registered Agnocast process, so it passes its own ROS_DOMAIN_ID.
+int agnocast_ioctl_discovery_agent_should_exit(
+  const struct ipc_namespace * ipc_ns, const uint32_t domain_id, bool * ret_should_exit)
+{
+  down_read(&global_htables_rwsem);
+  *ret_should_exit = (get_process_num_in_domain(ipc_ns, domain_id) == 0);
+  up_read(&global_htables_rwsem);
+  return 0;
+}
```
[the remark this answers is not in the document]
k1832: Same rework: the self-exit path is `discovery_agent_should_exit(commit=true)`, which uses `get_alive_process_num_in_domain()` (skips `->exited`) under the write lock, so a lingering exited entry never blocks the exit. (84e43bbc)

agnocast_kmod/agnocast_kunit/agnocast_kunit_add_process.c
```diff
@@ -92,6 +92,51 @@ void test_case_add_process_perf_manager_per_domain(struct kunit * test)
...
+    test, agnocast_ioctl_add_process(pid++, current->nsproxy->ipc_ns, false, 7, &args), 0);
+
+  KUNIT_ASSERT_EQ(
+    test, agnocast_ioctl_discovery_agent_should_exit(current->nsproxy->ipc_ns, 7, &should_exit), 0);
+  KUNIT_EXPECT_FALSE(test, should_exit);  // domain 7 now has a process
+
+  KUNIT_ASSERT_EQ(
+    test, agnocast_ioctl_discovery_agent_should_exit(current->nsproxy->ipc_ns, 8, &should_exit), 0);
+  KUNIT_EXPECT_TRUE(test, should_exit);  // a different, empty domain
+}
```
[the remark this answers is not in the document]
k1832: Good catch — re-added as `test_case_discovery_agent_commit_ignores_exited_process`: it drives a process to exited-but-not-drained (`agnocast_enqueue_exit_pid` + wait) and asserts the commit still deregisters the agent (the exited entry does not veto). (84e43bbc)

k1832:
Self review in progress

agnocast_kmod/agnocast.h
```diff
@@ -266,6 +267,29 @@ struct ioctl_check_and_request_bridge_shutdown_args
...
+  uint32_t domain_id;
+  bool ret_already_exists;  // true: another agent already owns this (ns, domain); caller must exit
+};
+
+struct ioctl_discovery_agent_exists_args
+{
+  uint32_t domain_id;
+  bool
+    ret_exists;  // whether an agent is registered for this (ns, domain); read-only liveness query
+};
```
k1832, line 297: Add a comment above to each of these structs. The comment should be like:
```
// IPC NS identity will be resolved with pid, so it does not have to be passed explicitly.
```

agnocast_kmod/agnocast_internal.h
```diff
@@ -44,6 +44,8 @@ extern struct rw_semaphore global_htables_rwsem;
 #define PUB_INFO_HASH_BITS 3
 #define SUB_INFO_HASH_BITS 5
 #define PROC_INFO_HASH_BITS 10
+// At most one agent per (IPC namespace, domain), so the table is tiny.
+#define DISCOVERY_AGENT_HASH_BITS 4
```
k1832, line 48: How do we make sure this is enough? Is it possible to catch the overflow at runtime?

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -2326,6 +2330,26 @@ static int get_process_num_in_domain(const struct ipc_namespace * ipc_ns, const
...
+  const struct ipc_namespace * ipc_ns, const uint32_t domain_id)
+{
+  int count = 0;
+  struct process_info * proc_info;
+  int bkt_proc_info;
+  hash_for_each(proc_info_htable, bkt_proc_info, proc_info, node)
+  {
+    if (
+      ipc_eq(ipc_ns, proc_info->ipc_ns) && proc_info->domain_id == domain_id &&
+      !proc_info->exited) {
```
k1832, line 2346: What's the proc_info for, when this comment is present?
```
// The discovery agent's liveness, owned by the kmod so the fork gate and the
// singleton claim share one source of truth (no userspace flock). Hashed by pid
// (removal and is_agnocast_pid() are by pid); a (ns, domain) lookup scans. Unlike
// process_info there is no `exited` flag: the entry is removed the moment the
// agent exits, so "registered" always means "alive".
```

agnocast_kmod/agnocast.h
k1832: Done (62134582) — added above each of the three discovery-agent ioctl arg structs. One small rewording from the template: the kmod reads **both** the IPC namespace (`current->nsproxy->ipc_ns`) and the pid (`current->tgid`) from the caller directly — the ns isn't derived *from* the pid (the agent has no `proc_info` to look it up in). So:

```c
// The caller's IPC namespace and pid come from `current` in the kmod, so they are
// not passed here; only the domain is (the kernel cannot infer a ROS_DOMAIN_ID).
```

agnocast_kmod/agnocast_internal.h
k1832: There's nothing to overflow: `DECLARE_HASHTABLE(_, 4)` is 2⁴ = 16 **buckets**, and `hash_add` chains entries onto a bucket's `hlist` — so more entries than buckets just lengthen chains, never fail. It's a collision bound, not a capacity limit, so there's no runtime overflow to catch.

Is 4 enough? Live agents are bounded by "one per `(IPC namespace, ROS_DOMAIN_ID)`" actually in use — realistically single digits — so 16 buckets gives ~no collisions. (And `find_discovery_agent()` scans all buckets anyway, since it looks up by `(ns, domain)`, not the pid key; the bucket count only affects the pid-keyed `remove_by_pid` / `is_agnocast_pid`.) Even a pathological population would only make those marginally slower, never overflow.

agnocast_kmod/agnocast_ioctl.c
k1832: `process_info` tracks **regular Agnocast processes** (pub/sub apps + the perf bridge manager): it owns a shared-memory mempool, exit-subscription lists, and an `exited` flag that lingers until the unlink daemon drains that process's exit MQ. The discovery agent is deliberately **not** a `proc_info`:

1. it has no mempool and does no Agnocast pub/sub — a `proc_info` would allocate one it never uses;
2. it must **not** be counted by `get_alive_process_num_in_domain()` — that count is exactly what the agent's own `should_exit` checks ("any *app* left?"); counting itself would keep it from ever seeing the domain empty;
3. it wants "registered == alive" (removed the instant it exits), with no `exited`-flag drain window, since there's no exit MQ to drain for it.

So `process_info` = the app processes the agent observes and whose presence gates its lifecycle; the registry = the agent itself.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -2337,6 +2361,92 @@ int agnocast_ioctl_notify_bridge_shutdown(const pid_t pid)
...
+
+// Atomic singleton claim; replaces the userspace flock. The first caller for a (ns, domain) wins
+// and is recorded; a later caller loses (ret_already_exists = true) and must exit.
+int agnocast_ioctl_add_discovery_agent(
+  const pid_t pid, const struct ipc_namespace * ipc_ns, const uint32_t domain_id,
+  struct ioctl_add_discovery_agent_args * ioctl_ret)
+{
+  int ret = 0;
+  down_write(&global_htables_rwsem);
+
```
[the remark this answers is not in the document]
k1832: Fixed in 0f06fa57 — `ret_already_exists` is now initialized to `false` at the top of the function, so the `-ENOMEM` path returns a deterministic value instead of whatever the caller passed in. (The redundant assignment on the success path was dropped in favor of that single default.)

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -3114,6 +3271,18 @@ int agnocast_get_alive_proc_num(void)
...
+{
+  int count = 0;
+  struct discovery_agent_info * agent;
+  int bkt;
+  hash_for_each(discovery_agent_htable, bkt, agent, node)
+  {
+    count++;
+  }
+  return count;
+}
```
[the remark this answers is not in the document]
k1832: Fixed in 0f06fa57 — `agnocast_get_discovery_agent_num()` now takes `down_read(&global_htables_rwsem)` around the traversal, serializing it against the exit path's `hash_del_rcu()` + `kfree_rcu()` (which run under the write lock). This matches the discovery-agent htable's documented discipline (read to find/traverse, write to remove), so the count stays stable even while the orphan-race test's worker deletes entries.

Koichi98 (approved):

PR #1429, opened 2026-06-29
fix(ros2agnocast): don't flag a missing discovery agent when there is no Agnocast

k1832:
## Description

The `_agnocast` CLI verbs printed `NOTE: no /_agnocast_discovery agent visible; start one ...` whenever gossip was empty and they fell back to ioctl. In a namespace with no Agnocast at all that note is misleading. It is now emitted only when the ioctl fallback actually found local Agnocast state; the publisher-visible diagnostic is unchanged.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

Python-only CLI change (no kmod / data-plane impact): `colcon test --packages-select ros2agnocast` → 81 passed, including the updated and added `warn_if_using_fallback` cases.

## Post-merge checklist

`need-patch-update` (bug fix).

### DIFF
--- a/src/ros2agnocast/ros2agnocast/discovery.py
+++ b/src/ros2agnocast/ros2agnocast/discovery.py
@@ -15,7 +15,7 @@ from rclpy.qos import (
 
 from ros2agnocast._ioctl import self_ns_snapshot
 
-from ros2agnocast_discovery_msgs.msg import AgnocastDaemonState
+from ros2agnocast_discovery_msgs.msg import AgnocastDaemonState, AgnocastEndpoint
 
 
 GOSSIP_TOPIC = '/_agnocast_discovery'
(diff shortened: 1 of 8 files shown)


src/ros2agnocast/ros2agnocast/discovery.py
```diff
@@ -132,6 +136,11 @@ def warn_if_using_fallback(node, used_fallback: bool, timeout_sec: float) -> Non
         publishers = []
 
     if not publishers:
+        # No agent is gossiping. Only worth a note when this namespace actually
+        # has Agnocast endpoints to show: ros2 CLI runs constantly in namespaces
+        # without Agnocast, where "no agent" would be misleading noise.
+        if not snapshots:
```
Koichi98: Doesn't `self_ns_snapshot()` return a non-None state (with empty `topics`) even when the kmod is not loaded or nothing is registered? If so, the first return value of `collect_announcements_with_fallback` is `[state]`, not an empty list, so this `if not snapshots: return` never triggers.

src/ros2agnocast/ros2agnocast/discovery.py
```diff
@@ -120,8 +120,12 @@ def _resolve_spin_node(node):
...
 
 
-def warn_if_using_fallback(node, used_fallback: bool, timeout_sec: float) -> None:
-    """Best-effort stderr note when gossip was unavailable and ioctl fallback ran."""
+def warn_if_using_fallback(node, used_fallback: bool, timeout_sec: float, snapshots) -> None:
+    """Best-effort stderr note when gossip was unavailable and ioctl fallback ran.
+
+    ``snapshots`` is the fallback result; an empty list means this namespace has
+    no Agnocast state at all, so a missing agent is not worth mentioning.
+    """
```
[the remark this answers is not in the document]
k1832: Done. (Amended)

src/ros2agnocast/ros2agnocast/discovery.py
k1832: Thanks for catching this. Done (amended).

To improve this further, I also modified the annotated types and the docs of the functions.

Koichi98 (approved):

PR #1430, opened 2026-06-29
refactor(agnocastlib): add TypeErasedPublisher

bdm-k:
## Description

This PR introduces a new internal `TypeErasedPublisher` abstraction and refactors `GenericPublisher` to inherit from it. This low-level abstraction will be needed when we implement `GenericClient` in the future.

### What is `TypeErasedPublisher`

Please refer to the doc comment for information about what `TypeErasedPublisher` is. The comment summarizes the differences from `BasicPublisher`.

`TypeErasedPublisher` will be used to implement `GenericClient` in the near future. The interface of `GenericClient` has nothing to do with message serialization and is similar to the standard `Client`. In Agnocast, therefore, `GenericClient` would need to provide an API to allocate requests in shared memory (i.e., `borrow_loaned_request()`). In order to provide such an API, `GenericPublisher` is too specialized, and we need `TypeErasedPublisher` instead.

### `ipc_shared_ptr` updates for type-erased safety

`TypeErasedPublisher` uses `ipc_shared_ptr<void>` for its interface. A few changes are made to allow usage of `ipc_shared_ptr<void>`

- Guarded `operator*` and `oeprator->` for non-void specialization only.
- Added fatal behavior if an `ipc_shared_ptr<void>` is dropped before it is either publisher or canceled. We need this assertion because `ipc_shared_ptr<void>` cannot know how to free the message it points to.

## How was this PR tested?

- [ ] Autoware
- [x] sample application (`minimal_generic_publisher` works fine)

## Post-merge checklist

### DIFF
--- a/src/agnocastlib/include/agnocast/agnocast_publisher.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_publisher.hpp
@@ -274,6 +274,85 @@ public:
 };
 
 /**
+ * @brief A type-erased Agnocast publisher.
+ *
+ * There are four differences between this and BasicPublisher:
+ *
+ * 1. borrow_loaned_message() takes a size and returns an ipc_shared_ptr<void> that points to a
+ *    shared memory block of the requested size.
+ * 2. publish() takes a deleter as well as the message. Because the message is type-erased, the
+ *    publisher cannot know how to free it, so the caller must provide a deleter.
+ * 3. If a user decides not to publish a borrowed message, they must call cancel_message() with a
+ *    deleter to free the memory.
+ * 4. The constructor is just a thin wrapper around init_base(). As this class is intended for
+ *    internal use only, flexibility is prioritized.
+ */
+class TypeErasedPublisher : public PublisherBase
+{
+public:
+  using SharedPtr = std::shared_ptr<TypeErasedPublisher>;
+
+  TypeErasedPublisher(
+    rclcpp::Node * node, const std::string & topic_name, const std::string & topic_type,
+    const rclcpp::QoS & qos, const PublisherOptions & options, const bool is_bridge);
+
+  TypeErasedPublisher(
+    agnocast::Node * node, const std::string & topic_name, const std::string & topic_type,
+    const rclcpp::QoS & qos, const PublisherOptions & options, const bool is_bridge);
+
+  ipc_shared_ptr<void> borrow_loaned_message(size_t size);
+
+  template <typename Deleter>
+  void cancel_message(ipc_shared_ptr<void> && message, Deleter && deleter)
+  {
+    if (!message || topic_name_ != message.get_topic_name()) {
+      RCLCPP_ERROR(logger, "Invalid message to cancel.");
+      close(agnocast_fd);
+      exit(EXIT_FAILURE);
+    }
+
+    message.invalidate_all_references();
+
+    decrement_borrowed_publisher_num();
+
+    deleter(message.get());
+
+    message.reset();
+  }
+
+  template <typename Deleter>
+  void publish(ipc_shared_ptr<void> && message, Deleter && deleter)
+  {
+    if (!message || topic_name_ != message.get_topic_name()) {
+      RCLCPP_ERROR(logger, "Invalid message to publish.");
+      close(agnocast_fd);
+      exit(EXIT_FAILURE);
+    }
+
+    const uint64_t msg_virtual_address = reinterpret_cast<uint64_t>(message.get());
+
+    message.invalidate_all_references();
+
+    decrement_borrowed_publisher_num();
+
+    union ioctl_publish_msg_args publish_msg_args;
+    {
+      std::lock_guard<std::mutex> lock(opened_mqs_mtx_);
+      publish_msg_args = publish_core(this, topic_name_, id_, msg_virtual_address, opened_mqs_);
+    }
+
+    for (uint32_t i = 0; i < publish_msg_args.ret_released_num; i++) {
+      void * release_ptr = reinterpret_cast<void *>(publish_msg_args.ret_released_addrs[i]);
+      deleter(release_ptr);
+    }
+
+    message.reset();
+  }
+};
+
+// TOOD(bdm-k): Incorporate TypeErasedPublisher into GenericPublisher to increase code reuse.
+
+/**
  * @brief Mirrors `rclcpp::GenericPublisher` semantics: the topic type is supplied as a
  * runtime string (e.g. "std_msgs/msg/String") rather than a compile-time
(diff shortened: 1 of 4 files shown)


src/agnocastlib/include/agnocast/agnocast_publisher.hpp
```diff
@@ -240,6 +240,83 @@ class Publisher : public PublisherBase
   }
 };
 
+/**
+ * @brief A type-erased Agnocast publisher.
+ *
+ * There are four differences between this and BasicPublisher:
```
ruth561: `BasicPublisher` no longer exists.

```suggestion
 * There are four differences between this and Publisher:
```

ruth561 (commented):
Overall, this looks great! 👍
Just one minor suggestion about testing. Although this class is intended to be used internally within the `agnocast` library, it would be even better if you could add some tests for it, especially for the `cancel_message()` code path.

src/agnocastlib/include/agnocast/agnocast_publisher.hpp
```diff
@@ -240,6 +240,83 @@ class Publisher : public PublisherBase
...
+      RCLCPP_ERROR(logger, "Invalid message to cancel.");
+      close(agnocast_fd);
+      exit(EXIT_FAILURE);
+    }
+
+    message.invalidate_all_references();
+
+    decrement_borrowed_publisher_num();
+
+    deleter(message.get());
```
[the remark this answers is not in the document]
bdm-k: `ipc_shared_ptr::get()` does not modify the pointer.

src/agnocastlib/include/agnocast/agnocast_publisher.hpp
```diff
@@ -240,6 +240,83 @@ class Publisher : public PublisherBase
...
+    union ioctl_publish_msg_args publish_msg_args;
+    {
+      std::lock_guard<std::mutex> lock(opened_mqs_mtx_);
+      publish_msg_args = publish_core(this, topic_name_, id_, msg_virtual_address, opened_mqs_);
+    }
+
+    for (uint32_t i = 0; i < publish_msg_args.ret_released_num; i++) {
+      void * release_ptr = reinterpret_cast<void *>(publish_msg_args.ret_released_addrs[i]);
+      deleter(release_ptr);
+    }
```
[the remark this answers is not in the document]
bdm-k: I don't think that is really necessary.

src/agnocastlib/include/agnocast/agnocast_publisher.hpp
bdm-k: Updated in [6e721f6](https://github.com/autowarefoundation/agnocast/pull/1430/commits/6e721f68d283832f399d3451703c5f84dfd60fff)

bdm-k:
I added two unit tests:

- `AgnocastTypeErasedPublisherTest.test_cancel` verifies that `cancel_message()` correctly prevents a message from being published, decrements the borrowed publisher count, and invokes the provided custom deleter.
- `AgnocastSmartPointerTest.reset_void` is a death test that verifies the behavior when an `ipc_shared_ptr<void>` is dropped without being published or canceled.

I also mocked the `PublisherBase` destructor to suppress non-critical warnings.

src/agnocastlib/test/unit/test_mocked_agnocast.cpp
```diff
@@ -208,6 +210,50 @@ TEST_F(AgnocastPublisherTest, test_multiple_borrows_mixed_publish_and_drop)
...
+  };
+
+  // Act
+  dummy_publisher->cancel_message(std::move(message), deleter);
+
+  // Assert
+  EXPECT_EQ(publish_core_mock_called_count, 0);
+  EXPECT_EQ(agnocast_get_borrowed_publisher_num(), 0);
+  EXPECT_TRUE(deleter_invoked);
+}
```
ruth561, line 257: I checked the pointer value passed to the deleter using the test below. The test fails because nullptr is passed. Is this the intended behavior?

```diff
 TEST_F(AgnocastTypeErasedPublisherTest, test_cancel)
 {
   // Arrarnge
   agnocast::ipc_shared_ptr<void> message = dummy_publisher->borrow_loaned_message(4);
   bool deleter_invoked = false;
-  auto deleter = [&deleter_invoked](void * p) {
+  void *passed_ptr = nullptr;
+  auto deleter = [&](void * p) {
     deleter_invoked = true;
     passed_ptr = p;
     ::operator delete(p);
   };

   // Act
   dummy_publisher->cancel_message(std::move(message), deleter);

   // Assert
   EXPECT_EQ(publish_core_mock_called_count, 0);
   EXPECT_EQ(agnocast_get_borrowed_publisher_num(), 0);
   EXPECT_TRUE(deleter_invoked);
+  EXPECT_NE(passed_ptr, nullptr);
 }
```

src/agnocastlib/test/unit/test_mocked_agnocast.cpp
```diff
@@ -208,6 +210,50 @@ TEST_F(AgnocastPublisherTest, test_multiple_borrows_mixed_publish_and_drop)
...
+
+  void TearDown() override { rclcpp::shutdown(); }
+
+  std::shared_ptr<rclcpp::Node> node;
+  std::shared_ptr<agnocast::TypeErasedPublisher> dummy_publisher;
+};
+
+TEST_F(AgnocastTypeErasedPublisherTest, test_cancel)
+{
+  // Arrarnge
```
ruth561: typo
```suggestion
  // Arrange
```

src/agnocastlib/test/unit/test_mocked_agnocast.cpp
bdm-k: fixed in [514b06e](https://github.com/autowarefoundation/agnocast/pull/1430/commits/514b06edcede3e7ddc3d7d1dbc43f2c5881f5f25)

Thank you for pointing this out. I had misunderstood the semantics of `invalidate_all_references()`. Since `get()` always returns `nullptr` after invalidation, I fixed the issue by saving the message pointer before invalidation. I also updated the test to verify that the correct pointer is passed to the deleter, just like you did in the code block above.

src/agnocastlib/test/unit/test_mocked_agnocast.cpp
bdm-k: fixed in [514b06e](https://github.com/autowarefoundation/agnocast/pull/1430/commits/514b06edcede3e7ddc3d7d1dbc43f2c5881f5f25)

ruth561 (dismissed):

ruth561 (approved):

PR #1272, opened 2026-04-20
feat[needs minor version update]: replace POSIX MQ publish notification with eventfd

k1832:
## Description

Replace the POSIX MQ-based publish notification mechanism with kernel-managed eventfd.
The previous MQ path required the publisher to call `mq_send()` once per subscriber
in userspace (N+1 syscalls per publish including the PUBLISH ioctl). With eventfd,
the kernel module signals subscriber eventfds directly within `ioctl(PUBLISH)`,
reducing the cost to a single syscall regardless of subscriber count.

Key changes:
- Kernel: `subscriber_info` stores `eventfd_ctx`. `publish_msg` collects notify_ctx
  pointers under `topic_rwsem` write lock, then signals outside the lock to avoid
  contention. Stack buffer (64 entries) with heap fallback for larger subscriber counts.
- Userspace: Subscriber creates eventfd and passes to kernel via `ioctl(ADD_SUBSCRIBER)`.
  Publisher's `publish_core` is simplified from ~80 to ~15 lines (MQ send loop removed).
- Compatibility: `agnocast_eventfd_signal` wrapper for kernel < 6.8 API difference.

Detailed design, benchmark analysis, and alternative considered: see design doc (linked below).

## Benchmark Results

Measured with [isorc26_tmp](https://github.com/sykwer/isorc26_tmp),
10 topics, 100Hz, 3-iter mean of per-iter p50/p99, `--no-rt`. Test machine: 32 logical CPUs.

### Publish latency (publisher wall-clock, from `loan()` to `publish()` completion)

| Subs | MQ p50 / p99 | eventfd p50 / p99 | p50 diff |
|------|-------------|-------------------|----------|
| 4    | 11.5 / 37.1 us | 7.1 / 20.9 us | -38% |
| 8    | 17.0 / 51.3 us | 8.3 / 30.3 us | -51% |
| 16   | 24.5 / 61.6 us | 18.4 / 47.4 us | -25% |
| 32   | 56.9 / 133.5 us | 58.7 / 138.2 us | +3% |
| 64   | 186.4 / 404.2 us | 141.5 / 330.0 us | -24% |
| 128  | 674.2 / 1084.4 us | 594.4 / 1206.2 us | -12% |

### E2E latency (from publisher timestamp just before `loan()` to subscriber callback entry)

| Subs | MQ p50 / p99 | eventfd p50 / p99 | p50 diff |
|------|-------------|-------------------|----------|
[...]

PR #1432, opened 2026-07-02
feat(agnocast)[need-minor-update]: replace POSIX MQ publish notification with eventfd

k1832:
## Description

Replace the POSIX MQ publish notification with kernel-managed eventfd.

Publishing used to cost N+1 syscalls: the PUBLISH ioctl, plus one `mq_send()` per subscriber from userspace. The kernel now signals the subscribers' eventfds inside `ioctl(PUBLISH)`, so it is one syscall regardless of fan-out.

- **kmod**: `subscriber_info` holds an `eventfd_ctx *`, acquired in `ADD_SUBSCRIBER` and released on every path that drops a subscriber. `publish_msg` collects the contexts to signal under `topic_rwsem` — applying the same take / domain / local-publication filters it used to apply when building the subscriber-ID list — then signals them after dropping the lock, so a woken subscriber's `RECEIVE_MSG` is not blocked by the publisher.
- **userspace**: subscribers create an `EFD_NONBLOCK | EFD_CLOEXEC` eventfd, pass it via `ADD_SUBSCRIBER`, and epoll it instead of a per-subscription MQ. `publish_core` drops the `mq_send` loop and the `opened_mqs` cache.
- **ABI**: `ioctl_add_subscriber_args` gains `eventfd`; `ioctl_publish_msg_args` drops the subscriber-ID buffer → `need-minor-update`.

**Known issue**: collecting the contexts costs one `kmalloc_array()` per publish, taken while holding `topic_rwsem`. That does not belong on the hot path. **#1504 removes it** by precomputing each publisher's list when endpoints register, leaving publish with nothing to allocate and nothing to walk. The naive version is kept here so the mechanism swap can be reviewed on its own; #1503 then deletes the MQ machinery both make dead.

## Benchmark Results

`agnocast_bench` (#1451), MQ baseline vs eventfd. The baseline is this branch's merge-base with `main` (`528fd011`), so the two arms differ only by the notification mechanism, and both carry byte-identical instrumentation built from the same out-of-tree probe. **Measured on this head, i.e. WITHOUT #1504** — the per-publish `kmalloc_array()` is still in place, so the eventfd arm here is the pessimistic case.

Conditions: 100 Hz requested, 2 s warmup + 30 s measurement (3000 publish samples per cell), QoS depth 10, `performance` governor on every CPU, publisher and subscribers pinned to the host's E-cores (publisher `cpu16`, subscribers round-robin over `cpu17-31`). Single trial per cell. Negative Δ = eventfd faster.

### Publish latency — publisher wall-clock around `publish()` (µs)

| subs | MQ p50 / p99 | eventfd p50 / p99 | p50 Δ | p99 Δ |
|-----:|--------------|-------------------|:-----:|:-----:|
| 1  | 23.1 / 38.1  | 11.5 / 23.6  | **−50%** | −38% |
| 2  | 24.6 / 40.1  | 12.2 / 26.5  | **−50%** | −34% |
| 4  | 28.3 / 49.3  | 14.3 / 28.2  | **−50%** | −43% |
| 8  | 36.0 / 64.5  | 18.1 / 32.9  | **−50%** | −49% |
| 16 | 48.6 / 77.1  | 26.4 / 47.5  | **−46%** | −38% |
| 32 | 74.5 / 114.5 | 40.1 / 72.8  | **−46%** | −36% |

### E2E latency — publish timestamp → callback entry (µs)

| subs | MQ p50 / p99 | eventfd p50 / p99 | p50 Δ | p99 Δ |
|-----:|--------------|-------------------|:-----:|:-----:|
| 1  | 64.4 / 133.6  | 69.4 / 103.2  | +8% | −23% |
| 2  | 65.0 / 119.8  | 68.4 / 102.7  | +5% | −14% |
| 4  | 87.6 / 138.2  | 87.6 / 131.6  | ±0% | −5% |
| 8  | 99.3 / 157.5  | 96.8 / 150.5  | −2% | −4% |
| 16 | 114.0 / 219.3 | 107.4 / 237.5 | −6% | +8% |
| 32 | 126.3 / 940.8 | 119.2 / 887.8 | −6% | −6% |

Publish drops by roughly half at every fan-out. Both arms still grow with N — the eventfd arm because this head walks the subscriber table and allocates once per publish; #1504 removes that.

E2E shows no meaningful difference either way. It seems like it is dominated by scheduler wakeup latency, which swamps the notification cost. The p99 column reflects callback batching in `receive_and_execute_message()` — one `RECEIVE_MSG` yields several entries and each callback after the first is timestamped only once its predecessors have run — so it is not a comparison of notification mechanisms either.

The case for this change is therefore the publish path, not e2e: publish stops scaling with subscriber count, and the MQ machinery becomes deletable (#1503). All figures will be re-measured with #1504 applied.

## Related links

- Supersedes #1272
- Follow-up removing the per-publish allocation: #1504
- Follow-up removing the dead MQ machinery: #1503
- Measurement harness + methodology: #1451
- [TIER IV Internal Design Document](https://tier4.atlassian.net/wiki/spaces/CRL/pages/5132517589/Publish+eventfd)

## How was this PR tested?

- [ ] Autoware
- [x] `bash scripts/test/e2e_test_domain_bridge.bash`
- [ ] sample application

The benchmark figures above were re-measured on this head, so they no longer depend on #1504. The `scripts/test/*` e2e runs still predate the #1504 split and need re-running before merge. KUnit is green on this head.

## Notes for reviewers

Userspace gtests green: unit (292), tf2 broadcasters (18), tf2 transform listener (6), and the executor integration test (35) — the last drives the executor end-to-end over eventfd (epoll on the fd, a timer writes to it, the callback fires) with a mocked ioctl.

The `kmalloc_array()` is placed before anything is published, so an `-ENOMEM` cannot report failure for an entry that was already inserted, and every error path reaches the signal loop with nothing collected.

#1443's rename grouping and direction filter carry over unchanged: eventfd signals each subscriber's own fd off the shared `sub_info_htable` under `domain_delivery_allowed`, which is name-agnostic. The domain-bridge e2e run confirmed it (domain-1 publisher → domain-2 subscriber, 128/128 received).

## Post-merge checklist

`need-minor-update` — kmod/agnocastlib ABI change (also flagged in the title).


Koichi98 (dismissed):

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
```diff
@@ -250,15 +248,15 @@ class Subscription : public SubscriptionBase
...
+    // fd reuse conflicts. When the eventfd is closed below, the OS may later reuse the same fd
+    // number for a new subscription. If the old entry remains in id2_callback_info, adding the new
+    // fd to epoll (EPOLL_CTL_ADD) can fail with EEXIST because epoll still associates that fd
+    // number with the stale entry.
     {
       std::lock_guard<std::mutex> lock(id2_callback_info_mtx);
       id2_callback_info.erase(callback_info_id_);
     }
-    remove_mq(mq_subscription_);
+    close_notify_eventfd(notify_eventfd_);
```
atsushi421: The notify eventfd is created in SubscriptionBase::initialize() and stored in a base member, but it is closed only here in ~Subscription(). If the derived constructor throws after init_base() succeeds (register_callback can throw: allocate_callback_info_id() on callback-ID overflow, or bad_alloc on the map insertion), ~Subscription() never runs and the fd leaks, and any future non-take subclass of SubscriptionBase would leak it silently as well. Consider moving close_notify_eventfd(notify_eventfd_) into ~SubscriptionBase(): base destructors run after the derived destructor body, so the erase-before-close ordering documented above is preserved, and the -1 guard keeps TakeSubscription unaffected.

agnocast_kmod/agnocast_kunit/agnocast_kunit_publish_msg.c
```diff
@@ -58,41 +59,67 @@ static void setup_one_publisher(
   KUNIT_ASSERT_EQ(test, ret2, 0);
 }
 
-static void setup_pub_sub_same_process(
-  struct kunit * test, topic_local_id_t * publisher_id, topic_local_id_t * subscriber_id,
-  bool ignore_local_publications, uint64_t * ret_addr)
+// The setup helpers above pass -1, which leaves notify_ctx NULL and so keeps the subscriber out of
```
atsushi421: This comment states an invariant that does not hold: in the KUnit build, agnocast_eventfd_get(-1) returns the non-NULL unobserved_slot sentinel, so these subscribers are stored with a non-NULL notify_ctx and are actually collected and signaled on publish, just invisibly to assertions (and in a production build, -1 makes ADD_SUBSCRIBER fail outright for a non-take subscriber, so no such subscriber exists at all). Consider rewording this comment (and the same one in agnocast_kunit_domain_bridge.c) to match the fake's actual semantics, e.g. "-1 maps to an unobserved fake slot; the subscriber is still signaled, but no assertion can see it".

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -2644,36 +2666,9 @@ static long publish_msg_cmd(union ioctl_publish_msg_args __user * arg)
...
-      if (copy_to_user(
-            (topic_local_id_t __user *)subscriber_ids_buffer_addr, subscriber_ids_buf,
-            copy_count * sizeof(topic_local_id_t))) {
-        kfree(subscriber_ids_buf);
-        return -EFAULT;
-      }
-    }
-  }
-  kfree(subscriber_ids_buf);
+    &publish_msg_args);
```
atsushi421, line 2671: agnocast_ioctl_publish_msg now inserts the entry and signals all subscriber eventfds before the final copy_to_user, so when that copy fails the ioctl returns -EFAULT and publish_core exits, yet subscribers were already woken and will consume the message. Under the old design the notification was sent from userspace only after the ioctl succeeded, so a failed copy meant no one had been notified. Consider either documenting that -EFAULT from PUBLISH means the publish itself already happened, or moving the signal loop after the successful copy_to_user.

atsushi421 (commented):
docs/message_queue.md still describes the POSIX MQ publish notification as the current mechanism (the MQ naming scheme, the MqMsgAgnocast protocol, subscribers opening an MQ in create_subscription), and the README's message-queue cleanup / fs.mqueue.queues_max sections become stale as well once this merges. Consider updating them to describe the eventfd-based path, or explicitly folding the in-repo docs update into #1503; the post-merge checklist currently only mentions the external agnocast_doc.

Koichi98:
@atsushi421 
I'll fold the in-repo docs update into #1503, so the mqueue machinery and the docs describing it are deleted together. Added to this PR's Related links.

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
Koichi98: Thank you for pointing out!
Fixed in https://github.com/autowarefoundation/agnocast/pull/1432/commits/99e4402674a96694b5d4f689452fae6401ceeaca.

agnocast_kmod/agnocast_kunit/agnocast_kunit_publish_msg.c
Koichi98: Fixed comments in https://github.com/autowarefoundation/agnocast/pull/1432/commits/a6b510bcc4032963545ffbd8746cc99f1c9f9e17.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Added comments in https://github.com/autowarefoundation/agnocast/pull/1432/commits/f0406f732523338ac423b5cd52b7605058440097.

atsushi421 (approved):

k1832:
👀🙌

PR #1433, opened 2026-07-02
refactor(agnocastlib): replace bridge MQ with UDS

ruth561:
## Description

This PR replaces the POSIX message queue (MQ) used by the bridge with **an abstract-namespace UNIX Domain Socket (UDS)**.

Previously, the bridge manager used POSIX MQ to interact with Agnocast endpoints and the discovery agent. However, a POSIX MQ is not owned by any particular process, so it is not cleaned up by the OS on process exit. This leads to an MQ leak when the bridge manager crashes unexpectedly, which has been a serious pain point for Agnocast users developing Agnocast applications.

An abstract-namespace UNIX domain socket, on the other hand, is bound to a specific process and released by the OS automatically when that process exits.

Main changes:

* POSIX MQ `/dev/mqueue/agnocast_bridge_manager@-1[_d<domain ID>]` -> Abstract-namespace UDS **`\0agnocast_bridge_manager_<IPC-namespace-inode>[_d<domain ID>]`**
* Renamed "MQ" to "UDS" in the corresponding filenames, data structures, comments, etc.
* Added tests for the newly introduced UDS helpers.
* Updated the related documentation.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

### Namespace scope

We include the IPC namespace inode in the UDS name because the bridge manager runs one instance per IPC namespace.

During testing, we often use `unshare --ipc` to split only the IPC namespace, leaving the network namespace shared. If we used a fixed name like `\0agnocast_bridge_manager`, a second instance would fail to bind with an "address already in use" error.

If applications always split both namespaces together, the fixed name would be fine. However, to make testing easier and avoid conflicts, I chose to add the inode to the UDS name.

## Post-merge checklist

**Important notes:**

### DIFF
--- a/src/agnocastlib/include/agnocast/bridge/agnocast_bridge_ipc_event_loop_base.hpp
+++ b/src/agnocastlib/include/agnocast/bridge/agnocast_bridge_ipc_event_loop_base.hpp
@@ -1,13 +1,12 @@
 #pragma once
 
-#include "agnocast/agnocast_mq.hpp"
 #include "agnocast/agnocast_utils.hpp"
+#include "agnocast/bridge/agnocast_bridge_uds.hpp"
 
 #include <rclcpp/logger.hpp>
 #include <rclcpp/logging.hpp>
 
 #include <fcntl.h>
-#include <mqueue.h>
 #include <sys/epoll.h>
 #include <sys/signalfd.h>
 #include <sys/socket.h>
@@ -17,6 +16,8 @@
 #include <array>
 #include <cerrno>
 #include <csignal>
+#include <cstddef>
+#include <cstdint>
 #include <cstring>
 #include <functional>
 #include <string>
@@ -30,12 +31,12 @@ namespace agnocast
 class IpcEventLoopBase
 {
 public:
-  using EventCallback = std::function<void(int)>;
+  using MessageCallback = std::function<void(const void * data, std::size_t size)>;
   using SignalCallback = std::function<void()>;
   using SocketCallback = std::function<std::string()>;
 
   IpcEventLoopBase(
-    const rclcpp::Logger & logger, const std::string & mq_name, long mq_msg_size,
+    const rclcpp::Logger & logger, const std::string & uds_addr, long max_msg_size,
     const std::vector<int> & signals_to_block, const std::vector<int> & signals_to_ignore);
 
   virtual ~IpcEventLoopBase();
@@ -45,11 +46,11 @@ public:
 
   bool spin_once(int timeout_ms);
 
-  void set_mq_handler(EventCallback cb);
+  void set_message_handler(MessageCallback cb);
   void set_signal_handler(SignalCallback cb);
   void set_socket_handler(SocketCallback cb);
 
-  const std::string & get_mq_name() const { return mq_name_; }
+  const std::string & get_uds_addr() const { return uds_addr_; }
 
 protected:
   rclcpp::Logger logger_;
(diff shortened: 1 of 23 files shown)


src/agnocastlib/include/agnocast/bridge/agnocast_bridge_ipc_event_loop_base.hpp
```diff
@@ -218,9 +216,26 @@ inline void IpcEventLoopBase::set_socket_handler(SocketCallback cb)
   socket_cb_ = std::move(cb);
 }
 
-inline void IpcEventLoopBase::setup_mq()
+inline void IpcEventLoopBase::drain_listener()
 {
-  mq_fd_ = create_and_open_mq(mq_name_);
+  std::vector<uint8_t> buf(static_cast<size_t>(max_msg_size_));
+  while (true) {
+    ssize_t n = recv(listener_fd_, buf.data(), buf.size(), 0);
```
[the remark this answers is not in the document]
ruth561: Fixed in [d7a43a3](https://github.com/autowarefoundation/agnocast/pull/1433/commits/d7a43a3474312a2075ef9d25c18b0ab3f55e47dc).

src/agnocastlib/test/unit/test_daemon_bridge_uds.cpp
```diff
@@ -77,13 +77,40 @@ TEST(DaemonBridgeMqTest, BridgeMsgWireLayout)
 }
 
 // An empty ROS_DOMAIN_ID (set but "") means "no domain": no `_d` suffix.
-TEST(DaemonBridgeMqTest, BridgeMqNameEmptyDomainIdHasNoSuffix)
+TEST(DaemonBridgeMqTest, BridgeUdsAddrEmptyDomainIdHasNoSuffix)
```
Koichi98: Better to rename these tests to `DeamonBridgeUdsTest`?

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_uds.hpp
```diff
@@ -0,0 +1,125 @@
...
+  const std::string display_name = (!addr.empty() && addr.front() == '\0') ? addr.substr(1) : addr;
+
+  ssize_t send_result = -1;
+  int last_errno = 0;
+  for (int retry = 0; retry <= BRIDGE_UDS_SEND_MAX_RETRIES; ++retry) {
+    send_result = sendto(fd, data, size, MSG_NOSIGNAL, reinterpret_cast<sockaddr *>(&sa), alen);
+    if (send_result >= 0) break;
+    last_errno = errno;
+    if (
+      last_errno != ECONNREFUSED && last_errno != ENOENT && last_errno != EAGAIN &&
```
Koichi98: Do we need `ENOENT`, although we now use abstract uds?

Koichi98:
@ruth561 
The PR look good overall. Please just take a look at two nit comments above.

src/agnocastlib/test/unit/test_daemon_bridge_uds.cpp
Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1433/commits/13b5b8cff544e974ec2fc643f36eea94873692be.

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_uds.hpp
Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1433/commits/13b5b8cff544e974ec2fc643f36eea94873692be.

Koichi98 (approved):

PR #1435, opened 2026-07-06
feat(agnocast)[needs minor version update]: support topic rename (remap) in the domain bridge

k1832:
## What

Adds **topic rename (remap)** to the domain bridge. The bridge previously required identical topic names on both domains; redundant-Autoware setups need renaming (e.g. prefixing a sub-ECU topic so it can be logged under the main domain without a name collision). The rename target is the standard ROS 2 `domain_bridge` `remap:` field — previously parsed and dropped, now honored.

Stacked on #1418 (base branch `feat/domain-bridge-userspace`); reviewing just this PR's diff shows only the rename change.

## Design

The merged `topic_struct`, `domain_delivery_allowed`, and the refcnt lifetime are **name-agnostic and untouched**. The change is confined to rule matching:

- `domain_bridge_rule` stores a name **per domain** (`topic_name_a` / `topic_name_b`; equal when there's no rename).
- rules are keyed by the **cell `(name, domain)`**; `find_grouped_topic_struct` resolves the partner by the **partner's** name.
- **Invariant enforced in `add_domain_bridge`:** each cell belongs to at most one rule, and a rule pairs exactly two cells. `r_a == r_b` (non-NULL) → re-declaration/reverse direction (OR the flags); any other overlap (e.g. **two sources remapped onto one target**, or a 3-cell chain) → `-EBUSY`. The "rule before any endpoint" guard applies to both names.
- Same-name bridging is the special case `name_a == name_b`, so existing a2a and same-name bridge behavior is **byte-identical**.

Userspace: the config parser reads `remap` → `(from_topic, to_topic, from_domain, to_domain)`; the wrapper symbol and the `ADD_DOMAIN_BRIDGE` ioctl carry both names.

Cross-ECU rename already works via the external ROS 2 `domain_bridge` (a2r/r2a are name-local); this PR adds the same-IPC-namespace **zero-copy (Case 2)** side.

### 1. kmod KUnit — rename logic incl. corner cases (automated)

The suite (`agnocast_kunit_domain_bridge.c`) gains 4 rename cases. Runs in the `kunit` CI job (QEMU); locally on a `CONFIG_KUNIT=y` kernel:

```bash
bash scripts/test/run_kunit.bash
```

Ran the CI-equivalent (`kunit.py --arch=x86_64`) locally: **185/185 pass**, including:
- `test_case_domain_bridge_rename_groups_wrappers` — two differently-named cells share one `topic_struct` (found from either name, partner by partner name).
- `test_case_domain_bridge_rename_cross_domain_delivery` — `/src@1` publish reaches the `/dst@2` subscriber.
- `test_case_domain_bridge_rename_fanout_rejected` — two sources onto one target → `-EBUSY`.
- `test_case_domain_bridge_rename_multi_publisher` — extra publisher on the source + a native publisher on the renamed target still deliver exactly once (the "the bridged node isn't the only publisher" case).

All pre-existing domain-bridge tests still pass (same-name backward compat).

### 2. userspace config → ioctl plumbing (automated)

```bash
cd src/ros2agnocast_discovery_agent && python3 -m pytest test/
```

Covers `remap` parsing (`test_remap_sets_the_target_topic_name`) and the tool passing both names to the ioctl wrapper.

### 3. full-stack cross-domain renamed delivery — Case 2 zero-copy (manual)

On a **freshly loaded** kmod (the rule must precede any endpoint for the topic):

```bash
bash scripts/dev/build_all.bash
sudo rmmod agnocast 2>/dev/null; sudo insmod agnocast_kmod/agnocast.ko

# Register a RENAME rule: /my_topic@1  ->  /renamed@2
cat > /tmp/bridge.yaml <<'YAML'
from_domain: 1
to_domain: 2
topics:
  "/my_topic":
    remap: "/renamed"
YAML
ros2 run ros2agnocast_discovery_agent register_domain_bridge --config /tmp/bridge.yaml
#   expect: "registered: /my_topic@1 -> /renamed@2"
```

This alone exercises the whole new userspace path (YAML `remap` → parser → tool → wrapper → `ADD_DOMAIN_BRIDGE` ioctl → kmod rule with two names). For **live delivery**, publish on the source name in domain 1 and subscribe the renamed name in domain 2 — the stock samples both use `/my_topic`, so remap the listener's subscription to `/renamed` (Agnocast resolves topic names through the node, so a ROS remap `/my_topic:=/renamed` on the listener node applies):

```bash
# domain-2 listener, its /my_topic subscription remapped to /renamed:
AGNOCAST_BRIDGE_MODE=off ROS_DOMAIN_ID=2 \
  ros2 run agnocast_sample_application <listener_node> --ros-args -r /my_topic:=/renamed &
sleep 5
# domain-1 talker on the source name /my_topic (stock):
AGNOCAST_BRIDGE_MODE=off ROS_DOMAIN_ID=1 \
  ros2 launch agnocast_sample_application talker.launch.xml
```

Expected: the domain-2 listener receives the domain-1 talker's messages under the renamed topic — cross-domain delivery **with rename**, zero-copy, no bridge node. (The renamed cross-domain delivery itself is also proven deterministically by the KUnit case in method 1.)

Regression — the existing same-name path still works:

```bash
bash scripts/test/e2e_test_domain_bridge.bash
```

## Notes

- `[needs minor version update]`: extends the `ADD_DOMAIN_BRIDGE` ioctl struct (a second name). The bump lives with the domain-bridge stack; labeled `need-minor-update`.
- Service domain bridge is still out of scope (see the design doc): services are request/response topics, so the rule mechanism would extend, but full support is gated on the unimplemented A2R service bridge (Case 8) and dynamic rules for the per-client response topics (Case 2).

### DIFF
--- a/.github/workflows/checkpatch.yaml
+++ b/.github/workflows/checkpatch.yaml
@@ -45,6 +45,9 @@ jobs:
           # COMPLEX_MACRO: the KUnit suite's TEST_CASES_* macros are comma-separated
           # KUNIT_CASE(...) lists that cannot be parenthesized; this is a false positive.
           IGNORE="$IGNORE,COMPLEX_MACRO"
+          # FILE_PATH_CHANGES: agnocast is an out-of-tree module with no entry in the
+          # kernel MAINTAINERS file, so adding or moving kmod files never needs one.
+          IGNORE="$IGNORE,FILE_PATH_CHANGES"
 
           git diff origin/${{ github.base_ref }}...HEAD -- agnocast_kmod/ \
             | ./checkpatch.pl --no-tree --show-types --ignore="$IGNORE" -
--- a/README.md
+++ b/README.md
@@ -200,7 +200,6 @@ rm /dev/shm/agnocast@*
 
 # Remove leftover message queues
 rm /dev/mqueue/agnocast@*
-rm /dev/mqueue/agnocast_bridge_manager@*
 ```
 
 If you encounter `mq_open failed: No space left on device`, the system has reached the maximum number of message queues. Run the cleanup commands above, and if the error persists, increase the system-wide `queues_max` limit (e.g., `sudo sysctl -w fs.mqueue.queues_max=1024`). See [System Configuration](https://autowarefoundation.github.io/agnocast_doc/environment-setup/configuration/) for details.
--- a/agnocast_kmod/agnocast.h
+++ b/agnocast_kmod/agnocast.h
@@ -280,7 +280,9 @@ struct ioctl_set_ros2_publisher_num_args
 
 struct ioctl_add_domain_bridge_args
 {
-  struct name_info topic_name;
+  // topic_name_from / topic_name_to may differ (rename); equal for a plain bridge.
+  struct name_info topic_name_from;
+  struct name_info topic_name_to;
   uint32_t from_domain;
   uint32_t to_domain;
 };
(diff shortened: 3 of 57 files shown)


k1832:
Split into two stacked PRs by layer: kmod #1436 and userspace #1437 (userspace based on kmod).

PR #1436, opened 2026-07-06
feat(agnocast_kmod)[needs minor version update]: support cross-domain topic rename in the domain bridge

k1832:
## Description

Adds cross-domain **topic rename** to the domain bridge (kmod side). The bridge required identical topic names on both domains; this lets a rule map a source topic in one domain to a differently-named topic in the other (the ROS 2 `domain_bridge` `remap` target) — e.g. prefixing a sub-ECU topic so it can be logged under the main domain without a name collision.

The merged `topic_struct`, `domain_delivery_allowed`, and the refcnt lifetime are **name-agnostic and untouched**. The change is confined to rule matching:

- `domain_bridge_rule` stores a name **per domain** (`topic_name_a` / `topic_name_b`; equal when there's no rename).
- Rules are keyed by the **cell `(name, domain)`**; `find_grouped_topic_struct` resolves the partner by the **partner's** name.
- **Invariant in `add_domain_bridge`:** each cell belongs to at most one rule, and a rule pairs exactly two cells. `r_a == r_b` (non-NULL) → re-declaration / reverse direction; any other overlap — e.g. two sources remapped onto one target, or a 3-cell chain — → `-EBUSY`. The "rule before any endpoint" guard applies to both names.
- Same-name bridging is the special case `name_a == name_b`, so existing a2a and same-name bridge behavior is **byte-identical**.
- The `ADD_DOMAIN_BRIDGE` ioctl struct now carries a topic name per domain.

Base of the rename stack: **#1436 (this)** → #1443 (canonical notification MQ) → #1437 (userspace YAML `remap` → wrapper → ioctl).

## Related links

- Prior (merged) domain-bridge stack: #1417 (kmod) → #1418 (userspace).
- Rename stack (bottom-up): **#1436 (this)** → #1443 (canonical notification MQ) → #1437 (userspace `remap`).

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

KUnit: ran the full suite via the `kunit` CI path (`kunit.py --arch=x86_64` under QEMU) — **185/185 pass**, including 4 new rename cases:

- `test_case_domain_bridge_rename_groups_wrappers` — two differently-named cells share one `topic_struct`.
- `test_case_domain_bridge_rename_cross_domain_delivery` — `/src@1` publish reaches the `/dst@2` subscriber.
- `test_case_domain_bridge_rename_fanout_rejected` — two sources onto one target → `-EBUSY`.
- `test_case_domain_bridge_rename_multi_publisher` — an extra publisher on the source + a native publisher on the renamed target still deliver exactly once.

All pre-existing domain-bridge tests still pass (same-name backward compat). The `(required)` e2e / `run_requires_kernel_module` boxes are left for the run-build-test pass.

## Notes for reviewers

- kmod half of the rename, split from the (now-closed) single PR. The core sharing/lifetime code is unchanged; the diff is the rule table + matching + KUnit.
- The one-pair-per-cell invariant (`r_a == r_b` in `add_domain_bridge`) handles all the conflict cases uniformly — the main thing to sanity-check.

## Post-merge checklist

- [ ] Reflect the changes in [`agnocast_doc`](https://github.com/autowarefoundation/agnocast_doc) after merging (the domain-bridge guide should note rename support)

## Version Update Label

`need-minor-update` — kmod internal ABI change (the `ADD_DOMAIN_BRIDGE` ioctl struct gains a per-domain name); reflected in the PR title.

### DIFF
--- a/agnocast_kmod/agnocast.h
+++ b/agnocast_kmod/agnocast.h
@@ -310,7 +310,9 @@ struct ioctl_set_ros2_publisher_num_args
 
 struct ioctl_add_domain_bridge_args
 {
-  struct name_info topic_name;
+  // topic_name_from / topic_name_to may differ (rename); equal for a plain bridge.
+  struct name_info topic_name_from;
+  struct name_info topic_name_to;
   uint32_t from_domain;
   uint32_t to_domain;
 };
@@ -486,8 +488,8 @@ int agnocast_ioctl_remove_bridge(
   const char * topic_name, const pid_t pid, bool is_r2a, const struct ipc_namespace * ipc_ns);
 
 int agnocast_ioctl_add_domain_bridge(
-  const char * topic_name, uint32_t from_domain, uint32_t to_domain,
-  const struct ipc_namespace * ipc_ns);
+  const char * topic_name_from, const char * topic_name_to, uint32_t from_domain,
+  uint32_t to_domain, const struct ipc_namespace * ipc_ns);
 
 int agnocast_ioctl_get_version(struct ioctl_get_version_args * ioctl_ret);
 
@@ -574,8 +576,8 @@ bool agnocast_is_in_topic_htable(const char * topic_name, const struct ipc_names
 bool agnocast_is_in_bridge_htable(const char * topic_name, const struct ipc_namespace * ipc_ns);
 pid_t agnocast_get_bridge_owner_pid(const char * topic_name, const struct ipc_namespace * ipc_ns);
 bool agnocast_get_domain_rule(
-  const char * topic_name, const struct ipc_namespace * ipc_ns, uint32_t * domain_a,
-  uint32_t * domain_b, bool * a_to_b, bool * b_to_a);
+  const char * topic_name, const struct ipc_namespace * ipc_ns, uint32_t domain,
+  uint32_t * domain_a, uint32_t * domain_b, bool * a_to_b, bool * b_to_a);
 // Returns the shared topic_struct's wrapper refcount for the wrapper in domain_id,
 // or 0 if no such wrapper exists. Used to observe domain-bridge grouping.
 int agnocast_topic_wrapper_refcnt(
--- a/agnocast_kmod/agnocast_ioctl.c
+++ b/agnocast_kmod/agnocast_ioctl.c
@@ -75,26 +75,31 @@ static struct topic_wrapper * find_topic_for_current(
 }
 
 static struct domain_bridge_rule * find_domain_rule(
-  const char * topic_name, const struct ipc_namespace * ipc_ns);
+  const char * topic_name, const struct ipc_namespace * ipc_ns, uint32_t domain_id);
 
-// If a domain bridge rule pairs this domain with another whose wrapper already
-// exists, return that partner's topic_struct so the new wrapper can share it.
+// If a domain bridge rule pairs this cell (topic_name, domain_id) with another
+// whose wrapper already exists, return that partner's topic_struct so the new
+// wrapper can share it. The partner may use a different name (rename), so look it
+// up by the rule's name for the partner domain.
 static struct topic_struct * find_grouped_topic_struct(
   const char * topic_name, const struct ipc_namespace * ipc_ns, uint32_t domain_id)
 {
-  const struct domain_bridge_rule * rule = find_domain_rule(topic_name, ipc_ns);
+  const struct domain_bridge_rule * rule = find_domain_rule(topic_name, ipc_ns, domain_id);
   if (!rule) return NULL;
 
   uint32_t partner_domain;
+  const char * partner_name;
   if (domain_id == rule->domain_a) {
     partner_domain = rule->domain_b;
+    partner_name = rule->topic_name_b;
   } else if (domain_id == rule->domain_b) {
     partner_domain = rule->domain_a;
+    partner_name = rule->topic_name_a;
   } else {
     return NULL;
   }
 
-  struct topic_wrapper * partner = find_topic(topic_name, ipc_ns, partner_domain);
+  struct topic_wrapper * partner = find_topic(partner_name, ipc_ns, partner_domain);
   return partner ? partner->topic : NULL;
 }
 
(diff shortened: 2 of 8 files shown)


agnocast_kmod/agnocast.h
```diff
@@ -310,7 +310,9 @@ struct ioctl_set_ros2_publisher_num_args
 
 struct ioctl_add_domain_bridge_args
 {
-  struct name_info topic_name;
+  // topic_name_from / topic_name_to may differ (rename); equal for a plain bridge.
+  struct name_info topic_name_from;
+  struct name_info topic_name_to;
   uint32_t from_domain;
```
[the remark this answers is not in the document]
k1832: By design — this is a stacked pair. #1437 (stacked on this PR) updates the in-repo ioctl-wrapper mirror (`ioctl_add_domain_bridge_args` and `add_agnocast_domain_bridge_rule`) to carry both `topic_name_from` and `topic_name_to`, so the userspace side matches the new struct size. They merge together — same kmod→wrapper split as #1426 → #1427 — and #1436 isn't merged on its own, so there's no window where the wrapper is out of sync.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -2167,57 +2172,66 @@ int agnocast_ioctl_add_bridge(
...
-  hash_for_each_possible(domain_rule_htable, rule, node, hash_val)
+  int bkt;
+  hash_for_each(domain_rule_htable, bkt, rule, node)
   {
-    if (ipc_ns == rule->ipc_ns && strcmp(rule->topic_name, topic_name) == 0) {
+    if (ipc_ns != rule->ipc_ns) continue;
+    if (
+      (domain == rule->domain_a && strcmp(rule->topic_name_a, topic_name) == 0) ||
+      (domain == rule->domain_b && strcmp(rule->topic_name_b, topic_name) == 0)) {
       return rule;
```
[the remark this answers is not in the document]
k1832: Intentional. A rule is keyed by the cell `(name, domain)`, and a topic can be either side of a bridged pair, so a single-name hash bucket can't cover both — hence the full scan (noted in the comment above `find_domain_rule`). Rules are few (one per bridged pair) and the buckets are almost all empty, so a lookup is a handful of pointer checks. If the rule count ever grows enough to matter, a keyed/dual-name structure would be the fix.

agnocast_kmod/agnocast.h
```diff
@@ -310,7 +310,9 @@ struct ioctl_set_ros2_publisher_num_args
 
 struct ioctl_add_domain_bridge_args
 {
-  struct name_info topic_name;
+  // topic_name_from / topic_name_to may differ (rename); equal for a plain bridge.
+  struct name_info topic_name_from;
+  struct name_info topic_name_to;
```
Koichi98, line 315: Let's also update `src/agnocast_ioctl_wrapper/src/agnocast_ioctl.hpp` to avoid ABI being destroyed.

agnocast_kmod/agnocast.h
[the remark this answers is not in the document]
Koichi98: https://github.com/autowarefoundation/agnocast/pull/1436/changes#r3555519363

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -2167,57 +2172,66 @@ int agnocast_ioctl_add_bridge(
   return ret;
 }
 
+// A rule is keyed by cell = (name, domain). Rules are few (one per bridged topic
+// pair), so a full scan is cheaper than maintaining a dual-name hash.
 static struct domain_bridge_rule * find_domain_rule(
-  const char * topic_name, const struct ipc_namespace * ipc_ns)
+  const char * topic_name, const struct ipc_namespace * ipc_ns, uint32_t domain)
```
Koichi98: Might be better to rename this to `domain_id`? Just thought it should be better to standardizing on either `domain_id` or `domain`.

agnocast_kmod/agnocast_kunit/agnocast_kunit_domain_bridge.h
```diff
@@ -2,21 +2,25 @@
 #pragma once
 #include <kunit/test.h>
```
Koichi98, line 4: This is something I should have flagged back in #1417, but: going forward I'd like to move to one kunit file per ioctl function, rather than feature-grouped files like `agnocast_kunit_domain_bridge.c` (which currently covers `add_domain_bridge` plus its interaction with several other ioctls).

The point of one-file-per-ioctl is to be able to look at a single file and tell whether all of that ioctl's code paths are covered. That gets harder once tests for multiple ioctls are interleaved in one file.

I'll leave it on you whether to address in this PR or in a separate PR, but I'd really like this to be fixed.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -2167,57 +2172,66 @@ int agnocast_ioctl_add_bridge(
...
+  // these two cells -- a re-declaration or the reverse direction, so just OR in the
+  // direction. Any other overlap (a cell already paired with a different cell) is a
+  // fan-out and is rejected. This is the one place that enforces one pair per cell.
+  // TODO: support >2 domains per topic by storing a domain group instead of a fixed pair.
+  struct domain_bridge_rule * r_a = find_domain_rule(name_a, ipc_ns, domain_a);
+  struct domain_bridge_rule * r_b = find_domain_rule(name_b, ipc_ns, domain_b);
+  if (r_a || r_b) {
+    if (r_a == r_b) {
+      r_a->a_to_b |= from_is_a;
+      r_a->b_to_a |= !from_is_a;
```
Koichi98, line 2225: Seems like the current tests missing the one that covers this code path.

agnocast_kmod/agnocast.h
k1832: Done!

agnocast_kmod/agnocast_ioctl.c
k1832: Good point. Done

agnocast_kmod/agnocast_ioctl.c
k1832: Done

agnocast_kmod/agnocast_kunit/agnocast_kunit_domain_bridge.h
k1832: Sorry for missing that consistency... Let's do it in a follow-up PR.

k1832 (commented):
Resolving all the comments with one point that will need to be followed-up (one kunit file per ioctl)

Koichi98 (approved):

PR #1438, opened 2026-07-06
refactor(agnocastlib): add void specialization of Subscription

bdm-k:
## Description

This PR adds a `void` specialization of `Subscription` and refactors `GenericSubscription` to use it. The only difference in the `void` specialization is its constructors. While standard (old) constructors do not take the type name of the message, the `void` specialization's constructors do. `GenericSubscription` now inherits from `Subscription<void>` instead of `SubscriptionBase`. The wrapped callback technique is leveraged here, similar to the Agnocast service implementation.

Overall, this PR is the subscription counterpart of #1430. However, instead of defining `TypeErasedSubscription`, it adds `void` specialization of `Subscription` because `TypeErasedSubscription` would have shared most of its code with `Subscription`.

The type-erased generic callback machinery has been removed. `agnocast_callback_info` is now largely the same as it was previously. The unit tests for the generic callback machinery (`test_agnocast_generic_callback`) have also been removed.

## How was this PR tested?

- [ ] Autoware
- [x] sample application

## Notes for reviewers

- We might want to address the inconsistency between `TypeErasedPublisher` and `Subscription<void>`. I'd appreciate your thoughts on this.

## Post-merge checklist


src/agnocastlib/src/agnocast_subscription.cpp
```diff
@@ -184,44 +184,18 @@ TypeSupportBundle GenericSubscription::load_typesupport_impl(const std::string &
...
-  // adding the new fd to epoll (EPOLL_CTL_ADD) can fail with EEXIST because epoll still
-  // associates that fd number with the stale entry.
-  {
-    std::lock_guard<std::mutex> lock(id2_callback_info_mtx);
-    id2_callback_info.erase(callback_info_id_);
+  const rmw_ret_t ret = rmw_serialize(raw, type_support, &out.get_rcl_serialized_message());
+  if (ret != RMW_RET_OK) {
+    RCLCPP_ERROR(
+      logger, "rmw_serialize failed (rmw_ret=%d); skipping message", static_cast<int>(ret));
+    rmw_reset_error();
```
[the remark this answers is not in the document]
bdm-k: The existing error handling for `rmw_serialize()` return value is sufficient.

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
```diff
@@ -2,7 +2,6 @@
 
 #include "agnocast/agnocast_callback_info.hpp"
 #include "agnocast/agnocast_ioctl.hpp"
-#include "agnocast/agnocast_mq.hpp"
 #include "agnocast/agnocast_public_api.hpp"
 #include "agnocast/agnocast_smart_pointer.hpp"
 #include "agnocast/agnocast_tracepoint_wrapper.h"
```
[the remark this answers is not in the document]
bdm-k: fixed in [34dd14e](https://github.com/autowarefoundation/agnocast/pull/1438/commits/34dd14e83512f9908d764ff0b5d67750d28d7841)

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
```diff
@@ -456,20 +456,60 @@ class PollingSubscriber
...
+        }
+        callback(std::move(serialized));
+      } else {
+        rclcpp::SerializedMessage serialized;
+        if (!serialize_message(message.get(), ts_bundle.handle, serialized)) {
+          return;
+        }
+        callback(serialized);
+      }
+    };
```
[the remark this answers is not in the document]
bdm-k: fixed in [34dd14e](https://github.com/autowarefoundation/agnocast/pull/1438/commits/34dd14e83512f9908d764ff0b5d67750d28d7841)

A null check is skipped since `ipc_shared_ptr` cannot be null.

src/agnocastlib/CMakeLists.txt
```diff
@@ -155,7 +155,6 @@ if(BUILD_TESTING)
     test/unit/test_agnocast_context.cpp
     test/unit/test_diagnostic_updater.cpp
     test/unit/test_type_registry_writer.cpp
-    test/unit/test_agnocast_generic_callback.cpp
```
Koichi98, line 159: The deleted unit test covered the error paths (wrong type / null pointer / serialize failure). Are these intentionally dropped, or should they be ported to cover the new `get_subscription_callback`?

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
```diff
@@ -456,20 +459,63 @@ class PollingSubscriber
...
+      "Callback must be invocable with one of the following arguments "
+      "(or any types implicitly convertible from them, e.g., const variants): "
+      "std::unique_ptr<rclcpp::SerializedMessage>, "
+      "std::shared_ptr<rclcpp::SerializedMessage>, or "
+      "rclcpp::SerializedMessage &.");
+
+    TypeSupportBundle ts_bundle = load_typesupport_impl(topic_type);
+
+    return [callback = std::forward<Func>(callback),
+            ts_bundle = std::move(ts_bundle)](ipc_shared_ptr<void> && message) {
```
Koichi98, line 503: The null-pointer guard from the old get_erased_generic_callback `(if (!raw_ptr) return;)` is gone here. Is a null message no longer possible on this path?

Koichi98: Also, this lambda used to be mutable. Was dropping it intentional? Stateful (non-const-callable) callbacks would no longer compile.

Koichi98:
@bdm-k 
> We might want to address the inconsistency between TypeErasedPublisher and Subscription<void>. I'd appreciate your thoughts on this.

I don't think this is a problem with an appropriate reason. Could add the small comment about this inconsistency in the source code? both for `TypeErasedPublisher` and `Subscription<void>`

src/agnocastlib/CMakeLists.txt
bdm-k: I was lazy on this. I will definitely check if I can add equivalent tests.

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
bdm-k: I don't think a null check is necessary here. Generally, `ipc_shared_ptr` cannot be null. After all, we do not add a null-pointer guard when writing a subscription callback.

bdm-k: I overlooked this, but you're right. We should make the wrapper `mutable` as well to support stateful callbacks. Thanks for catching this. Actually, it reminds me of the service/client implementations, where `mutable` is also missing. I will address them in follow-up PRs.

bdm-k: In fact, creating a `mutable` subscription callback isn't currently possible because `get_erased_callback()` (in `agnocast_callback_info.hpp`), at the root of the call chain, returns a stateless lambda. It might be better to address the related issues together in a separate PR.

src/agnocastlib/CMakeLists.txt
bdm-k: After review, I don't think we really need to bring back the unit tests. Here are the reasons:

- **Test 1: normal path - rmw_serialize round-trip via String.** The `GenericSubscriptionRoundTripTest` integration test already verifies serialization round-trip.
- **Test 2: null raw pointer — callback must be skipped (not called).** The null guard is gone.
- **Test 3: wrong envelope type — must exit with failure.** The entire type-erased generic callback machinery was removed, so it no longer makes sense.

bdm-k:
I documented design rationale for `TypeErasedPublisher` and `Subscription<void>` inconsistency.

Koichi98 (approved):

PR #1440, opened 2026-07-07
feat(agnocastlib): add vendored GenericService

bdm-k:
## Description

This PR vendors `GenericService` from rclcpp. Although `GenericService` exists in the latest rclcpp, it is not available in the versions of ROS 2 Humble or Jazzy. We need it to implement generic service bridging in the near future, so we have decided to vendor it. It is located within the `agnocast::vendor_rclcpp` namespace. Much of the implementation is taken directly from `rclcpp`, with a few minor refactorings to align it with Agnocast's conventions.

### Key differences from upstrem

1. **Simplified callback model.** The number of callback variants is reduced from 5 to 2 (`BasicCallback` + `DeferredCallback`), similar to Agnocast's `Service`.
2. **Factory function.** In order to pack everything inside the `GenericService` class, the factory function (`create_generic_service()`) is defined as a static member function of `GenericService`.
3. **Introspection loading.** The way of loading introspection is changed to make it work with both Humble and Jazzy.
4. **Tracing removed.** No `tracetools` dependency.

## Related links

The implementation of `GenericService` in rclcpp:

- https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/generic_service.hpp
- https://github.com/ros2/rclcpp/blob/rolling/rclcpp/src/rclcpp/generic_service.cpp

## How was this PR tested?

- [ ] Autoware
- [x] sample application

I used a client and a generic server to verify that they work as expected. The source code is available at the `bdm-k/vendored-generic-service-test` branch.

## Post-merge checklist

### DIFF
--- a/src/agnocastlib/include/agnocast/vendor/rclcpp/generic_service.hpp
+++ b/src/agnocastlib/include/agnocast/vendor/rclcpp/generic_service.hpp
@@ -1,0 +1,201 @@
+#pragma once
+
+#include <rclcpp/expand_topic_or_service_name.hpp>
+#include <rclcpp/function_traits.hpp>
+#include <rclcpp/macros.hpp>
+#include <rclcpp/node.hpp>
+#include <rclcpp/service.hpp>
+#include <rcpputils/shared_library.hpp>
+#include <rosidl_typesupport_introspection_cpp/message_introspection.hpp>
+
+#include <rcl/node.h>
+#include <rcl/service.h>
+#include <rclcpp/version.h>
+#include <rmw/types.h>
+
+#include <memory>
+#include <type_traits>
+#include <variant>
+
+namespace agnocast::vendor_rclcpp
+{
+
+#if RCLCPP_VERSION_MAJOR < 28
+
+const rosidl_service_type_support_t * get_service_typesupport_handle(
+  const std::string & type, const std::string & typesupport_identifier,
+  rcpputils::SharedLibrary & library);
+
+#endif
+
+class GenericService;
+
+class GenericServiceCallback
+{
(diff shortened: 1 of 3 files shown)


src/agnocastlib/include/agnocast/vendor/rclcpp/generic_service.hpp
```diff
@@ -0,0 +1,201 @@
+#pragma once
```
Koichi98, line 20: Could you add the lisence info here as we've done in https://github.com/autowarefoundation/agnocast/blob/3a9b92077c0db1833f5b87bedc11a6228fef4941/src/agnocastlib/src/node/tf2/transform_listener.cpp#L1?

Koichi98:
@bdm-k 
Could you also add in the Description that this is the temporary solution since `GenericService` isn't available in ros2 humble/jazzy?

Please also add what you've changed from the original code of 
- https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/generic_service.hpp
- https://github.com/ros2/rclcpp/blob/rolling/rclcpp/src/rclcpp/generic_service.cpp
in this branch.

src/agnocastlib/include/agnocast/vendor/rclcpp/generic_service.hpp
bdm-k: added in [ac1410b](https://github.com/autowarefoundation/agnocast/pull/1440/commits/ac1410bd4cf1501fc8be4e201cb1d44589265698)

As Apache 2.0 requires it, I added modification notices just below the license comments.

Koichi98 (dismissed):

Koichi98 (approved):

PR #1444, opened 2026-07-09
fix(ros2agnocast): error handling for discovery agent

Koichi98:
## Description
Guard the `get_node_names()` call in `ros2 node list_agnocast`) with try/except.

### Why
The RMW layer can raise `RCLError: empty node name returned by the RMW layer` when a participant without a valid node name is present in the DDS graph. This aborted the whole command with a traceback, even though the Agnocast nodes had already been collected via gossip/ioctl.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Post-merge checklist

**Important notes:**


src/ros2agnocast/ros2agnocast/verb/node_list_agnocast.py
```diff
@@ -64,7 +66,20 @@ def likely_shadow_node(fqn):
...
 
             # Get ros2 node names.
-            ros2_node_name_list = get_node_names(node=node, include_hidden_nodes=args.all)
+            # The RMW layer can raise (e.g. "empty node name returned by the RMW
+            # layer" when a participant without a valid node name is present in the
+            # DDS graph). Don't let that abort the whole command: fall back to the
+            # Agnocast nodes gathered from gossip so the output is still useful.
+            try:
+                ros2_node_name_list = get_node_names(node=node, include_hidden_nodes=args.all)
+            except Exception as e:
```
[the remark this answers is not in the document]
Koichi98: I'd keep the broad `except` here. Through the `ros2` daemon this surfaces as an `xmlrpc.client.Fault` wrapping the `RCLError`, not `RCLError` itself, so `except RCLError` would miss the case this PR fixes. It's not silent either — we log a WARNING with the original exception and just fall back to the gossip-based node list.

src/ros2agnocast/ros2agnocast/verb/node_list_agnocast.py
```diff
@@ -64,7 +66,20 @@ def likely_shadow_node(fqn):
                     return False
 
             # Get ros2 node names.
-            ros2_node_name_list = get_node_names(node=node, include_hidden_nodes=args.all)
+            # The RMW layer can raise (e.g. "empty node name returned by the RMW
+            # layer" when a participant without a valid node name is present in the
+            # DDS graph). Don't let that abort the whole command: fall back to the
+            # Agnocast nodes gathered from gossip so the output is still useful.
+            try:
+                ros2_node_name_list = get_node_names(node=node, include_hidden_nodes=args.all)
```
k1832, line 74: Check for other occurence!

k1832 (dismissed):
nit only

src/ros2agnocast/ros2agnocast/verb/node_list_agnocast.py
Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1444/commits/ed23c526715611ccf4a69d36b10e37a4a014993b.

kobayu858 (approved):

PR #1454, opened 2026-07-14
feat(agnocast_cie_thread_configurator): support wildcard callback group ids (<node name>/*) in the config YAML

atsushi421:
## Description

In the `agnocast_cie_thread_configurator` config YAML, a `callback_groups` `id` previously had to match the announced callback-group ID exactly. Since the ID embeds the callback group's entities (`<node>@Subscription(/topic)@Timer(...)`), every change to a node's callbacks forced a rewrite of all of its YAML entries.

This PR adds a wildcard form: `id: <fully qualified node name>/*` applies the scheduling attributes to **every** callback group whose node-name part (before the first `@`) equals the given node name, within the same domain.

- Matching is per node: `/perception/lidar_node/*` does not match other nodes or child-namespace nodes.
- An exact `id` entry takes precedence over a wildcard entry for the same callback group, so individual callback groups can still be overridden.
- Wildcards are supported only for `callback_groups`; `non_ros_threads` names remain exact-match.
- Malformed patterns (`/*`, `/node*`, `/a/*/b`, `/node@Timer(100)/*`) are rejected at parse time (fail-fast).
- A wildcard entry tracks its matched instances (full callback group ID to tid). The `reapply_config` service re-issues syscalls per instance and reports one `"<domain>:<full id>"` key per instance in `applied`/`failed`; the pattern key appears in `skipped` only while nothing has matched yet.
- Callback groups are matched to entries at announcement time only, and reapply carries state over per entry within the same form (exact entries carry `thread_id`, wildcard entries carry their matched instances). Changing an entry between exact and wildcard form, or adding an exact override for a callback group already matched by a wildcard, therefore takes effect at the next announcement (i.e. when the target application restarts); reapply is for updating the attributes (policy / priority / affinity) of existing entries.

```yaml
callback_groups:
  # Applies to all callback groups of /perception/lidar_node
  - id: /perception/lidar_node/*
    affinity: [2, 3]
    policy: SCHED_FIFO
    priority: 50

  # Exact entry overrides the wildcard for this specific callback group
  - id: /perception/lidar_node@Subscription(/points)
    affinity: [4]
    policy: SCHED_FIFO
    priority: 80
```

## Related links

N/A

## How was this PR tested?

- [ ] Autoware
- [ ] kunit tests (required when modifying the kernel module) - N/A (no kernel module changes)
- [ ] sample application
- [x] Newly added tests

## Post-merge checklist

**Important notes:**

### DIFF
--- a/src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
+++ b/src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
@@ -413,9 +413,8 @@ void ThreadConfiguratorNode::callback_group_callback(
     config = it->second;
     already_seen = config->applied;
   } else {
-    const std::string node_part =
-      agnocast_cie_thread_configurator::extract_node_part(msg->callback_group_id);
-    auto wit = node_to_wildcard_config_.find(std::make_pair(domain_id, node_part));
+    auto wit = node_to_wildcard_config_.find(std::make_pair(
+      domain_id, agnocast_cie_thread_configurator::extract_node_part(msg->callback_group_id)));
     if (wit == node_to_wildcard_config_.end()) {
       RCLCPP_INFO(
         this->get_logger(),
--- a/src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
+++ b/src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
@@ -433,8 +433,7 @@ void ThreadConfiguratorNode::callback_group_callback(
   if (already_seen) {
     // Always re-apply: the OS may reuse the same thread IDs after an application
     // restarts, so we cannot use thread_id equality to skip reconfiguration.
-    // "tracked", not "configured": for a wildcard instance already_seen only
-    // means the tid was recorded; its first syscall may have failed.
+    // "tracked", not "configured": a recorded wildcard tid's syscall may have failed.
     RCLCPP_INFO(
       this->get_logger(),
       "Re-applying configuration for already tracked callback group "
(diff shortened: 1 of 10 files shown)


src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
```diff
@@ -565,12 +601,37 @@ void ThreadConfiguratorNode::on_reapply_config_request(
...
+      // reported as skipped only while no instance has been announced yet.
+      if (cfg.matched_tids.empty()) {
+        response->skipped_callback_groups.push_back(
+          std::to_string(cfg.domain_id) + ":" + cfg.thread_str);
+        continue;
+      }
+      bool any_applied = false;
+      for (const auto & [full_id, tid] : cfg.matched_tids) {
+        std::string key = std::to_string(cfg.domain_id) + ":" + full_id;
+        if (issue_syscalls(cfg, tid)) {
```
kobayu858, line 622: Since `matched_tids` is only ever appended to and never pruned, could exact and wildcard entries end up double-applying to the same instance? e.g.:

1. With only `/node/*` in the YAML, `/node@Timer`(tid=100) is announced → recorded in `matched_tids`.
2. A reapply adds an exact `/node@Timer` (wildcard kept). The exact is `skipped` (no known tid), but `matched_tids` still holds `/node@Timer`.
3. `/node@Timer` is re-announced → the announce path routes to the exact entry, but the wildcard's `matched_tids` isn't updated/pruned.
4. Next reapply → this loop applies to `matched_tids["/node@Timer"]` *and* the exact entry applies to the same tid.

That would mean a double syscall + a duplicate key in `applied_callback_groups`, and since the loop iterates `callback_group_configs_` in YAML order, a wildcard listed after the exact would win and override it — breaking the "exact takes precedence" contract after reapply.

The trigger is fairly narrow (exact added via reapply, then re-announced), so if this is acceptable, maybe just note it as a limitation in `ReapplyConfig.srv`. Otherwise, skipping `full_id`s that have an exact entry in this loop (or pruning them from `matched_tids` when the announce path hits the exact entry) would avoid it.

atsushi421: Good catch. Fixed in f32a5078 by taking the pruning approach: when the announce path routes an instance to an exact entry, the instance is now erased from the wildcard's `matched_tids`, so each instance is tracked by exactly one entry again. Also documented the takeover semantics in `ReapplyConfig.srv` and added a regression test that triggers the re-announcement by publishing a synthetic `CallbackGroupInfo`.

kobayu858 (approved):

PR #1455, opened 2026-07-15
fix(bridge): A2R bridge subscription lost when its group is classified before the sub exists

Koichi98:
## Description

An A2R pubsub bridge could be created successfully yet silently deliver **zero** messages, while other topics in the same process worked.

**Cause.** `CallbackIsolatedAgnocastExecutor` picks a callback group's executor type (plain `rclcpp` vs agnocast-capable) from a one-time snapshot of its agnocast subscriptions. The A2R plugin creates the group (`automatically_add_to_executor_with_node() == true`) and only *then* adds the subscription, so the executor's poll can classify the group before the subscription exists. It is then bound permanently to a plain `rclcpp` executor with no agnocast epoll: the subscription's queue is created but never read, its depth-1 notification queue fills, and every `mq_send` returns `EAGAIN` and is dropped. Being a race, it reproduces intermittently.

**Fix.** Make the ordering deterministic:

- **Plugin**: create the A2R group with `automatically_add_to_executor_with_node = false`.
- **Manager**: add the group to the executor explicitly, after the subscription is registered, in both `create_pubsub_bridge_if_needed` and `activate_daemon_forced_bridge`.
- **Executor**: the monitoring loop also scans `weak_groups_to_nodes_` (manually added groups), like the startup scan, so explicitly added groups get a child executor.

R2A is unaffected (its bridge-side entity is a plain `rclcpp` subscription).

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Notes for reviewers

The bug is hit via the bridge, but the underlying executor snapshot behavior is reachable by any code that adds an agnocast subscription to a group dynamically after `add_node`. A more general executor-side fix may be worth considering separately.

## Post-merge checklist

**Important notes:**


src/agnocastlib/src/bridge/performance/agnocast_performance_bridge_manager.cpp
```diff
@@ -480,12 +488,20 @@ void PerformanceBridgeManager::create_pubsub_bridge_if_needed(
...
+            result.callback_group, container_node_->get_node_base_interface());
+        }
         active_pubsub_r2a_bridges_[topic_name] = result;
       } else {
         if (!update_ros2_subscriber_num(container_node_.get(), topic_name)) {
           RCLCPP_ERROR(
             logger_, "Failed to update ROS 2 subscriber count for topic '%s'.", topic_name.c_str());
         }
+        if (result.callback_group) {
+          executor_->add_callback_group(
```
atsushi421, line 508: This add_callback_group inserts an entry into the executor's weak_groups_to_nodes_, but teardown (check_and_remove_pubsub_bridges) only calls stop_callback_group, which touches child_callback_groups_ and not weak_groups_to_nodes_, and never calls remove_callback_group. Every bridge create/destroy cycle therefore leaks a map entry, which the new monitoring-loop scan then re-locks on every poll. Pair each add with executor_->remove_callback_group(...) on teardown, applied symmetrically since remove_callback_group exit()s on a missing group, and perform the removal under mutex_ in an order consistent with stop/erase so the scan cannot re-spawn a group mid-teardown.

src/agnocastlib/src/bridge/performance/agnocast_performance_bridge_loader.cpp
```diff
@@ -275,7 +275,7 @@ PerformancePubsubBridgeResult PerformanceBridgeLoader::create_a2r_pubsub_bridge_
     topic_name, message_type,
     rclcpp::QoS(agnocast::DEFAULT_QOS_DEPTH).reliable().transient_local());
 
-  auto cb_group = node->create_callback_group(rclcpp::CallbackGroupType::MutuallyExclusive);
+  auto cb_group = node->create_callback_group(rclcpp::CallbackGroupType::MutuallyExclusive, false);
```
atsushi421, line 340: The template and executor explain why auto_add is false (the manager adds the group explicitly after the subscription is created), but this identical non-obvious change here, in create_r2a_pubsub_bridge_generic, and in generic_functions.cpp has no comment. Since roughly 30 other call sites use the 1-arg form, a future cleanup could drop false at one of these sites and make the manager's explicit add_callback_group abort the process. Add the same explanatory comment at these sites.

atsushi421 (commented):
add_callback_group now runs the "already exists in node" rejection only when automatically_add_to_executor_with_node() is true, but the symmetric guard in add_node (agnocast_callback_isolated_executor.cpp:325-338) still exits unconditionally for any group of the node found in weak_groups_to_nodes_. Manually adding an auto_add=false group and then adding its owning node would abort even though no real double-add occurs. This is not reachable from the current bridge manager flow, but the paired checks are now inconsistent, so apply the same auto_add condition to add_node.

---

Service bridges still create their callback groups with the default auto_add=true (service_bridge_plugin.cpp.em) and tear down with stop_callback_group only, so they remain on the auto-add classification path this PR argues is race-prone. Worth confirming whether service bridges are subject to the same classify-before-entity-registered race; if so, apply the same create-with-false plus explicit-add pattern, or note in the PR why they are out of scope.

Koichi98:
@atsushi421 
> This is not reachable from the current bridge manager flow, but the paired checks are now inconsistent, so apply the same auto_add condition to add_node.

Fixed in https://github.com/autowarefoundation/agnocast/pull/1455/commits/efd3b72c8f45e38637a4cba1291ca2fe1c0630c0!

> Worth confirming whether service bridges are subject to the same classify-before-entity-registered race; if so, apply the same create-with-false plus explicit-add pattern, or note in the PR why they are out of scope.

Service bridges aren't affected. Classification uses `get_agnocast_topics_by_group()`, which reads `id2_callback_info` — written only by subscription registration. Agnocast Service/Client never register there, so a service bridge's group always reports zero agnocast topics and is correctly placed on a plain rclcpp executor, regardless of timing. There's no misclassification to race, so no fix is needed.

src/agnocastlib/src/bridge/performance/agnocast_performance_bridge_loader.cpp
Koichi98: Added in https://github.com/autowarefoundation/agnocast/pull/1455/commits/7071de3d63fddedd5adcb0a35cc3179244368d6a.

src/agnocastlib/src/bridge/performance/agnocast_performance_bridge_manager.cpp
Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1455/commits/b5c593f3b268d45a4ae6578b7ac9be6319915d18.

src/agnocastlib/src/bridge/performance/agnocast_performance_bridge_manager.cpp
```diff
@@ -480,12 +488,20 @@ void PerformanceBridgeManager::create_pubsub_bridge_if_needed(
           RCLCPP_ERROR(
             logger_, "Failed to update ROS 2 publisher count for topic '%s'.", topic_name.c_str());
         }
+        if (result.callback_group) {
+          executor_->add_callback_group(
+            result.callback_group, container_node_->get_node_base_interface());
+        }
```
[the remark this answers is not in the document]
Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1455/commits/b5c593f3b268d45a4ae6578b7ac9be6319915d18.

src/agnocastlib/src/bridge/performance/agnocast_performance_bridge_manager.cpp
```diff
@@ -242,12 +242,20 @@ void PerformanceBridgeManager::activate_daemon_forced_bridge(
         RCLCPP_ERROR(
           logger_, "Failed to update ROS 2 publisher count for topic '%s'.", topic_name.c_str());
       }
+      if (result.callback_group) {
+        executor_->add_callback_group(
+          result.callback_group, container_node_->get_node_base_interface());
+      }
```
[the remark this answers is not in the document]
Koichi98: Fixed.

src/agnocastlib/src/agnocast_callback_isolated_executor.cpp
```diff
@@ -127,6 +127,21 @@ void CallbackIsolatedAgnocastExecutor::spin()
...
+        auto group = weak_group_to_node.first.lock();
+        if (!group || group->get_associated_with_executor_atomic().load()) {
+          continue;
+        }
+        auto node = weak_group_to_node.second.lock();
+        if (!node) {
+          continue;
+        }
+        new_groups.emplace_back(group, node);
+      }
```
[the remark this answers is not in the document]
Koichi98: Restore the valide tests for `CallbackIsolatedAgnocastExecutor`.

src/agnocast_e2e_test/test/config_test_1to1.yaml
```diff
@@ -1,7 +1,7 @@
 use_agnocast_pub: true
 launch_pub_before_sub: true
-pub_qos_depth: 10
+pub_qos_depth: 1
 pub_transient_local: true
-sub_qos_depth: 10
+sub_qos_depth: 1
 sub_transient_local: true
```
kobayu858, line 6: Could you please revert? It looks like some leftover files from your local testing were pushed.

kobayu858:
@Koichi98 

> Service bridges aren't affected. Classification uses get_agnocast_topics_by_group(), which reads id2_callback_info — written only by subscription registration. Agnocast Service/Client never register there, so a service bridge's group always reports zero agnocast topics and is correctly placed on a plain rclcpp executor, regardless of timing. There's no misclassification to race, so no fix is needed.



It looks like it does register there — `BasicClient`/`BasicService` each create a real `Subscription` under that group ([agnocast_subscription.hpp#L172-L189](https://github.com/autowarefoundation/agnocast/blob/c1f1ac9dfe0f7cf856aa863c6385bc9d11179f30/src/agnocastlib/include/agnocast/agnocast_subscription.hpp#L172-L189)), and its constructor calls `register_callback()` unconditionally regardless of `role`, which writes into `id2_callback_info` ([agnocast_callback_info.hpp#L92-L122](https://github.com/autowarefoundation/agnocast/blob/c1f1ac9dfe0f7cf856aa863c6385bc9d11179f30/src/agnocastlib/include/agnocast/agnocast_callback_info.hpp#L92-L122)). Am I missing something that would exempt this internal subscription from that path?

Koichi98:
@kobayu858 

> It looks like it does register there — BasicClient/BasicService each create a real Subscription under that group ([agnocast_subscription.hpp#L172-L189](https://github.com/autowarefoundation/agnocast/blob/c1f1ac9dfe0f7cf856aa863c6385bc9d11179f30/src/agnocastlib/include/agnocast/agnocast_subscription.hpp#L172-L189)), and its constructor calls register_callback() unconditionally regardless of role, which writes into id2_callback_info ([agnocast_callback_info.hpp#L92-L122](https://github.com/autowarefoundation/agnocast/blob/c1f1ac9dfe0f7cf856aa863c6385bc9d11179f30/src/agnocastlib/include/agnocast/agnocast_callback_info.hpp#L92-L122)).

You're right, thanks for catching this — my earlier reasoning was wrong. Fixed with the same pattern as the pubsub bridges in https://github.com/autowarefoundation/agnocast/pull/1455/commits/4d945f2b60c0be2bb79ac578fc5dff16b63ce514.

src/agnocast_e2e_test/test/config_test_1to1.yaml
Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1455/commits/a3572d6dbb970f43a93c493c37f062202d0a7a04.

kobayu858 (dismissed):

kobayu858 (approved):

atsushi421 (approved):

PR #1466, opened 2026-07-16
feat(agnocastlib): add vendored GenericClient

bdm-k:
## Description

This PR vendors `GenericClient` to make it available in ROS 2 Humble. We will need it to implement generic A2R service bridging. Since `GenericClient` does not exist in ROS 2 Humble but is available in ROS 2 Jazzy, the vendored `GenericClient` is guarded by `RCLCPP_VERSION_MAJOR < 28` to ensure the original `GenericClient` is used in ROS 2 Jazzy.

### Key differences from upstream

1. **Stripped down `async_send_request()`.** Only the overload with a callback is implemented. As part of this change, the management of pending requests is also simplified.
1. **Factory function.** In order to pack everything inside the `GenericClient` class, the factory function (`create_generic_client()`) is defined as a static member function of `GenericClient`.
1. **Introspection loading.** The way of loading introspection is changed to make it work in ROS 2 Humble.

## How was this PR tested?

- [ ] Autoware
- [x] sample application

I added a sample client application that uses the vendored `GenericClient` and verified it works as expected with a ROS 2 server application. The source code is available at the `bdm-k/vendored-generic-client-test` branch.

## Post-merge checklist

### DIFF
--- a/src/agnocastlib/include/agnocast/vendor/rclcpp/generic_client.hpp
+++ b/src/agnocastlib/include/agnocast/vendor/rclcpp/generic_client.hpp
@@ -17,6 +17,8 @@
 // GenericClient is not available in ROS 2 Humble, so it is vendored here to implement Agnocast's
 // service bridging feature.
 
+#pragma once
+
 #include <rclcpp/version.h>
 
 #if RCLCPP_VERSION_MAJOR < 28
--- a/src/agnocastlib/include/agnocast/vendor/rclcpp/generic_client.hpp
+++ b/src/agnocastlib/include/agnocast/vendor/rclcpp/generic_client.hpp
@@ -1,0 +1,98 @@
+// Copyright 2023 Sony Group Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


src/agnocastlib/include/agnocast/vendor/rclcpp/generic_client.hpp
```diff
@@ -0,0 +1,98 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// This file has been modified from the original.
+
+// GenericClient is not available in ROS 2 Humble, so it is vendored here to implement Agnocast's
+// service bridging feature.
+
```
Koichi98, line 19: Better to add `#pragma once`.

bdm-k: fixed

Koichi98 (dismissed):

Koichi98 (approved):

PR #1468, opened 2026-07-17
refactor(agnocastlib): extract ClientBase class with shared client functionality

bdm-k:
## Description

This PR extracts `ClientBase` class with shared client functionality to eliminate code duplication. It also renames `BasicClient` to `Client` so that the naming convention matches `Publisher` and `Subscription`.

### Changes

**Extract `ResponseCallInfo` to a standalone template.** The `ResponseCallInfo` struct was previously duplicated as a private nested struct in each class. It's now defined once at the top level, parameterized on the response type. Both `Client` and `GenericClient` alias it via a private `using` declaration.

**Introduce `ClientBase` class.** A new `ClientBase` base class consolidates the shared client members and methods:

- Members: `next_sequence_numbers_`, `node_name_`, `service_name_`, `check_context_ok_`
- Methods: `get_service_name()`, `service_is_ready()`, `wait_for_service()`

### Impact

The service bridge plugin template is updated to use `Client` instead of `BasicClient`, so bridge plugins need recompilation.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Post-merge checklist


src/agnocastlib/include/agnocast/agnocast_client.hpp
```diff
@@ -43,12 +43,82 @@ enum class ClientRole : uint8_t {
   AgnocastOnly,
 };
 
-// Internal implementation - users should use agnocast::Client<ServiceT> instead.
+/// ResponseCallInfo template - parameterized on the response type.
+template <typename Response>
+struct ResponseCallInfo
```
Koichi98, line 51: `ResponseCallInfo` leaks into the agnocast namespace, but it's an internal detail — rclcpp seems like it keeps its equivalent protected inside the client class. Suggest making it a protected member of `ClientBase`.

src/agnocastlib/include/agnocast/agnocast_client.hpp
```diff
@@ -43,12 +43,82 @@ enum class ClientRole : uint8_t {
...
+
+  ResponseCallInfo() = default;
+
+  explicit ResponseCallInfo(std::function<void(SharedFuture)> && cb) : callback(std::move(cb))
+  {
+    shared_future = promise.get_future().share();
+  }
+};
+
+class ClientBase
```
Koichi98, line 69: `ClientBase` is a base class in a public header, so it'd be good to add a protected destructor to prevent deletion through a base pointer.

src/agnocastlib/include/agnocast/agnocast_client.hpp
bdm-k: fixed in [55f77cf](https://github.com/autowarefoundation/agnocast/pull/1468/commits/55f77cfece0a152e3e5e8e6019b7636ea45c75f9)

I hid `ResponseCallInfo` behind the `detail` namespace because I didn't want to template-parameterize `ClientBase`.

src/agnocastlib/include/agnocast/agnocast_client.hpp
bdm-k: fixed in [d2dc417](https://github.com/autowarefoundation/agnocast/pull/1468/commits/d2dc41755f6b36c435b934968c9c36657d96b7a8)

Koichi98 (approved):

PR #1470, opened 2026-07-21
feat(agnocastlib): add GenericService

bdm-k:
## Description

This PR introduces `GenericService`, a type-erased Agnocast service server. `GenericService` mirrors `BasicService` but operates on type-erased `ipc_shared_ptr<void>` payloads. Under the hood, it uses `TypeErasedPublisher` and `Subscription<void>` to implement the functionality.

### Interface summary

| Function | Type |
| -------- | ---- |
| Basic callback | `void(ipc_shared_ptr<void> && request, ipc_shared_ptr<void> && response)` |
| Deferred callback | `void(std::shared_ptr<GenericService> service_handle, ipc_shared_ptr<void> && request)` |
| `send_response()` | `void(ipc_shared_ptr<void> && request, ipc_shared_ptr<void> && response)` |
| `cancel_response()` | `void(ipc_shared_ptr<void> && request, ipc_shared_ptr<void> && response)` |
| `borrow_loaned_response()` | `ipc_shared_ptr<void>(const ipc_shared_ptr<void> & request)` |

## How was this PR tested?

- [ ] Autoware
- [x] sample application

I wrote a sample server application using `GenericService` and verified that it communicates with the sample client application without any problems. The code is available in branch `bdm-k/generic-service-test`. Though the sample server uses the basic callback mode, I will verify the deferred callback mode thoroughly when I implement the generic fallback for A2R service bridges.

## Notes for reviewers

- `BasicService` and `GenericService` share much code. I will refactor them to deduplicate code in the next PR.

## Post-merge checklist

### DIFF
--- a/src/agnocastlib/include/agnocast/agnocast_service.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_service.hpp
@@ -276,7 +276,20 @@ class GenericService : public std::enable_shared_from_this<GenericService>
       ipc_shared_ptr<void> response = std::move(res_wrapper).take_response();
       ipc_shared_ptr<void> response_double(response);
 
-      callback(std::move(req_wrapper).take_request(), std::move(response_double));
+      // If the callback throws, we destroy the `response` (ipc_shared_ptr<void>) via
+      // cancel_message() to prevent ipc_shared_ptr::reset() from calling std::terminate(), and then
+      // rethrow. We only need to destroy `response`, not `response_double`:
+      // (1) If `response_double` was moved from, it is empty and does not need to be destroyed.
+      // (2) If `response_double` was not moved from, it will be invalidated when `response` is
+      //     destroyed.
+      try {
+        callback(std::move(req_wrapper).take_request(), std::move(response_double));
+      } catch (...) {
+        publisher->cancel_message(std::move(response), [this](void * p) {
+          GenericResponseWrapper::free(p, this->response_members_);
+        });
+        throw;
+      }
 
       publisher->publish(std::move(response), [this](void * p) {
         GenericResponseWrapper::free(p, this->response_members_);
--- a/src/agnocastlib/include/agnocast/agnocast_service.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_service.hpp
@@ -221,6 +221,17 @@ public:
   }
 };
 
+/**
+ * @brief Generic service server for zero-copy Agnocast service communication.
+ *
+ * The service type is supplied as a runtime string, rather than a compile-time template argument.
+ * If the given service type is invalid, the constructor will throw an exception.
+ *
+ * The usage is mostly the same as agnocast::Service. One difference is cancel_response(), which
+ * only exists in GenericService. This is relevant for deferred callbacks. The user must either call
+ * send_response() or cancel_response() for every response borrowed via borrow_loaned_response().
+ * Otherwise, the process will terminate.
+ */
 class GenericService : public std::enable_shared_from_this<GenericService>
 {
   template <typename Func>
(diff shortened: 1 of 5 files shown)


src/agnocastlib/include/agnocast/agnocast_service.hpp
```diff
@@ -221,6 +221,156 @@ class BasicService : public std::enable_shared_from_this<BasicService<ServiceT>>
...
+
+      auto res_wrapper = GenericResponseWrapper::allocate(
+        response_members_,
+        [&publisher](size_t size) { return publisher->borrow_loaned_message(size); });
+      res_wrapper.seqno() = req_wrapper.seqno();
+
+      ipc_shared_ptr<void> response = std::move(res_wrapper).take_response();
+      ipc_shared_ptr<void> response_double(response);
+
+      callback(std::move(req_wrapper).take_request(), std::move(response_double));
```
Koichi98: A throwing basic callback destroys `response` before `publish()`, which calls `std::terminate()` for `ipc_shared_ptr<void>` — so it crashes the process, unlike the typed path. Maybe wrap it in try/catch + `cancel_message()`.

src/agnocastlib/include/agnocast/agnocast_service.hpp
```diff
@@ -221,6 +221,156 @@ class BasicService : public std::enable_shared_from_this<BasicService<ServiceT>>
...
+      //   dereferencing `response_double` after publication is disallowed, preventing accidental
+      //   (and erroneous) writes to the response via `response_double`.
+    };
+  }
+
+  template <typename Func>
+  auto wrap_deferred_service_callback_for_subscriber(Func && callback)
+  {
+    return [this, callback = std::forward<Func>(callback)](ipc_shared_ptr<void> && request) {
+      callback(this->shared_from_this(), std::move(request));
```
Koichi98, line 310: `shared_from_this()` throws `std::bad_weak_ptr` if the instance isn't held by a `shared_ptr`. With no factory and a public constructor, stack/`new` construction compiles but crashes here at runtime on the first deferred call. Consider adding a factory (mirroring `create_service`) to enforce `make_shared`.

src/agnocastlib/include/agnocast/agnocast_service.hpp
bdm-k: If the callback takes `response_double` by value or moves it to a local variable, it will crash regardless of whether we wrap the callback in a try/catch. What do you think about this?

src/agnocastlib/include/agnocast/agnocast_service.hpp
bdm-k: [809f7de](https://github.com/autowarefoundation/agnocast/pull/1470/commits/809f7def71abdb17ac8e429df007c31aac6a2d25)

I added `create_generic_service` to `agnocast.hpp`, but intentionally did not make it public. I'd like to make it public once we're confident that the generic service/client changes are working correctly.

src/agnocastlib/include/agnocast/agnocast_service.hpp
Koichi98: `reset()` only reaches `std::terminate()` on the **last** reference while unpublished. The wrapper keeps `response` alive and hands the callback only a shared copy (`response_double`), so the callback's by-value arg or local is never the last reference — destroying it during unwinding is `was_last == false` and returns normally. The terminate actually comes from *our* `response` being destroyed after the callback, which a try/catch can catch and `cancel_message()`. `cancel_message()` also invalidates all references, so a stashed copy is fine too.

bdm-k: Ah, that's right.

bdm-k: fixed in [fd7e4d5](https://github.com/autowarefoundation/agnocast/pull/1470/commits/fd7e4d59a7787b93fce6092a4ee8791c12efe099)

Koichi98 (approved):

PR #1473, opened 2026-07-21
perf(agnocastlib)[need-minor-update]: exec the discovery agent directly instead of via ros2 run

Koichi98:
## Description

Auto-start the discovery agent by exec'ing `agnocast_discovery_agent` directly instead of going through `ros2 run`.

The kmod keys agent liveness on the PID that claims the per-(IPC namespace, domain) singleton, but `ros2 run` starts the agent as a `subprocess.Popen` grandchild under a different PID, so the claim cannot be taken before the agent starts. Every process coming up while no agent exists therefore pays for a Python runtime, only for all but one to lose the claim and exit. The forked child now claims the singleton itself and then execs the agent with `--already-registered`, so the losers exit before starting Python and a concurrent startup brings up exactly one agent. `ros2 launch` / `ros2 run` still claim for themselves.

Each process still forks once, because the claim has to come from the process that becomes the agent. Reserving in the parent instead would need a new kmod path to release a reservation the parent cannot observe, since the daemon child is `setsid`'d and never waited on. A read-only `AGNOCAST_DISCOVERY_AGENT_EXISTS_CMD` check just before the fork would skip most of these forks with no kmod change; left as follow-up.

## Related links

#1496 renamed the executable to `agnocast_discovery_agent`.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Notes for reviewers

Resolution reuses `ament_index_cpp::get_package_prefix()`, the same lookup `ros2 run` performs, so the search range is unchanged: only sourced ROS prefixes, never the wider `PATH`, which could otherwise pick up a same-named executable from an unrelated or version-skewed install. An exec failure in the child is reported with `write()`, which — as for the other daemons spawned here — is only visible if the parent's stdout/stderr is not a pipe or socket.

## Post-merge checklist

`need-patch-update`


sykwer:
### N = 200

  | Metric | main | this PR | Improvement |
  |---|---:|---:|---:|
  | Peak discovery-agent processes | 388 | 1 | 388× |
  | Time until all N finished `initialize_agnocast()` | 0.998 s | 0.329 s | 3.0× |
  | Time until the launch stopped generating work | 14.43 s | 0.33 s | 43.9× |
  | Per-process init, p50 | 518 ms | 211 ms | 2.5× |
  | Per-process init, p95 | 864 ms | 270 ms | 3.2× |
  | Per-process init, worst | 918 ms | 288 ms | 3.2× |
  | CPU burned by the launch (above idle) | 222 CPU-s | 1.8 CPU-s | 124× |
  | Extra RAM at peak | +7.30 GB | +0.55 GB | 13.3× |
  | `import rclpy` during the launch (idle = 0.23 s) | 6.65 s | 0.50 s | 13.2× |

  ### N = 400

  | Metric | main | this PR | Improvement |
  |---|---:|---:|---:|
  | Peak discovery-agent processes | 800 | 1–2 | ~400× |
  | Time until all N finished `initialize_agnocast()` | 3.740 s | 0.612 s | 6.1× |
  | Time until the launch stopped generating work | 29.47 s | 0.61 s | 48.2× |
  | Per-process init, p50 | 1135 ms | 339 ms | 3.3× |
  | Per-process init, p95 | 3330 ms | 422 ms | 7.9× |
  | Per-process init, worst | 3377 ms | 469 ms | 7.2× |
  | CPU burned by the launch (above idle) | 446 CPU-s | 4.1 CPU-s | 109× |
  | Extra RAM at peak | +14.41 GB | +1.02 GB | 14.1× |
  | `import rclpy` during the launch (idle = 0.23 s) | 10.91 s | 0.79 s | 13.8× |

  ### How it scales, 200 → 400

  | Metric | main | this PR |
  |---|---:|---:|
  | Time until all N finished init | **3.75×** | 1.86× |
  | Per-process init, p95 | **3.86×** | 1.56× |
  | Time until the launch stopped generating work | 2.04× | 1.86× |
  | CPU burned by the launch | 2.00× | 2.28× (1.8 → 4.1 CPU-s) |
  | Peak discovery-agent processes | 2.00× (388 → 800) | flat (1) |

sykwer (approved):

PR #1475, opened 2026-07-24
feat(agnocastlib): add generic A2R service bridge fallback

bdm-k:
## Description

This PR implements the generic A2R service bridge, completing the generic service bridging story alongside the existing R2A generic service bridge. 

The most significant change is `create_a2r_service_bridge_generic()`. It mirrors the plugin template, but applies the details to `agnocast::GenericService` and the vendored `vendor_rclcpp::GenericClient`. Similar to `create_r2a_service_bridge_generic()`, it leverages CDR serialization/deserialization to copy the ROS 2 response into Agnocast shared memory.

### Other changes

- **Vendored `GenericClient` unguarded.** The initial plan was to use rclcpp's `GenericClient` for ROS 2 Jazzy. But it turned out that `GenericClient` in Jazzy does not support response callbacks. Therefore, I decided to use the vendored one regardless of whether it is Humble or Jazzy.
- **Unified plugin-load error logging.** `get_bridge_factory_symbol()` no longer contains service-specific error paths.

## How was this PR tested?

- [ ] Autoware
- [x] sample application

Verified that the sample client can send requests to a sample ROS 2 server.

## Post-merge checklist


src/agnocastlib/src/bridge/performance/agnocast_performance_bridge_loader.cpp
```diff
@@ -392,4 +365,52 @@ ServiceBridgeEntity PerformanceBridgeLoader::create_r2a_service_bridge_generic(
...
+  auto ros_client = agnocast::vendor_rclcpp::GenericClient::create_generic_client(
+    node.get(), service_name, service_type, qos, client_cbg);
+
+  auto agno_srv = std::make_shared<agnocast::GenericService>(
+    node.get(), service_name, service_type,
+    [ros_client = std::move(ros_client), response_copier = get_copier(response_type)](
+      std::shared_ptr<agnocast::GenericService> service_handle, ipc_shared_ptr<void> && agno_req) {
+      const void * ros_req_ptr = agno_req.get();
+
+      ros_client->async_send_request(
```
Koichi98: This can throw at runtime. Since this runs in a deferred service callback, the exception would escape the spin thread and `std::terminate` the process.

Could we wrap the `async_send_request` call in try/catch, log via `RCLCPP_ERROR`, and drop the request?

bdm-k: Sorry, I missed this case. Should exceptions from `async_send_request()` be rethrown, or should they be swallowed (my understanding is that we should rethrow them)? I'm not clear on the exception-handling standard in the bridge manager, so I'd appreciate your guidance.

Koichi98: Better swallowing it. Setup-time exceptions do have a handler — `start_a2r_bridge` catches them and turns them into an error string — but a callback-time one has none: it escapes the spin thread and terminates the whole bridge daemon over a single failed `rcl_send_request`. Log + drop also matches the serialization-error path right below it.

bdm-k: I see. Do we also need to update the A2R service bridge plugin similarly to keep exceptions of `ros_client->async_send_request()` from escaping the spin thread?

Koichi98: True. Could you address it in this PR?

bdm-k: fixed in [df5f402](https://github.com/autowarefoundation/agnocast/pull/1475/commits/df5f402206d2d005775d7f4b553a05d5664bb588)

`get_service_name()` is added to `BasicService` though it is not marked as `AGNOCAST_PUBLIC`. This will be addressed in the refactoring PR of service classes.

Koichi98 (dismissed):

Koichi98 (approved):

PR #1476, opened 2026-07-24
test(e2e_test): add e2e service tests

bdm-k:
## Description

This PR adds end-to-end integration tests for Agnocast service including service bridges. The test launch file is `test/test_service.py`. It uses the `launch_testing.parametrize` decorator to run the launch description multiple times, once per set of parameter values. For all parameter sets, the common procedure is to launch one server and one client in separate  processes, have them communicate `TARGET_COUNT` (the default is three) times, and then verify from the process output that no requests or responses are lost.

The launch test file defines five parameters:

| Parameter | Apply to | Description |
| --------- | ---- | ----------- |
| `server_plugin` | Server | `"TestServer"` or `"TestROS2Server"`. `"TestROS2Server"` is used for testing the A2R service bridge. |
| `use_deferred_callback` | Server | Boolean. If `True`, the deferred service callback is used, and the response is returned from a background thread after a certain amount of time. |
| `client_plugin` | Client | `"TestClient"` or `"TestROS2Client"`. `"TestROS2Client"` is used for testing the R2A service bridge. |
| `use_response_callback` | Client | Boolean. If `True`, the client uses a callback-based `async_send_request`; otherwise uses future-based awaiting. |
| `wait_response` | Client | Boolean. If `True`, the client waits for each response before sending the next request. Coerced to `True` when `use_response_callback` is `False`. |

Besides these parameters, the service name, QoS depth, and target communication count are also variables that can be configured via node parameters. That said, they are fixed for now.

### Test Matrix

Below is a table of the nine parameter-value combinations that the E2E test runs.

| # | Server Plugin | Deferred | Client Plugin | Response Callback | Wait Response |
| - | ------------- | -------- | ------------- | ----------------- | ------------ |
| 1 | TestServer | no | TestClient | no | yes |
| 2 | TestServer | yes | TestClient | no | yes |
| 3 | TestServer | no | TestClient | yes | yes |
| 4 | TestROS2Server | no | TestClient | no | yes |
| 5 | TestServer | no | TestROS2Client | no | yes |
| 6 | TestServer | no | TestClient | yes | no |
| 7 | TestServer | yes | TestClient | yes | no |
| 8 | TestROS2Server | no | TestClient | yes | no |
| 9 | TestServer | no | TestROS2Client | yes | no |

The former five combinations test the synchronous request pattern with `wait_response` being false, while the latter four test the asynchronous request pattern. Each parameter set differs from the baseline set (1 or 6) in exactly one parameter, allowing us to clearly test the specific part that differs.

### Follow-ups

- [ ] Add tests to check the `ros2 param` command.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Post-merge checklist


bdm-k:
@Koichi98 
How should we organize this service E2E test? Maybe a straightforward way would be adding `scripts/test/e2e_test_service.bash` and running it for every PR, though a new E2E test script makes it a bit more cumbersome to run E2E tests.

bdm-k:
The tests that use `TestROS2Server` fail on ROS 2 Jazzy. Changing the component container executable to `agnocast_component_container_cie` fixes the issue. `SingleThreadedAgnocastExecutor` may be missing something that was introduced in ROS 2 Jazzy.

Update:
This will be fixed by #1549.

Koichi98:
Let's go with `scripts/test/e2e_test_service.bash` in this PR. The wider question of how to organize the e2e scripts applies to the existing ones too, so I'll look at that separately rather than blocking this PR.

src/agnocast_e2e_test/src/test_client.cpp
```diff
@@ -0,0 +1,211 @@
...
+    return true;
+  }
+
+public:
+  explicit TestClient(const rclcpp::NodeOptions & options) : Node("test_client", options)
+  {
+    auto params = get_node_params(this);
+    use_response_callback_ = params.use_response_callback;
+    wait_response_ = params.wait_response;
+    // If the peer is a ROS 2 server, use a longer interval to account for the overhead introduced
```
Koichi98: There is no interval in `TestClient`, so this comment looks left over from an earlier implementation.

src/agnocast_e2e_test/src/test_client.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+class TestROS2Client : public rclcpp::Node
+{
+  bool wait_response_;
+  int64_t target_count_;
+
+  rclcpp::CallbackGroup::SharedPtr cbg_;
+  rclcpp::Client<ServiceT>::SharedPtr client_;
+
+  std::int64_t response_count_;
```
Koichi98: `response_count_` is not initialized here while `TestClient` initializes it at line 53, so the parameter sets using `TestROS2Client` either shut down after the first response or never shut down at all.

src/agnocast_e2e_test/src/test_client.cpp
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    return true;
+  }
+
+public:
+  explicit TestROS2Client(const rclcpp::NodeOptions & options) : Node("test_ros2_client", options)
+  {
+    auto params = get_node_params(this);
+    assert(params.use_response_callback && "TestROS2Client must use response callback");
```
Koichi98: `assert` is compiled out under `-DCMAKE_BUILD_TYPE=Release`, which both `scripts/dev/build_all.bash` and `e2e_test_1to1.bash` use, so this guard disappears in the build most people run. Same for `test_server.cpp:126`.

src/agnocast_e2e_test/src/test_server.cpp
```diff
@@ -0,0 +1,144 @@
...
+  {
+    RCLCPP_INFO(this->get_logger(), "Receiving %ld.", request->data[0]);
+
+    std::thread([this, srv_handle = std::move(srv_handle), request = std::move(request)]() {
+      // Wait for a while to simulate an asynchronous operation.
+      std::this_thread::sleep_for(100ms);
+
+      auto response = srv_handle->borrow_loaned_response(request);
+      response->sum = request->data[0];
+      auto request_movable = request;
```
Koichi98: `borrow_loaned_response()` does not consume `request`, so making the lambda `mutable` lets you move it into `send_response()` instead of copying through a second handle.

src/agnocast_e2e_test/src/test_server.cpp
```diff
@@ -0,0 +1,144 @@
...
+      srv_handle->send_response(std::move(request_movable), std::move(response));
+
+      received_count_ += 1;
```
Koichi98: With `use_deferred_callback=True` and `wait_response=False`, three response threads increment this concurrently, so a lost update leaves `received_count_` short of `target_count_` and the server never shuts down.

src/agnocast_e2e_test/src/test_server.cpp
```diff
@@ -0,0 +1,144 @@
...
+      if (received_count_ >= target_count_) {
+        RCLCPP_INFO(this->get_logger(), "All requests have been handled. Shutting down.");
+        rclcpp::shutdown();
+      }
+    }).detach();
```
Koichi98: These threads capture `this` but nothing keeps the node alive, so the container can destroy `TestServer` while a thread is still in `send_response()`. `test_client.cpp:130` and `:205` have the same pattern.

src/agnocast_e2e_test/test/test_service.py
```diff
@@ -0,0 +1,146 @@
...
+        # basic arrangement
+        ("TestServer", False, "TestClient", False, True),
+        # deferred server callback
+        ("TestServer", True, "TestClient", False, True),
+        # callback-based async_send_request()
+        ("TestServer", False, "TestClient", True, True),
+        # ROS2 server to test A2R service bridge
+        ("TestROS2Server", False, "TestClient", False, True),
+        # ROS2 client to test R2A service bridge
+        ("TestServer", False, "TestROS2Client", False, True),
```
Koichi98: `TestROS2Client` always uses the callback overload of `async_send_request`, so this should be `True` to match both the implementation and the assertion at `test_client.cpp:181`.

src/agnocast_e2e_test/test/test_service.py
```diff
@@ -0,0 +1,146 @@
...
+        proc_info.assertWaitForShutdown(process=client_container, timeout=10)
+
+@launch_testing.post_shutdown_test()
+class TestService(unittest.TestCase):
+
+    # Ensure that the server and client exit successfully.  It also sleeps for two seconds to avoid
+    # the known issue with the bridge manager (#1501).
+    def test_exit_codes(self, proc_info):
+        assertExitCodes(proc_info, allowable_exit_codes=[EXIT_OK])
+        time.sleep(2)
```
Koichi98: The dominant cause of #1501 has been fixed, so this two-second sleep can be dropped.

src/agnocast_e2e_test/src/test_client.cpp
Koichi98: Fixed in 9818c797.

src/agnocast_e2e_test/src/test_client.cpp
Koichi98: Fixed in a5436ad7.

src/agnocast_e2e_test/src/test_client.cpp
Koichi98: Fixed in 84dfbca1.

src/agnocast_e2e_test/src/test_server.cpp
Koichi98: Fixed in 64fb741e.

src/agnocast_e2e_test/src/test_server.cpp
Koichi98: Fixed in 7bfd5d55.

src/agnocast_e2e_test/src/test_server.cpp
Koichi98: Fixed in 44ae87f8.

src/agnocast_e2e_test/test/test_service.py
Koichi98: Fixed in 0830134b.

src/agnocast_e2e_test/test/test_service.py
Koichi98: Fixed in e843db6d.

src/agnocast_e2e_test/test/test_service.py
```diff
@@ -0,0 +1,144 @@
...
+from launch_testing.actions import ReadyToTest
+from launch_testing.asserts import EXIT_OK, assertExitCodes, assertInStderr
+
+SERVICE_NAME = '/test_service'
+QOS_DEPTH = 10
+TARGET_COUNT = 3
+
+
+@launch_testing.parametrize(
+    'server_plugin, use_deferred_callback, client_plugin, use_response_callback, wait_response',
```
Koichi98: The nine launches are indistinguishable in the output, so a failure gives no indication of which parameter set produced it. `e2e_test_1to1.bash:316` prints `====== N / TOTAL ======` before each combination.

src/agnocast_e2e_test/src/test_client.cpp
```diff
@@ -0,0 +1,236 @@
...
+                RCLCPP_INFO(this->get_logger(), "All responses have been received. Shutting down.");
+                rclcpp::shutdown();
+              }
+            })
+          .future;
+    } else {
+      sfut = client_->async_send_request(std::move(request)).share();
+    }
+
+    if (!wait_response_) {
```
Koichi98, line 104: The asynchronous pattern returns before the two-second wait below, so a response that never arrives leaves the client spinning until `assertWaitForShutdown` gives up. `:177` has the same shape.

src/agnocast_e2e_test/src/test_server.cpp
```diff
@@ -0,0 +1,177 @@
...
+  int64_t target_count_;
+  // Deferred callbacks respond from their own threads, so the counter is shared across them.
+  std::atomic<int64_t> received_count_{0};
+  rclcpp::CallbackGroup::SharedPtr cbg_;
+  agnocast::Service<ServiceT>::SharedPtr srv_;
+
+  std::mutex response_threads_mtx_;
+  std::vector<std::thread> response_threads_;
+
+  void count_and_shutdown_if_done()
```
Koichi98, line 76: This is the only path that shuts the server down, so a run where no request arrives never exits and the failure surfaces only as `assertWaitForShutdown` timing out on both processes at `test_service.py:118`.

src/agnocast_e2e_test/test/test_service.py
```diff
@@ -0,0 +1,144 @@
...
+
+@launch_testing.post_shutdown_test()
+class TestService(unittest.TestCase):
+
+    def test_exit_codes(self, proc_info):
+        assertExitCodes(proc_info, allowable_exit_codes=[EXIT_OK])
+
+    def test_server_received_all_requests(self, proc_output, server_container):
+        for i in range(TARGET_COUNT):
+            assertInStderr(proc_output, f'Receiving {i}.', server_container)
```
Koichi98: `assertInStderr` only checks for presence, so a response delivered twice or an `Agnocast internal implementation error: bad entry id` from `agnocast_client.hpp:193` still passes.

src/agnocast_e2e_test/test/test_service.py
Koichi98: Fixed in 032d1ab9.

src/agnocast_e2e_test/src/test_client.cpp
Koichi98: Fixed in bc899a89.

src/agnocast_e2e_test/src/test_server.cpp
Koichi98: Fixed in 78a065ee.

src/agnocast_e2e_test/test/test_service.py
Koichi98: Fixed in fcc4ab34.

Koichi98 (approved):

PR #1483, opened 2026-07-29
feat(ros2agnocast): add --spin-time to topic list_agnocast

Koichi98:
## Description

`ros2 topic list_agnocast` had no way to wait longer for discovery; its only knob was the hidden, transitional `--gossip-timeout`, which covers just the Agnocast half.

This adds `--spin-time` - same name, meaning and default (0.5s) as `ros2 topic list` - and drops `--gossip-timeout` from this verb. It covers both of the verb's waits, the ROS 2 graph and the Agnocast endpoints, and they now overlap instead of running back to back: the ROS 2 graph is discovered by time elapsing after the participant is created, not by spinning, so the gossip collection doubles as the ROS 2 wait.

- **Daemon running**: returns as soon as gossip arrives, whatever `--spin-time` says - the daemon answers the graph queries from its own long-running node. `ros2 topic list` ignores `--spin-time` under a daemon too.
- **No daemon**: waits the full `--spin-time` before querying the graph, like `ros2 topic list`.
- **No agent answering**: gossip cannot stop early, so it spends the whole window before falling back to the local-namespace ioctl view.

It also gets faster: the verb always builds a `DirectNode` for gossip, so it always paid ros2cli's 0.5s spin on top of up to 2.0s of gossip wait. That is now one 0.5s window.

## Related links

None.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application
- [x] `bash scripts/test/e2e_test_cross_ns_cli.bash`
- [x] Added `test_topic_list_agnocast.py`: 8 cases driving `main()` on a fake clock, each checked by mutation.

## Post-merge checklist


TetsuKawa (commented):
I went through this with Claude, including building the workspace and running the verb against a stand-in gossip publisher to get some actual numbers. A couple of things came out of that — one measurement worth sharing, and one small suggestion. Neither is blocking.

`warn_if_using_fallback` may go quiet when the budget runs out

Nice change — I measured the daemon path going from ~0.73 s to ~0.21 s, so the doubled wait is really gone. 👍

One interaction I noticed: `warn_if_using_fallback()` returns early on timeout_sec <= 0. That used to mean "the user explicitly asked not to wait", so staying silent made sense. Now gossip_timeout is a remainder, so it can also hit 0 when node construction ate a positive --spin-time — and then we silently fall back to the local-NS view, which looks like "no Agnocast topics elsewhere".

Not very likely in practice (I measured construction at ~33 ms vs the 0.5 s default), mostly a cold spawn_daemon / loaded-machine concern. Forcing it with --spin-time 0 does show the topics vanishing with no message.

Would it be worth splitting the conditions?


```
if not used_fallback:
    return
if timeout_sec <= 0:
    print('NOTE: spin time was exhausted before Agnocast discovery could run; '
          'showing local NS only. Consider a larger --spin-time.',
          file=sys.stderr)
    return
```
Fine as a follow-up if you'd rather keep the diff focused.

TetsuKawa (commented):
The help text could mention that --spin-time also applies with the daemon

ros2 core already has a --spin-time, and its help ends with "(only applies when not using an already running daemon)". Here it's different: gossip collection always runs in this process, so the flag matters either way. I confirmed that — with the daemon running, --spin-time 0 drops the Agnocast topics (0/5 runs).

Reusing the core name feels right to me, but the current help ("covering both the ROS 2 graph and the Agnocast endpoints") doesn't call out that difference, so someone used to the core flag may assume it's a no-op when the daemon is up. Maybe worth one extra clause?


```
help='Spin time in seconds to wait for discovery, covering both the '
     'ROS 2 graph and the Agnocast endpoints. Unlike the ros2 core flag, '
     'this applies even when the ros2 daemon is in use, because Agnocast '
     'discovery always runs in this process.'
```

Koichi98:
@TetsuKawa Thanks for the review — both addressed.

#### `warn_if_using_fallback` may go quiet when the budget runs out

Thanks for measuring this. The problem was in the accounting rather than in the warning, so I aligned with upstream ros2 instead.

`--spin-time` is now measured from when our participant exists, which is where ros2cli measures its own — `DirectNode` creates the node and *then* starts the spin timer. Gossip always gets the full window, so `timeout_sec` reaches 0 only when the operator asks for 0, which is what the `timeout_sec <= 0` early return originally meant. `discovery.py` is unchanged, so the verbs still on `--gossip-timeout` keep their current behaviour.

#### The help text could mention that `--spin-time` also applies with the daemon

Good catch, added — thanks for confirming it empirically.

```python
help='Spin time in seconds to wait for discovery, covering both the '
     'ROS 2 graph and the Agnocast endpoints. Unlike the ros2 core '
     'flag it also applies when the ros2 daemon is running, because '
     'Agnocast discovery always runs in this process'
```

Slightly reworded from your version, same content.

TetsuKawa:
**Suggestion: one more --spin-time test case?**

Moving the deadline to after node setup is a nicer fix than what I'd suggested — it removes the state instead of reporting it, and it keeps discovery.py untouched. I re-measured and the gossip window now gets 100% of the budget on the cold path (was ~93%). 👍

One small gap I found while poking at the tests: if I move deadline = time.monotonic() + spin_time back above NodeStrategy, all 8 tests still pass. test_node_setup_is_added_to_the_run_not_taken_out_of_the_wait uses gossip_elapsed=3.0, so the top-up sleep is zero there, and test_time_gossip_leaves_unused_is_slept_off uses no construction cost — the two never cross, so nothing pins the graph-wait half against setup cost.

Worth adding a case? This one is red on that mutation and green as-is:


```
def test_node_setup_does_not_eat_into_the_graph_wait_either():
    """Setup is added to the run for the ROS 2 graph half of the budget too."""
    clock, _ = _run_main(spin_time=3.0, construction_cost=0.4, gossip_elapsed=0.2)
    assert clock.elapsed == pytest.approx(3.4)
```
Nothing else from me — the rest looks good.

Koichi98:
@TetsuKawa Thanks for catching this — the 8 really were blind to it. Your case asserts the same `clock.elapsed == 3.4` the existing one already does, and `gossip_elapsed=3.0` just made it vacuous, so I dropped that to 0.2 rather than adding a ninth test. All 8 still pass as-is, and I confirmed that case now goes red with `deadline` moved back above `NodeStrategy`.

TetsuKawa (approved):

sykwer (approved):

PR #1488, opened 2026-07-31
fix(agnocastlib): callback type resolution failure in vendored GenericService

bdm-k:
## Description

This PR updates the `GenericServiceCallback` constructor:

1. **Stop using `rclcpp::function_traits::same_arguments`.** It performs too strict a type comparison, so it does not fit our use case. Instead, this PR uses `std::is_invocable_v` just like the Agnocast service implementation.
2. **Add a fallback else block.**  The block catches invalid callback types and asserts a failure at compile time.

## How was this PR tested?

- [ ] Autoware
- [x] sample application

## Post-merge checklist

### DIFF
--- a/src/agnocastlib/include/agnocast/vendor/rclcpp/generic_service.hpp
+++ b/src/agnocastlib/include/agnocast/vendor/rclcpp/generic_service.hpp
@@ -67,10 +67,17 @@ public:
       }
     }
 
-    if constexpr (::rclcpp::function_traits::same_arguments<Func, BasicCallback>::value) {
+    if constexpr (std::is_invocable_v<
+                    std::decay_t<Func>, std::shared_ptr<void> &&, const std::shared_ptr<void> &>) {
       callback_.template emplace<BasicCallback>(std::forward<Func>(callback));
-    } else if constexpr (::rclcpp::function_traits::same_arguments<Func, DeferredCallback>::value) {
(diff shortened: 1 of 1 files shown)


src/agnocastlib/include/agnocast/vendor/rclcpp/generic_service.hpp
```diff
@@ -67,10 +67,17 @@ class GenericServiceCallback
       }
     }
 
-    if constexpr (::rclcpp::function_traits::same_arguments<Func, BasicCallback>::value) {
+    if constexpr (std::is_invocable_v<
+                    std::decay_t<Func>, std::shared_ptr<void> &&, const std::shared_ptr<void> &>) {
```
Koichi98: `BasicCallback`/`DeferredCallback` take their parameters by value, so `std::function` seems to pass every argument on as an rvalue. If so, `const std::shared_ptr<void> &` looks a bit too strict and would reject a callback taking `std::shared_ptr<void> &&` — same for the deferred handles. Since we only need to know whether `emplace` can build the callback, maybe check that directly?

```cpp
if constexpr (std::is_constructible_v<BasicCallback, Func &&>) {
```

bdm-k: fixed

Koichi98 (approved):

PR #1241, opened 2026-04-08
feat(agnocastlib): add NodeGraphInterface support

tomiy-0x62:
## Description
add NodeGraphInterface support

- Updated `docs/agnocast_node_interface_comparison.md`.
- Implemented `agnocat_ioctl_get_node_names()` in `agnocast_kmod` to support the implementation of `NodeGraph::get_node_names()`.
- Added `agnocast_kunit_get_node_names` to `kunit_test` to verify the new ioctl cmd.

## How was this PR tested?

- [ ] Autoware (required)
- [ ] `bash scripts/test/e2e_test_1to1.bash` (required)
- [ ] `bash scripts/test/e2e_test_2to2.bash` (required)
- [ ] kunit tests (required when modifying the kernel module)
- [ ] `bash scripts/test/run_requires_kernel_module_tests.bash` (required)
- [ ] sample application

## Version Update Label (Required)

Please add **exactly one** of the following labels to this PR:

- `need-major-update`: User API breaking changes
- `need-minor-update`: Internal API breaking changes (heaphook/kmod/agnocastlib compatibility)
- `need-patch-update`: Bug fixes and other changes

**Important notes:**

- If you need `need-major-update` or `need-minor-update`, please include this in the PR title as well.
  - Example: `fix(foo)[needs major version update]: bar` or `feat(baz)[needs minor version update]: qux`
- After receiving approval from reviewers, add the `run-build-test` label. The PR can only be merged after the build tests pass.

See [CONTRIBUTING.md](../CONTRIBUTING.md) for detailed versioning rules.

PR #1494, opened 2026-08-03
feat(agnocastlib): add NodeGraphInterface skeleton

Koichi98:
## Description

Adds `agnocast::node_interfaces::NodeGraph`, an implementation of `rclcpp::node_interfaces::NodeGraphInterface` reachable via `agnocast::Node::get_node_graph_interface()`.

- `count_publishers()` / `count_subscribers()` are supported. They delegate to the same helpers as `agnocast::Publisher::get_subscription_count()` / `agnocast::Subscription::get_publisher_count()`, so the bridge-endpoint bookkeeping stays in one place.
- `notify_graph_change()` / `notify_shutdown()` are no-ops. agnocast has no graph events, and rclcpp utilities call these unconditionally when entities are created or the context shuts down, so throwing would break otherwise working code.
- The remaining methods throw `std::runtime_error`. The per-method status is documented in `docs/agnocast_node_interface_comparison.md`.

No kmod change. `get_node_names()` needs the kmod to report the nodes owning an agnocast endpoint and is added in a follow-up.

## Related links

- Follow-up that implements `get_node_names()`: #1493
- Supersedes #1241

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Notes for reviewers

Only `count_publishers()` / `count_subscribers()` are reachable behaviour here; everything else either throws or is a no-op.

## Post-merge checklist

### DIFF
--- a/src/agnocastlib/include/agnocast/node/node_interfaces/node_graph.hpp
+++ b/src/agnocastlib/include/agnocast/node/node_interfaces/node_graph.hpp
@@ -4,10 +4,13 @@
 #include "rclcpp/event.hpp"
 #include "rclcpp/node_interfaces/node_base_interface.hpp"
 #include "rclcpp/node_interfaces/node_graph_interface.hpp"
-#include "rclcpp/rclcpp.hpp"
 
+#include <chrono>
 #include <map>
 #include <memory>
+#include <string>
+#include <tuple>
+#include <utility>
 #include <vector>
 
 namespace agnocast::node_interfaces
--- a/src/agnocastlib/include/agnocast/node/node_interfaces/node_graph.hpp
+++ b/src/agnocastlib/include/agnocast/node/node_interfaces/node_graph.hpp
@@ -26,6 +26,9 @@ public:
 
   virtual ~NodeGraph() = default;
 
+  NodeGraph(const NodeGraph &) = delete;
+  NodeGraph & operator=(const NodeGraph &) = delete;
+
   size_t count_publishers(const std::string & topic_name) const override;
   size_t count_subscribers(const std::string & topic_name) const override;
 
--- a/src/agnocastlib/include/agnocast/node/node_interfaces/node_graph.hpp
+++ b/src/agnocastlib/include/agnocast/node/node_interfaces/node_graph.hpp
@@ -4,6 +4,7 @@
 #include "rclcpp/event.hpp"
 #include "rclcpp/node_interfaces/node_base_interface.hpp"
 #include "rclcpp/node_interfaces/node_graph_interface.hpp"
+#include "rclcpp/version.h"
 
 #include <chrono>
 #include <map>
@@ -64,6 +65,12 @@ public:
     const std::string & topic_name, bool no_mangle = false) const override;
   std::vector<rclcpp::TopicEndpointInfo> get_subscriptions_info_by_topic(
     const std::string & topic_name, bool no_mangle = false) const override;
+
+  // rclcpp 28+ (Jazzy) added these methods to NodeGraphInterface.
+#if RCLCPP_VERSION_MAJOR >= 28
+  size_t count_clients(const std::string & service_name) const override;
+  size_t count_services(const std::string & service_name) const override;
+#endif
 
 private:
   NodeBase::SharedPtr node_base_;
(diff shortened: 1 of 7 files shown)


src/agnocastlib/src/node/node_interfaces/node_graph.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+NodeGraph::NodeGraph(NodeBase::SharedPtr node_base) : node_base_(std::move(node_base))
+{
+}
+
+std::map<std::string, std::vector<std::string>> NodeGraph::get_topic_names_and_types(
+  bool no_demangle) const
+{
+  (void)no_demangle;
+  throw std::runtime_error("NodeGraph::get_topic_names_and_types is not supported in agnocast. ");
```
kobayu858: Please delete the extra spaces at the end. There are several instances of this.

src/agnocastlib/src/node/node_interfaces/node_graph.cpp
```diff
@@ -0,0 +1,153 @@
...
+};
```
kobayu858: nit: stray `;` after the function body (and ~14 more below). Legal as an empty declaration, but `-Wextra-semi` flags it and the no-op methods below don't have it.

src/agnocastlib/include/agnocast/node/node_interfaces/node_graph.hpp
```diff
@@ -0,0 +1,66 @@
+#pragma once
+
+#include "agnocast/node/node_interfaces/node_base.hpp"
+#include "rclcpp/event.hpp"
+#include "rclcpp/node_interfaces/node_base_interface.hpp"
+#include "rclcpp/node_interfaces/node_graph_interface.hpp"
+#include "rclcpp/rclcpp.hpp"
```
kobayu858: Please drop `rclcpp/rclcpp.hpp` — this header is reachable from the public `agnocast_node.hpp`, so it leaks all of rclcpp into every user TU. `node_graph_interface.hpp` + `event.hpp` suffice.

That will expose missing standard includes: `<string>`, `<tuple>`, `<utility>`, `<chrono>` are all used in the declarations below but only come in via `rclcpp.hpp` today.

src/agnocastlib/src/node/node_interfaces/node_graph.cpp
```diff
@@ -0,0 +1,153 @@
+#include "agnocast/node/node_interfaces/node_graph.hpp"
```
kobayu858, line 1: `<stdexcept>` is missing — `std::runtime_error` is used throughout. `node_logging.cpp` includes it explicitly.

src/agnocastlib/include/agnocast/node/node_interfaces/node_graph.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+class NodeGraph : public rclcpp::node_interfaces::NodeGraphInterface
+{
+public:
+  using SharedPtr = std::shared_ptr<NodeGraph>;
+  using WeakPtr = std::weak_ptr<NodeGraph>;
+
+  explicit NodeGraph(NodeBase::SharedPtr node_base);
+
+  virtual ~NodeGraph() = default;
```
kobayu858, line 28: Please also delete the copy constructor / assignment operator here, as `NodeServices` and `NodeClock` do — this class holds a `NodeBase::SharedPtr` in the same way.

src/agnocastlib/src/node/node_interfaces/node_graph.cpp
Koichi98: Fixed in 515abe69!

src/agnocastlib/src/node/node_interfaces/node_graph.cpp
Koichi98: Fixed in f0f86dbd!

src/agnocastlib/include/agnocast/node/node_interfaces/node_graph.hpp
Koichi98: Fixed in f2b64fd3!

src/agnocastlib/src/node/node_interfaces/node_graph.cpp
Koichi98: Fixed in f2b64fd3!

src/agnocastlib/include/agnocast/node/node_interfaces/node_graph.hpp
Koichi98: Fixed in 1c8958bc!

kobayu858 (approved):

PR #1495, opened 2026-08-04
fix(agnocastlib): log daemon output to a file instead of /dev/null

Koichi98:
## Description

The forked daemons (shm_unlink, performance bridge manager, discovery agent) sent stdout/stderr to `/dev/null`, so nothing they logged was ever recoverable. Point them at a log file instead. The redirect itself stays: `dup2` closing the inherited descriptor is what stops the daemon from holding a pipe whose reader waits for EOF from a process that outlives it. Only the placeholder changes, so `RCLCPP_*` in a daemon needs no change. Watching the output on a terminal instead is #1500.

The file goes where node logs go (`AGNOCAST_DAEMON_LOG_DIR`, else `ROS_LOG_DIR` -> `$ROS_HOME/log` -> `~/.ros/log`), named `agnocast_<daemon>_ns<ipc_ns_inode>[_d<domain>].log`. No pid or timestamp, so a duplicate daemon that exits immediately leaves no empty file behind. Open failures warn and fall back to `/dev/null`. `RCUTILS_COLORIZED_OUTPUT=0` is set for the daemon so an environment forcing colour on does not put ANSI escapes in the file.

`AGNOCAST_DAEMON_LOG_DIR` is experimental and warns when set, so we can tell whether anyone relies on it. The warning is emitted where the value is read, which means once per process that actually spawns a daemon: the variable has no effect anywhere else, so that is where reporting it is useful.

Alternatives considered and rejected:

- Keep writing to the inherited pipe so `ros2 launch` captures daemon output like a node's: that is the hang. Deployments that want launch-managed logging can start the agent from `discovery_agent.launch.xml` with `AGNOCAST_NO_DISCOVERY_AGENT=1`.
- Configure rcl logging inside the daemon rather than redirecting descriptors: heavier in a forked child, and it would still not capture the exec'd Python agent's stderr or tracebacks.
- Per-instance file names like rcl's `<node>_<pid>_<ts>.log`: a duplicate daemon that loses its race and exits immediately would create an empty file on every node start.

## Related links

None.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Notes for reviewers

Worth confirming CTest still completes promptly, since that is what the `/dev/null` redirect was originally added for.

## Post-merge checklist

**Important notes:**

### DIFF
--- a/src/agnocastlib/src/agnocast.cpp
+++ b/src/agnocastlib/src/agnocast.cpp
@@ -463,6 +463,12 @@ std::string resolve_daemon_log_path(const char * daemon_name, const uint32_t * d
 
   const std::string dir = resolve_daemon_log_dir();
   if (dir.empty() || !make_directories(dir)) {
+    // Without a log file the daemon falls back to /dev/null, so say so rather than silently
+    // discarding its output.
+    RCLCPP_WARN_ONCE(
+      logger,
+      "No writable daemon log directory (tried ROS_LOG_DIR, $ROS_HOME/log, ~/.ros/log), so "
+      "Agnocast daemon output is discarded.");
     return "";
   }
 
@@ -486,19 +492,19 @@ std::string resolve_daemon_log_path(const char * daemon_name, const uint32_t * d
 void redirect_stdio_for_daemon(const char * daemon_name, const std::string & log_path)
 {
   const int devnull = open("/dev/null", O_RDWR);
-  if (devnull >= 0) {
-    static_cast<void>(dup2(devnull, STDIN_FILENO));
+  if (devnull >= 0 && dup2(devnull, STDIN_FILENO) < 0) {
+    _exit(EXIT_FAILURE);
   }
 
   int out_fd = -1;
-  bool wrote_to_log_file = false;
+  bool opened_log_file = false;
   // Opt-in only; the log file is the default target.
   if (env_is_truthy("AGNOCAST_DAEMON_LOG_TO_TTY")) {
     out_fd = open("/dev/tty", O_WRONLY);
   }
   if (out_fd < 0 && !log_path.empty()) {
     out_fd = open(log_path.c_str(), O_WRONLY | O_CREAT | O_APPEND, 0644);
-    wrote_to_log_file = out_fd >= 0;
+    opened_log_file = out_fd >= 0;
   }
   if (out_fd < 0) {
     out_fd = devnull;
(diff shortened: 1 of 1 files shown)


src/agnocastlib/src/agnocast.cpp
kobayu858: All three calls to `resolve_daemon_log_path()` sit inside `if (!add_process_args.ret_*_daemon_exist)` (lines 631, 639, 649), so when a daemon already exists in the same IPC namespace we never reach this function and the experimental-knob warning never fires.

Confirmed by starting the same node three times with `AGNOCAST_DAEMON_LOG_TO_TTY=1`:

```
run1: warned 2x   (spawned the daemons)
run2: warned 0x   (daemons still alive)
run3: warned 0x
```

Under `ros2 launch` only the first node to get there spawns the daemons, so the warning lands on that one node's stderr — and if a daemon from a previous run is still alive, nowhere at all. Given "Both are candidates for removal if unused", a warning that stays silent while someone is actually using the knob doesn't quite serve that purpose.

Moving the call to the top of `initialize_agnocast()` (before the spawn checks) makes it fire exactly once whenever the variable is set — `RCLCPP_WARN_ONCE` keeps it to one line even with several daemons. It also gets logging out of a `resolve_`-prefixed function.

src/agnocastlib/src/agnocast.cpp
```diff
@@ -486,19 +492,19 @@ std::string resolve_daemon_log_path(const char * daemon_name, const uint32_t * d
 void redirect_stdio_for_daemon(const char * daemon_name, const std::string & log_path)
```
kobayu858, line 477: The fds are `dup2`'d once at startup, with no reopen path and no `SIGHUP` handler, which makes `copytruncate` mandatory for logrotate.

I hit this during review: after deleting a log file, the still-running daemon kept writing to the unlinked inode and everything it emitted in that window was lost. The disk space isn't reclaimed until the daemon exits either.

Since there's no rotation or size cap, operators will reach for logrotate — and without `copytruncate`, daemon output from the first rotation onward is silently lost, which quietly undoes what this PR sets out to fix. Could we either document the `copytruncate` requirement or reopen on `SIGHUP`?

src/agnocastlib/src/agnocast.cpp
kobayu858, line 530: ANSI escapes end up in the log file. Confirmed with `od -c`:

```
0000000 033   [   0   m   [   I   N   F   O       1   7   8   5   8   9
```

Colorization stays on even though the target is a regular file rather than a tty, which gets in the way when reading with `grep` or `less`. Passing `RCUTILS_COLORIZED_OUTPUT=0` to the daemon child (lifting it only in TTY mode) seems like the straightforward fix.

kobayu858:
@Koichi98 
Verified locally from a clean build.

| Item                                         | Result       |
| -------------------------------------------- | ------------ |
| `e2e_test_1to1.bash`                         | 128/128 PASS |
| Integration tests (`requires_kernel_module`) | 10/10 PASS   |
| **CTest total time**                         | **48.49 s**  |

The "does CTest still complete promptly" question from Notes for reviewers looks fine. I also extracted the new functions into a standalone harness and confirmed that with stdout/stderr on a pipe the pipeline is released in 2 ms — the implicit close in `dup2` is doing its job.

The value shows up in practice too: an rclcpp warning that previously went to `/dev/null` is now recoverable.

```
[WARN] [agnocast_bridge_node_performance_...]: New publisher discovered on topic
  '/test_topic', offering incompatible QoS. No messages will be sent to it.
  Last incompatible policy: DURABILITY_QOS_POLICY
```

One observation: in `e2e_test_2to2.bash`, mode `agno2ros` failed 3 out of 7 runs (the `"Publishing 0."` assertion at `test_2to2.py:177`). The same binary both fails and passes, so it's flaky rather than deterministic. The merge base `dbe60b26` was 0 out of 6. It may well be specific to my machine — could you check whether it reproduces on yours?

Koichi98:
@kobayu858 Thanks for the review. Closing this and splitting it in two.

The terminal half is #1500, standalone on `main`: daemon output goes to the controlling terminal instead of `/dev/null`, falling back where there is no terminal.

The file half needs its own design first. As your logrotate comment shows, writing to a file also decides where operators look and how it rotates — user-facing surface that should not be settled inside a bug fix. I will open a PR once that is worked out.

PR #1496, opened 2026-08-04
refactor(ros2agnocast_discovery_agent): rename the executable to agnocast_discovery_agent

Koichi98:
## Description

Rename the discovery agent executable from `discovery_agent` to `agnocast_discovery_agent`. The old
name is generic enough to collide with an unrelated executable of the same name, which matters now
that the executable is also resolved by name rather than only through `ros2 run`.

The recommended way to start an agent manually becomes:

```
ros2 run ros2agnocast_discovery_agent agnocast_discovery_agent
```

The old name stays installed as a deprecated alias (both as a `bin/` console script and under
`lib/<pkg>/`), so launch files, systemd units and scripts that spawn `discovery_agent` keep working;
it prints a one-line deprecation notice on stderr and then runs the agent. Removing the alias is a
user-facing break and is left to a release that may make one.

`register_domain_bridge` is left as is.

## Related links
https://github.com/autowarefoundation/agnocast/issues/1497

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

`ros2 pkg executables ros2agnocast_discovery_agent` lists both names,
`ros2 launch ros2agnocast_discovery_agent discovery_agent.launch.xml` brings the agent up under the
new name, and `ros2 run ros2agnocast_discovery_agent discovery_agent` still starts it after printing
the deprecation notice.

## Notes for reviewers

`autoware_core`'s `autoware_agnocast_wrapper` spawns the agent with `executable="discovery_agent"`
(`launch/discovery_agent.launch.py`). A `launch_ros` `Node` whose executable cannot be found aborts
the whole launch tree, not just that node, which is why the alias is kept rather than renaming
outright.

## Post-merge checklist

Will be updated in the following PR.
https://github.com/autowarefoundation/agnocast_doc/pull/50

`need-patch-update`

### DIFF
--- a/src/agnocastlib/package.xml
+++ b/src/agnocastlib/package.xml
@@ -48,9 +48,9 @@
   <!-- Uncomment when agnocast_heaphook is released from PPA -->
   <!-- <exec_depend>agnocast_heaphook</exec_depend> -->
 
-  <!-- initialize_agnocast() forks+execs `ros2 run ros2agnocast_discovery_agent discovery_agent`
-       (the auto-forked discovery agent), so the package must be present at runtime. Runtime-only,
-       so it does not affect build order. -->
+  <!-- initialize_agnocast() forks+execs the auto-forked discovery agent
+       (`ros2 run ros2agnocast_discovery_agent agnocast_discovery_agent`), so the package must be
+       present at runtime. Runtime-only, so it does not affect build order. -->
   <exec_depend>ros2agnocast_discovery_agent</exec_depend>
 
   <test_depend>ament_cmake_gmock</test_depend>
--- a/scripts/test/e2e_test_cross_ns_cli.bash
+++ b/scripts/test/e2e_test_cross_ns_cli.bash
@@ -64,7 +64,7 @@ cleanup() {
 trap cleanup EXIT
 
 yellow "Starting agnocast_discovery_agent and talker..."
-ros2 run ros2agnocast_discovery_agent discovery_agent > "$LOG_DIR/agent.log" 2>&1 &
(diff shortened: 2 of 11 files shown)


src/agnocastlib/package.xml
```diff
@@ -48,9 +48,9 @@
   <!-- Uncomment when agnocast_heaphook is released from PPA -->
   <!-- <exec_depend>agnocast_heaphook</exec_depend> -->
 
-  <!-- initialize_agnocast() forks+execs `ros2 run ros2agnocast_discovery_agent discovery_agent`
-       (the auto-forked discovery agent), so the package must be present at runtime. Runtime-only,
-       so it does not affect build order. -->
+  <!-- initialize_agnocast() forks+execs the auto-forked discovery agent
```
kobayu858, line 51: This comment now says the auto-forked agent is started as `agnocast_discovery_agent`, but the
`execlp()` that actually starts it still uses the old name:

https://github.com/autowarefoundation/agnocast/blob/9d90f8429a88983f7531e0f59016afde58b93f08/src/agnocastlib/src/agnocast.cpp#L245-L249

Changing the 5th `execlp()` argument to `"agnocast_discovery_agent"` should be enough.

kobayu858 (changes requested):

src/agnocastlib/package.xml
Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1496/commits/7417cd8e4076c0bcba5c1f6524a93966f34e485e.

kobayu858 (approved):

sykwer (approved):

PR #1503, opened 2026-08-06
refactor(agnocast)[need-minor-update]: remove the subscriber MQ machinery eventfd made dead

Koichi98:
## Description

With eventfd notification (#1432) subscribers no longer create a POSIX MQ, so everything built to name and clean up those MQs is dead. This removes it.

- **Canonical MQ name**: `agnocast_notify_mq_topic_name()`. Only existed so a publisher and a renamed bridged subscriber derived the same MQ name.
- **Exit-cleanup pipeline**: `exit_subscription_entry` / `exit_subscription_mq_info` and the `ioctl_get_exit_process_args` buffer fields. `agnocast_commit_exit_process()` reduces to freeing `proc_info`; `poll_for_unlink()` to `shm_unlink`.
- **Userspace**: `create_mq_name_for_agnocast_publish()`, `agnocast_mq.hpp`, leftover `<mqueue.h>` includes.
- **Docs**: `docs/message_queue.md` and the README cleanup section. Two already-stale mentions — `docs/shared_memory.md`'s `/new_publisher@pid` and the bridge manager comment in `agnocast_ioctl.c` (UDS since #1433) — are corrected here too.
- **Stress test**: `e2e_test_stress.bash`'s `stress-ng --mq` load. CPU and VM loads are kept.
- **Leftover comments**: `get_exit_process_cmd()`'s now-pointless `copy_from_user`, `PublisherBase`'s "message queue management", the mocked `get_bridge_mode()`'s "skip MQ sending", `clang_tidy_suppression.md`'s `mq_open()` example, `e2e_test_domain_bridge.bash`'s two-exchange rationale, `test_bridge_decider.py`'s docstring, and the Promela model's notification annotations.
- **Now output-only**: `AGNOCAST_GET_EXIT_PROCESS_CMD` becomes `_IOR`, and `agnocast_ioctl_get_exit_process()` returns the global pid instead of taking an out-param.

Split out of #1432 so that PR is only the notification mechanism swap. No behavior change.

## Related links

- Follows #1432 (merged), which swapped the notification mechanism
- User-facing docs: autowarefoundation/agnocast_doc#52 — do not merge it before this PR is released, since `fs.mqueue.msg_max` is still a required setup step on 2.3.5

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Notes for reviewers

The two-phase get/commit split in `agnocast_ioctl_get_exit_process` is kept, but for a narrower reason: `ret_pid` must reach userspace before `proc_info` is freed, or the exited process's shm is never unlinked. The original motivation — avoiding `copy_to_user` of the MQ-info buffer under the global write lock — is gone with the buffer.

Three KUnit cases are dropped because the behavior they guarded no longer exists: `do_exit_subscription_mq_info{,_multi_topic}` and #1443's `domain_bridge_rename_exit_cleanup_uses_canonical_name`.

Left for follow-up PRs — all are changes rather than dead-code removal, so they are out of scope here:

- Phase 1 is now read-only, so its `down_write(&global_htables_rwsem)` can be `down_read`.
- No case covers process exit under a *renaming* bridge rule anymore; `domain_bridge_exit_frees_shared_struct` only covers the same-name case.
- No case asserts `ret_daemon_should_exit`; the dropped `do_exit_subscription_mq_info` cases carried the only ones.
- A failed `copy_to_user` here kills the daemon, and `get_process_num()` still counts exited entries so no replacement is forked — the reported pid's shm leaks. Pre-existing and practically unreachable; the comment no longer claims otherwise.

`checkpatch` error count is unchanged from the base (79 → 79).

## Post-merge checklist

https://github.com/autowarefoundation/agnocast_doc/pull/52

`need-minor-update` — kmod/agnocastlib ABI change (also flagged in the title).


atsushi421 (commented):
`AGNOCAST_GET_EXIT_PROCESS_CMD` is still encoded as `_IOWR` in both `agnocast_kmod/agnocast.h` and `src/agnocastlib/include/agnocast/agnocast_ioctl.hpp`, but with the input fields and the `copy_from_user` gone the ioctl is now output-only, so `_IOR` is the accurate direction (precedent: `AGNOCAST_GET_VERSION_CMD`). Since the struct shrink already changes the command number and the PR carries `need-minor-update`, this is the free moment to fix it; deferring it means a second ABI break later.

---

A few stale MQ references survive this sweep: the header comment of `scripts/test/e2e_test_domain_bridge.bash` still justifies the rename test's two-exchange design entirely in publish-notification-MQ terms (a mechanism this PR deletes), the module docstring of `src/ros2agnocast_discovery_agent/test/test_bridge_decider.py` still says `nor a POSIX MQ` (bridge communication is UDS since #1433), and `model_checking/agnocast_callback_isolated_executor` still annotates the notification path as MQ (`// mq_send() for publish` in `agnocast_pub_sub.pml`, `the corresponding mqueue is not modeled` in `data_structure.pml`). Worth updating here or noting as a follow-up.

---

In `src/agnocastlib/test/unit/tf2/test_transform_listener.cpp`, the deleted `open_mq_for_subscription` mock was the file's only user of `std::make_pair`, so the `#include <utility>` is now unused and can be removed with it.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -1396,26 +1394,19 @@ int agnocast_ioctl_get_publisher_num(
...
-// performs copy_to_user without any lock held.
-//
-// As a secondary benefit, Phase 2 runs only after the critical copies (ret_pid and
-// ret_subscription_mq_info_num) succeed, so subscription entries are never permanently lost.
-// ret_daemon_should_exit is patched via a separate copy_to_user after Phase 2; if that final copy
-// fails, the daemon merely stays alive one extra poll cycle — no resource leak.
+// Splitting the two lets the dispatch handler copy ret_pid out with no lock held, and only commit
+// once that copy succeeded: a failed copy leaves the entry for the next poll instead of dropping
+// the pid whose shm the daemon still has to unlink. ret_daemon_should_exit is patched by a
+// separate copy_to_user after Phase 2; if that one fails the daemon merely stays alive one extra
```
atsushi421: This rewritten comment doesn't match the end-to-end behavior: any failed `copy_to_user` makes the ioctl return `-EFAULT`, which `poll_for_unlink()` treats as fatal (close + exit), so there is no next poll, and no replacement daemon is ever forked since `get_process_num()` still counts the exited entries. In particular, when only this final patch copy fails, `proc_info` is already committed and the dying daemon never unlinks the reported pid's shm, which is a resource leak. The behavior is pre-existing and practically unreachable, but the new wording overstates the guarantee; consider describing only the kernel-side guarantee (a failed `ret_pid` copy returns `-EFAULT` before commit, so the entry is not dropped kernel-side; the daemon treats any ioctl error as fatal).

agnocast_kmod/agnocast_kunit/agnocast_kunit_publish_msg.c
```diff
@@ -612,18 +612,13 @@ void test_case_publish_msg_no_process(struct kunit * test)
...
-    agnocast_ioctl_get_exit_process(
-      current->nsproxy->ipc_ns, &get_exit_args, mq_info_buf, ARRAY_SIZE(mq_info_buf), &global_pid),
+    test, agnocast_ioctl_get_exit_process(current->nsproxy->ipc_ns, &get_exit_args, &global_pid),
     0);
   KUNIT_ASSERT_EQ(test, get_exit_args.ret_pid, exiting_pid);
   bool daemon_should_exit = false;
-  agnocast_commit_exit_process(
-    current->nsproxy->ipc_ns, global_pid, get_exit_args.ret_subscription_mq_info_num,
-    &daemon_should_exit);
+  agnocast_commit_exit_process(current->nsproxy->ipc_ns, global_pid, &daemon_should_exit);
```
atsushi421, line 620: The dropped `do_exit_subscription_mq_info` cases also carried the suite's only `KUNIT_EXPECT_TRUE(test, daemon_should_exit)` assertions, and that behavior still exists: `agnocast_commit_exit_process()` sets the flag when `get_process_num()` reaches 0. After this PR no KUnit case asserts the daemon-termination flag anywhere (this remaining call site never checks it). Consider adding a small do_exit case: register one process, enqueue its exit, get (expect `ret_pid`), commit (expect `daemon_should_exit == true`), then a second get (expect `ret_pid == -1`).

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -1396,26 +1394,19 @@ int agnocast_ioctl_get_publisher_num(
...
+// poll cycle -- no resource leak.
 int agnocast_ioctl_get_exit_process(
   const struct ipc_namespace * ipc_ns, struct ioctl_get_exit_process_args * ioctl_ret,
-  struct exit_subscription_mq_info * mq_info_buf, uint32_t mq_info_buf_size, pid_t * out_global_pid)
+  pid_t * out_global_pid)
```
atsushi421: With the buffer logic gone this function has no failure path left (a single `return 0;`), and its only production caller discards the return value. Since the signature is changing anyway, consider making it `void`, or returning the global pid directly and dropping the `out_global_pid` parameter.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -2765,50 +2699,21 @@ static long get_exit_process_cmd(struct ioctl_get_exit_process_args __user * arg
...
-
-  uint64_t mq_buf_addr = get_exit_process_args.subscription_mq_info_buffer_addr;
-  if (mq_buf_size > 0 && mq_buf_addr == 0) return -EINVAL;
-
-  struct exit_subscription_mq_info * mq_info_buf = NULL;
-  if (mq_buf_size > 0) {
-    mq_info_buf = kvcalloc(mq_buf_size, sizeof(*mq_info_buf), GFP_KERNEL);
-    if (!mq_info_buf) return -ENOMEM;
-  }
+  struct ioctl_get_exit_process_args get_exit_process_args = {};
```
atsushi421, line 2698: In the slimmed-down handler, `int ret = 0;` above is never reassigned (every failure path returns `-EFAULT` directly), so the tail `return ret;` is always 0. Also, the `(struct ioctl_get_exit_process_args __user *)arg` cast in the first `copy_to_user` is a no-op given `arg`'s type, as the second `copy_to_user` shows by using `arg` uncast. Consider dropping the dead variable (end with `return 0;`) and the cast.

src/agnocastlib/src/agnocast.cpp
```diff
@@ -184,18 +183,12 @@ initialize_agnocast_result acquire_agnocast_resources_for_bridge()
...
 void poll_for_unlink()
 {
-  std::vector<exit_subscription_mq_info> mq_info_buf(MAX_SUBSCRIPTION_NUM_PER_PROCESS);
-
   while (true) {
     sleep(1);
 
     struct ioctl_get_exit_process_args get_exit_process_args = {};
     do {
       get_exit_process_args = {};
```
atsushi421: This per-iteration reset was needed when the struct carried input fields that were re-populated each pass; now the kernel never reads the struct and fully overwrites it via `copy_to_user` on success, while a failed ioctl exits the daemon before the value is examined again. The `= {}` at the declaration is enough, so this line can go.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Comment fixed in da4f46a8 — it now states only the kernel-side guarantee. Dropped the same claim at the second `copy_to_user` too.

The leak itself is pre-existing, so it's listed under "Left for follow-up PRs".

agnocast_kmod/agnocast_kunit/agnocast_kunit_publish_msg.c
Koichi98: Agreed, the flag is untested after this PR. Adding a case is a change rather than dead-code removal, so it goes with the other deferred coverage gap — added to "Left for follow-up PRs".

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in 988bf775. Took the pid-returning option: `ret_pid` is the local pid and the out-param was the global one, so returning it drops the parameter and gives the return value a meaning.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in f9b3f032.

src/agnocastlib/src/agnocast.cpp
Koichi98: Fixed in 70f2bf5a.

Koichi98:
> `_IOR` is the accurate direction (precedent: `AGNOCAST_GET_VERSION_CMD`)

Fixed in 3fde99d1.

> A few stale MQ references survive this sweep

Fixed in 3fde99d1 — all three, plus the same wording in the script's body comments and `callback_info.mqdes` in the model.

> the `#include <utility>` is now unused and can be removed with it

Fixed in 3fde99d1.

atsushi421 (approved):

PR #1504, opened 2026-08-07
perf(agnocast_kmod): precompute each publisher's notify list at registration

Koichi98:
## Description

#1432 `kmalloc_array`s the eventfd context array inside `publish_msg`, while holding `topic_rwsem` (write) and `global_htables_rwsem` (read). Allocating on the publish path is the problem this PR removes: it can sleep and reclaim with those locks held, so an unlucky publisher stalls every other publisher and subscriber on the topic. The extra hash walk per publish goes away as a side effect.

None of the take / domain / local-publication filters depends on the message, so the answer is the same for every publish until an endpoint registers or leaves. Each `publisher_info` now owns a `notify_ctxs` array rebuilt at registration, and `publish_msg` only reads it — no allocation, no walk.

- `rebuild_all_notify_lists()` runs from `insert_subscriber_info()` / `insert_publisher_info()` under `global_htables_rwsem` (write).
- `agnocast_unlink_subscriber_info()` becomes the single place a subscriber leaves a live topic, so releasing its eventfd context and rebuilding the lists that pointed at it cannot be honored at one call site and forgotten at another.
- The array only grows, geometrically from `NOTIFY_CTXS_MIN_CAPACITY`, and never during publish.

Split out of #1432 so the notification mechanism and this optimization can be reviewed apart.

## Related links

- Stacked on #1432

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Post-merge checklist

`need-patch-update` — internal only, no ABI change.

### DIFF
--- a/agnocast_kmod/agnocast_internal.c
+++ b/agnocast_kmod/agnocast_internal.c
@@ -35,8 +35,7 @@ static void pre_handler_subscriber_exit(struct topic_wrapper * wrapper, const pi
 
     const topic_local_id_t subscriber_id = sub_info->id;
 
-    hash_del(&sub_info->node);
-    free_subscriber_info(sub_info);
+    agnocast_unlink_subscriber_info(wrapper, sub_info);
 
     if (subscriber_id < 0 || subscriber_id >= MAX_TOPIC_LOCAL_ID) {
       dev_warn(
--- a/agnocast_kmod/agnocast_internal.c
+++ b/agnocast_kmod/agnocast_internal.c
@@ -26,6 +26,9 @@ struct tracepoint * tp_sched_process_exit;
 
 static void pre_handler_subscriber_exit(struct topic_wrapper * wrapper, const pid_t pid)
 {
+  // Unlinked as a batch so the notify lists are rebuilt once instead of once per subscriber, which
+  // would repeat the same work and land on the same result.
+  HLIST_HEAD(leaving);
   struct subscriber_info * sub_info;
   int bkt_sub_info;
   struct hlist_node * tmp_sub_info;
@@ -33,9 +36,22 @@ static void pre_handler_subscriber_exit(struct topic_wrapper * wrapper, const pi
   {
     if (sub_info->pid != pid) continue;
 
+    hash_del(&sub_info->node);
+    hlist_add_head(&sub_info->node, &leaving);
+  }
+
+  if (hlist_empty(&leaving)) return;
+
+  // Before the frees below, so that no list is left pointing at a released context.
+  agnocast_rebuild_notify_lists(wrapper);
+
+  struct hlist_node * tmp_leaving;
+  hlist_for_each_entry_safe(sub_info, tmp_leaving, &leaving, node)
+  {
     const topic_local_id_t subscriber_id = sub_info->id;
 
-    agnocast_unlink_subscriber_info(wrapper, sub_info);
+    hlist_del(&sub_info->node);
+    free_subscriber_info(sub_info);
 
     if (subscriber_id < 0 || subscriber_id >= MAX_TOPIC_LOCAL_ID) {
       dev_warn(
--- a/agnocast_kmod/agnocast_internal.h
+++ b/agnocast_kmod/agnocast_internal.h
@@ -290,6 +290,10 @@ void agnocast_release_topic_wrapper(struct topic_wrapper * wrapper);
 void agnocast_unlink_subscriber_info(
   struct topic_wrapper * wrapper, struct subscriber_info * sub_info);
 
+// Recomputes the publishers' notify lists, for the one caller that unlinks subscribers in bulk and
+// so cannot use agnocast_unlink_subscriber_info(). Caller holds global_htables_rwsem (write).
+void agnocast_rebuild_notify_lists(struct topic_wrapper * wrapper);
+
 long agnocast_ioctl(struct file * file, unsigned int cmd, unsigned long arg);
 
 // Ring buffer to hold exited pids.
--- a/agnocast_kmod/agnocast_internal.h
+++ b/agnocast_kmod/agnocast_internal.h
@@ -113,7 +113,9 @@ struct publisher_info
   // reads it.
   struct eventfd_ctx ** notify_ctxs;
   uint32_t notify_num;
-  uint32_t notify_capacity;  // Always the whole subscriber count, so only a join can grow it.
+  // Never below the topic's notifiable subscriber count, and never shrinks, so only a join can grow
+  // it and every other rebuild is allocation-free.
+  uint32_t notify_capacity;
   struct hlist_node node;
 };
 
--- a/agnocast_kmod/agnocast_internal.h
+++ b/agnocast_kmod/agnocast_internal.h
@@ -48,6 +48,9 @@ extern struct rw_semaphore global_htables_rwsem;
 // At most one agent per (IPC namespace, domain), so the table is tiny.
 #define DISCOVERY_AGENT_HASH_BITS 4
 
+// Covers typical ROS 2 fan-out without reallocation, at a negligible per-publisher cost.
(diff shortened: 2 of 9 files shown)


agnocast_kmod/agnocast_ioctl.c
```diff
@@ -266,6 +349,19 @@ static int insert_subscriber_info(
   uint32_t hash_val = hash_min(new_id, SUB_INFO_HASH_BITS);
   hash_add(wrapper->topic->sub_info_htable, &(*new_info)->node, hash_val);
 
+  int ret = rebuild_all_notify_lists(wrapper);
+  if (ret < 0) {
```
atsushi421: This error path is the first failure point after current_pubsub_id++ (and current_entry_id++ for volatile subscribers), and it does not roll the counters back, so every failed ADD_SUBSCRIBER permanently burns an id out of the MAX_TOPIC_LOCAL_ID space; once it is exhausted the topic rejects all new endpoints with -ENOSPC. Consider decrementing the counters here (and in insert_publisher_info's new error path), which is safe because global_htables_rwsem is held for write.

agnocast_kmod/agnocast_internal.h
```diff
@@ -105,11 +108,18 @@ struct publisher_info
   bool qos_is_transient_local;
   uint32_t entries_num;
   bool is_bridge;
+  // The eventfd contexts PUBLISH signals, in no particular order. Membership depends only on the
+  // endpoints, never on the message, so the list is built as endpoints register and publish only
+  // reads it.
+  struct eventfd_ctx ** notify_ctxs;
+  uint32_t notify_num;
+  uint32_t notify_capacity;  // Always the whole subscriber count, so only a join can grow it.
```
atsushi421: This reads as notify_capacity == subscriber count, but the actual invariant is >=: the NOTIFY_CTXS_MIN_CAPACITY floor and geometric doubling overshoot, and capacity never shrinks on removal. Since the infallibility arguments in agnocast_ioctl.c lean on this invariant, consider rewording to match rebuild_notify_list()'s accurate "keeps capacity >= the topic's subscriber count".

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -204,12 +204,95 @@ static struct subscriber_info * find_subscriber_info(
...
+// Recomputes which subscribers this publisher notifies. None of the filters below depends on the
+// message, so evaluating them here leaves publish with nothing to do but signal.
+static int rebuild_notify_list(
+  struct topic_wrapper * wrapper, struct publisher_info * pub_info, const uint32_t sub_num)
+{
+  // The subscriber count bounds the list, so one reservation covers the fill below. Reserving for
+  // the whole count (not just the matches) keeps capacity >= the topic's subscriber count for
+  // every publisher, which is what makes every rebuild but a subscriber's insertion allocation-
+  // free, and so infallible.
+  int ret = reserve_notify_ctxs(pub_info, sub_num);
```
atsushi421: Sizing by the total subscriber count includes take subs, which the fill loop below skips, so a topic with many take subs leaves every publisher holding a large mostly-empty array, and take-sub join/leave still triggers full rebuilds whose results are byte-identical. Consider counting only subscribers with a non-NULL notify_ctx here, and early-returning in agnocast_unlink_subscriber_info() when sub_info->notify_ctx is NULL; the infallibility argument carries over unchanged.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -357,10 +453,22 @@ static int insert_publisher_info(
...
+  (*new_info)->notify_ctxs = NULL;
+  (*new_info)->notify_num = 0;
+  (*new_info)->notify_capacity = 0;
   INIT_HLIST_NODE(&(*new_info)->node);
   uint32_t hash_val = hash_min(new_id, PUB_INFO_HASH_BITS);
   hash_add(wrapper->topic->pub_info_htable, &(*new_info)->node, hash_val);
 
+  // Later subscribers are folded in as they register. Failing here beats failing in publish,
+  // which cannot report it.
+  int ret = rebuild_notify_list(wrapper, *new_info, agnocast_get_size_sub_info_htable(wrapper));
```
atsushi421: This rebuild only walks the subscriber table and reads pub_info fields that are already set, so it does not need the publisher to be in pub_info_htable yet. Building the list before hash_add would reduce the error path to free_publisher_info() with no hash_del undo.

agnocast_kmod/agnocast_internal.c
```diff
@@ -35,8 +35,7 @@ static void pre_handler_subscriber_exit(struct topic_wrapper * wrapper, const pi
 
     const topic_local_id_t subscriber_id = sub_info->id;
 
-    hash_del(&sub_info->node);
-    free_subscriber_info(sub_info);
+    agnocast_unlink_subscriber_info(wrapper, sub_info);
```
atsushi421: Calling the unlink inside this loop triggers a full rebuild of every publisher's list once per removed subscriber, so a process holding K subscribers on one topic rebuilds K times on exit where a single rebuild after removing all of them gives the same result. Consider hash_del-ing all matching subscribers first, rebuilding once, then freeing them.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -266,6 +349,19 @@ static int insert_subscriber_info(
   uint32_t hash_val = hash_min(new_id, SUB_INFO_HASH_BITS);
   hash_add(wrapper->topic->sub_info_htable, &(*new_info)->node, hash_val);
 
+  int ret = rebuild_all_notify_lists(wrapper);
+  if (ret < 0) {
+    hash_del(&(*new_info)->node);
+    // The publishers rebuilt before the failing one already hold this subscriber's notify_ctx, so
+    // they have to be rebuilt again before it is released. Cannot fail: the unlink above put the
+    // subscriber count back, so no list has to grow.
+    WARN_ON_ONCE(rebuild_all_notify_lists(wrapper) < 0);
```
atsushi421: This second full rebuild and its infallibility argument are only needed because reserve_notify_ctxs() discards the old contents, so the failed first rebuild leaves earlier publishers holding the new subscriber's ctx. If reserve preserved contents (alloc, copy, free), all growth could happen in a fallible pre-pass before hash_add, after which a single rebuild is infallible and this undo path disappears.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in 23d90e3c. The reservation is now the last failure point and sits ahead of the counters, so a failed ADD_SUBSCRIBER no longer burns an id.

agnocast_kmod/agnocast_internal.h
Koichi98: Fixed in d408ff6d.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in 7d9d3ec7.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in fa1198ff. Moved ahead of hash_add, so the error path is just free_publisher_info().

agnocast_kmod/agnocast_internal.c
Koichi98: Fixed in 017e2b56. pre_handler_subscriber_exit() now unlinks the process's subscribers as a batch and rebuilds once.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in 23d90e3c. reserve_notify_ctxs() now preserves contents, so the growth is a fallible pre-pass and the undo path is gone.

atsushi421 (approved):

PR #1506, opened 2026-08-07
feat(agnocastlib): document take() semantics and deprecate PollingSubscriber entry points

kobayu858:
## Description

`take()` and `take_data()` were documented as returning the most recent message, but with a history
depth greater than 1 they return the oldest entry still retained in the window. The returned message
lags the newest one by exactly `depth - 1` and the lag does not recover (measured: depth 1 → 0,
depth 3 → 2, depth 10 → 9), and the same entry is returned repeatedly until it is evicted. Depth 1
is the intended usage and the only depth where the two coincide.

This PR documents the actual behaviour rather than changing it, and deprecates the affected API.

- Docs: `TakeSubscription::take()` now describes both modes as implemented (one message per call,
  non-destructive, `true` → oldest in the window, `false` → oldest not yet received). Notes added on
  `agnocast_ioctl_take_msg()` and `PollingSubscriber`.
- `allow_same_message = true` is an argument value, so it cannot carry an attribute;
  `TakeSubscription::take()` emits `RCLCPP_WARN_ONCE` instead. `take(false)` is unaffected.

`PollingSubscriber` reproduces Autoware's polling subscriber, an Autoware-specific API that agnocast
does not intend to maintain as public API. It is planned to move to `autoware_agnocast_wrapper`, and
`allow_same_message` — which exists only for it — is expected to go with it, leaving `take()` with a
single FIFO meaning. Rejecting `allow_same_message = true` with depth > 1 was considered and
deliberately not added, since the API is expected to be removed rather than guarded. No removal
version is stated, as it is not decided yet.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

The change adds comments only, so no runtime behaviour can be affected. 

## Post-merge checklist

**Important notes:**


agnocast_kmod/agnocast_ioctl.c
```diff
@@ -1124,6 +1124,10 @@ int agnocast_ioctl_take_msg(
   ioctl_ret->ret_addr = 0;
   ioctl_ret->ret_entry_id = -1;
 
+  // allow_same_message exists only to serve agnocast::PollingSubscriber, which reproduces
```
Koichi98: This is placed above `searched_count` rather than the `allow_same_message` branch, and the same note already appears in agnocastlib. I don't think it is needed here.

src/agnocastlib/include/agnocast/agnocast.hpp
```diff
@@ -199,8 +199,14 @@ typename Subscription<MessageT>::SharedPtr create_subscription(
 /// @param topic_name Topic name.
 /// @param qos_history_depth History depth for the QoS profile.
 /// @return Shared pointer to the created polling subscription.
+/// @note agnocast::PollingSubscriber is planned to move to autoware_agnocast_wrapper and be removed
```
Koichi98: This @note is repeated 6 times (agnocast.hpp:202, :227; agnocast_node.hpp:510, :530; agnocast_subscription.hpp:363, :449). Suggest keeping only the class doc at :449.

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
```diff
@@ -346,15 +346,35 @@ class TakeSubscription : public SubscriptionBase
...
+   * regardless of the history depth; the depth bounds how far back the search goes.
+   *
+   * @param allow_same_message  If true, returns the oldest entry within the subscription's
+   *                            history depth, and may return the same message as the previous
+   *                            call (useful for always having the latest value with depth 1).
+   *                            If false, returns the oldest entry not yet received by this
+   *                            subscriber, i.e. FIFO.
    * @return Shared pointer to the message, or empty if unavailable.
+   *
+   * @note `allow_same_message = true` exists only to serve `agnocast::PollingSubscriber`, which is
```
Koichi98: The only caller passing `allow_same_message = true` is `AgnocastPollingSubscriber` in `autoware_agnocast_wrapper`, which holds a `TakeSubscription` directly and never goes through `agnocast::PollingSubscriber`, so removing the latter would not free this mode.

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
```diff
@@ -346,15 +346,35 @@ class TakeSubscription : public SubscriptionBase
...
+   *
+   * Reads are non-destructive: entries stay in shared memory and a per-subscriber watermark
+   * tracks how far this subscriber has read. Exactly one message is returned per call
+   * regardless of the history depth; the depth bounds how far back the search goes.
+   *
+   * @param allow_same_message  If true, returns the oldest entry within the subscription's
+   *                            history depth, and may return the same message as the previous
+   *                            call (useful for always having the latest value with depth 1).
+   *                            If false, returns the oldest entry not yet received by this
+   *                            subscriber, i.e. FIFO.
```
Koichi98: Not quite FIFO: the walk starts at `rb_last` and goes back only `qos_depth` entries, so a subscriber that falls behind by more than the depth has entries skipped silently.

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
```diff
@@ -421,11 +441,14 @@ class TakeSubscription : public SubscriptionBase
 /**
  * @brief Agnocast polling subscriber for a compile-time known message type.
  *
- * Wraps TakeSubscription<MessageT> and exposes a simple take_data() API
- * that always returns the most recent message (or an empty pointer if nothing
- * has been published yet).
+ * Wraps TakeSubscription<MessageT> and exposes a simple take_data() API that returns the most
```
Koichi98: This still says "the most recent message", which contradicts the `@note` on `take_data()` below (lags by `depth - 1`).

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
```diff
@@ -455,12 +478,25 @@ class PollingSubscriber
   };
 
   /// @deprecated Use take_data() instead.
-  const agnocast::ipc_shared_ptr<const MessageT> takeData() { return subscriber_->take(true); };
-  /// @brief Retrieve the latest message. Always returns the most recent message even if already
-  /// retrieved. Returns an empty pointer if no message has been published yet.
+  [[deprecated("Use take_data() instead.")]]
+  const agnocast::ipc_shared_ptr<const MessageT> takeData()
```
Koichi98, line 481: `takeData()` calls the same `take(true)` but lacks the depth note added to `take_data()`.

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
```diff
@@ -455,12 +478,25 @@ class PollingSubscriber
...
+  {
+    return subscriber_->take(true);
+  };
+  /// @brief Retrieve the latest message, returning it again on subsequent calls if nothing newer
+  /// has been published. Returns an empty pointer if no message has been published yet.
+  /// @note Assumes a history depth of 1. With a greater depth the returned message lags the newest
```
Koichi98: `up to` understates it: once the window is full the lag is exactly `depth - 1`, as the PR description says. Also `take()` itself never mentions the lag, so the cross-reference does not lead to an answer.

agnocast_kmod/agnocast_ioctl.c
kobayu858: fixed in https://github.com/autowarefoundation/agnocast/pull/1506/commits/7f7891d437ef871a8bceab21f6a20b0b856c7308

src/agnocastlib/include/agnocast/agnocast.hpp
kobayu858: fixed in https://github.com/autowarefoundation/agnocast/pull/1506/commits/7f7891d437ef871a8bceab21f6a20b0b856c7308

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
kobayu858: fixed in https://github.com/autowarefoundation/agnocast/pull/1506/commits/a372dd54ef4ba95191930de88f478e8e2408377a

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
kobayu858: fixed in https://github.com/autowarefoundation/agnocast/pull/1506/commits/1bf6fe0d61b8e657f9e49d3c614a0227c7355971

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
kobayu858: fixed in https://github.com/autowarefoundation/agnocast/pull/1506/commits/fc6a9b9aced05e6419735de986bc82ca5fb63514

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
kobayu858: fixed in https://github.com/autowarefoundation/agnocast/pull/1506/commits/fc6a9b9aced05e6419735de986bc82ca5fb63514

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
kobayu858: fixed in https://github.com/autowarefoundation/agnocast/pull/1506/commits/fc6a9b9aced05e6419735de986bc82ca5fb63514

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
Koichi98, line 1: On the `TakeSubscription` class doc at lines 273-274 (outside the diff, so commenting at file level):

Same "latest message" wording that was fixed at :444 — with depth > 1, `take()` returns the oldest entry in the window, so this contradicts the `@param` text below. It also recommends `PollingSubscriber`, which this PR deprecates.

kobayu858: @Koichi98 
Thanks for pointing that out. I fixed it in https://github.com/autowarefoundation/agnocast/pull/1506/commits/3fc5391e30a7949f9ae92c0cd5690b6e06ea6e7e

Koichi98 (approved):

Issue #1501, opened 2026-08-06
Bridge UDS bind() fails with EADDRINUSE on consecutive process launches


bdm-k:
### Problem Description

If you repeatedly launch a set of Agnocast applications in rapid succession, you're likely to get the following error message:
```
[ERROR] [1785979946.864548602] [Agnocast]: BridgeManager crashed: bridge UDS bind() failed: Address already in use
```
This indicates that a new bridge manager was launched before the previous bridge manager closed the UDS, thereby releasing the (abstract) socket address.

### Root Cause

There is a gap between when the kernel module logically removes the bridge manager in `agnocast_ioctl_check_and_request_bridge_shutdown()` and when the bridge manager actually closes the UDS during `PerformanceBridgeManager` destruction. Therefore the following sequence of events is possible:

1. The kmod logically removes the bridge manager.
2. A new Agnocast process arrives, which the kmod tells to spawn a new bridge manager.
3. The new bridge manager ties `bind()` to find the abstract socket address is already in use.
4. The old bridge manager closes the UDS to release the socket address.

The race window can be up to about **2.5 seconds**. First, the **worker thread** polls the kmod to determine whether shutdown should be initiated. This check is performed once every `WORKER_TIMEOUT`, which is currently 1 second. Second, the **event loop thread** is blocked in `epoll_wait()` with a timeout of `EVENT_LOOP_TIMEOUT_MS`, which is currently 1000. It must wait for the timeout to expire before checking whether a shutdown has been initiated. Taking these two timeouts and the associated overhead into account, the race window can be up to about 2.5 seconds.

### Mitigation

The second cause is relatively easy to mitigate. The event loop thread should be woken immediately when shutdown is requested, rather than waiting for the next timeout iteration. One possible solution could be to define a dedicated signal for shutdown notification and add it to the set of signals to block on `signalfd`.

> [!NOTE]
> The worker thread is woken immediately on shutdown requests via `pending_msgs_cv_.notify_all()` in `request_shutdown()`. The executor thread is also cancelled immediately via `executor_->cancel()` in `request_shutdown()`. Only the event loop thread does not have such a mechanism.

PR #1507, opened 2026-08-07
fix(bridge): immediately notify shutdown to event loop thread

bdm-k:
## Description

Problem: #1501 

This PR eliminates the second layer contributing to the race window during shutdown. Currently, when `request_shutdown()` is called, the event loop thread may remain blocked in `epoll_wait()` until its timeout expires, delaying shutdown response. This PR introduces an **eventfd**-based wakeup mechanism so the event loop thread is unblocked immediately.

Specifically, a non-blocking eventfd is created in `IpcEventLoopBase` and registered with epoll. `request_shutdown()` calls `event_loop_.wakeup()` after setting `shutdown_requested_`, causing the event loop's `epoll_wait()` to return immediately so the run loop can observer the shutdown flag without waiting for a timeout.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Post-merge checklist


Koichi98:
@bdm-k 
Could we use an `eventfd` registered on the existing epoll set instead of a self-directed SIGUSR1?

`kill(getpid(), SIGUSR1)` is process-directed, so it only reaches the event loop's `signalfd` as long as every other thread blocks SIGUSR1 — which holds today, but nothing states it, and SIGUSR1's default action is `Term` if it ever doesn't. An eventfd carries no such invariant, and it matches how the rest of the repo already does cross-thread wakeups:　https://github.com/autowarefoundation/agnocast/blob/0cafe7496baea4f0ab268c71a47bbcf26c55e9b2/src/agnocastlib/src/node/agnocast_only_executor.cpp#L143-L151.

bdm-k:
added `wakeup()` to `IpcEventLoopBase` that another thread can call to manually wake up the event loop.

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_ipc_event_loop_base.hpp
```diff
@@ -133,6 +140,12 @@ inline bool IpcEventLoopBase::spin_once(int timeout_ms)
       if (s == sizeof(struct signalfd_siginfo)) {
         handle_signal();
       }
+    } else if (fd == wakeup_efd_) {
+      uint64_t val;
+      ssize_t s = read(wakeup_efd_, &val, sizeof(val));
+      if (s == -1 && errno != EINTR) {
```
Koichi98: The fd is `EFD_NONBLOCK`, so `EAGAIN`/`EWOULDBLOCK` is a normal outcome here, not a failure — as written it would warn on a healthy path. Could we exclude it?

Koichi98:
@bdm-k Thanks for the rework — the eventfd version looks good.

One thing before merge: the description still describes the SIGUSR1 approach (`SIGUSR1` in the blocked-signals list, the early return in `on_signal()`). Could you update it? It becomes the merge commit body.

src/agnocastlib/include/agnocast/bridge/agnocast_bridge_ipc_event_loop_base.hpp
bdm-k: fixed in [b075062](https://github.com/autowarefoundation/agnocast/pull/1507/commits/b07506253d81fb647b1137943c706c39557be14e)

Koichi98 (approved):

PR #1508, opened 2026-08-07
feat: GPU IPC with cuda dispatch API

sykwer:
## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Post-merge checklist

**Important notes:**


src/agnocast_gpu/include/agnocast/gpu/dispatch.hpp
```diff
@@ -0,0 +1,348 @@
...
+// the message a node is about to publish holds whatever was there before.
+inline bool check(cudaError_t status, const char * what)
+{
+  if (status == cudaSuccess) return true;
+  RCLCPP_ERROR(agnocast::logger, "%s failed: %s", what, cudaGetErrorString(status));
+  return false;
+}
+
+// One stream per thread. A callback never runs two dispatches at once, and the
+// callback-isolated executor gives each callback its own thread, so this is one
```
nishikawa-masaki, line 187: Is CIE mandatory for GPU IPC support?
How is it enforced?
I understand Agnocast and CIE are independent from each other.

src/agnocast_gpu/include/agnocast/gpu/message_types.hpp
```diff
@@ -0,0 +1,85 @@
...
+  }
+  if (reserved != 0 && product > reserved) return static_cast<size_t>(reserved);
+  return static_cast<size_t>(product);
+}
+
+}  // namespace detail
+
+inline size_t gpu_data_size(const PointCloud2 & msg)
+{
+  return detail::bounded_extent({msg.width, msg.point_step, msg.height}, msg.data.size());
```
nishikawa-masaki, line 77: This line should be:
`return detail::bounded_extent({msg.height, msg.row_step}, msg.data.size());`

If `msg->is_dence` is `false`, `row_step` can be greater than `point_step * width`.

`row_step >= point_step * width`

src/agnocast_gpu/test/gpu_region_e2e.cpp
```diff
@@ -0,0 +1,277 @@
...
+    }
+    if (pool->available() != kSlotCount - 1) return 1;
+
+    // Non-const: get() on a const handle yields a const pointer, because a
+    // subscriber's mapping of the region is read-only.
+    agnocast::internal::gpu_array<uint8_t> data(
+      pool->region_id(), slot_index, kPayload, publisher_id);
+
+    std::vector<uint8_t> pattern(kPayload);
+    for (size_t i = 0; i < kPayload; i++) pattern[i] = static_cast<uint8_t>(i % 251);
```
nishikawa-masaki, line 152: `251` is a "magic number".
Define a constexpr with easy-to-understand name.

src/agnocast_gpu/test/gpu_region_e2e.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+  std::vector<uint8_t> readback(kPayload, 0);
+  if (cudaMemcpy(readback.data(), data.get(), kPayload, cudaMemcpyDeviceToHost) != cudaSuccess) {
+    std::fprintf(stderr, "subscriber cudaMemcpy failed\n");
+    return 1;
+  }
+
+  size_t mismatches = 0;
+  for (size_t i = 0; i < kPayload; i++) {
+    if (readback[i] != static_cast<uint8_t>(i % 251)) mismatches++;
```
nishikawa-masaki, line 240: `251` is a "magic number".
Define a constexpr with easy-to-understand name.

src/agnocast_gpu/test/test_gpu_backend.cpp
```diff
@@ -0,0 +1,280 @@
...
+{
+  auto * backend = gpu_backend_or_skip();
+  if (backend == nullptr) GTEST_SKIP() << "no VMM-capable GPU";
+
+  constexpr uint32_t kSlotSize = 1U << 20;
+  const MappedGpuRegion region = backend->create_region(kSlotSize, 2);
+  ASSERT_TRUE(region.valid());
+
+  std::vector<uint8_t> pattern(kPayload);
+  for (size_t i = 0; i < pattern.size(); i++) pattern[i] = static_cast<uint8_t>(i % 251);
```
nishikawa-masaki, line 196: `251` is a "magic number".
Define a constexpr with easy-to-understand name.

src/agnocast_sample_application/src/gpu_listen_talker.cpp
```diff
@@ -0,0 +1,110 @@
...
+using agnocast::gpu::uploads;
+using agnocast::gpu::writes;
+
+__global__ void filter_kernel(
+  const uint8_t * in, uint8_t * out, size_t size, const float * gain, uint32_t * kept)
+{
+  const size_t i = blockIdx.x * blockDim.x + threadIdx.x;
+  if (i < size) {
+    const uint32_t scaled = in[i] * static_cast<uint32_t>(gain[0]);
+    out[i] = static_cast<uint8_t>(scaled % 251);
```
nishikawa-masaki, line 28: `251` is a "magic number".
Define a constexpr with easy-to-understand name.

src/agnocast_sample_application/src/gpu_talker.cpp
```diff
@@ -0,0 +1,106 @@
...
+constexpr size_t kCapacity = static_cast<size_t>(kWidth) * kPointStep;
+}  // namespace
+
+__global__ void fill_kernel(
+  uint8_t * points, size_t size, uint8_t seq, const float * transform, const float * calibration)
+{
+  const size_t i = blockIdx.x * blockDim.x + threadIdx.x;
+  if (i < size) {
+    const size_t offset = static_cast<size_t>(transform[0]) + static_cast<size_t>(calibration[0]);
+    points[i] = static_cast<uint8_t>((i + seq + offset) % 251);
```
nishikawa-masaki, line 33: `251` is a "magic number".
Define a constexpr with easy-to-understand name.

src/agnocast_sample_application/src/gpu_talker.cpp
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    cloud->header.frame_id = "lidar_" + std::to_string(getpid());
+    cloud->height = 1;
+    cloud->width = kWidth;
+    cloud->point_step = kPointStep;
+    cloud->row_step = kWidth * kPointStep;
+    cloud->is_dense = true;
+
+    const auto seq = static_cast<uint8_t>(seq_ % 251);
```
nishikawa-masaki, line 72: `251` is a "magic number".
Define a constexpr with easy-to-understand name.

nishikawa-masaki (changes requested):
Though there are a lot of code lines to be read, I submit my review at this point.

src/agnocast_sample_application/src/gpu_talker.cpp
```diff
@@ -0,0 +1,106 @@
...
+class GpuTalker : public agnocast::Node
+{
+public:
+  GpuTalker() : Node("gpu_talker")
+  {
+    publisher_ = create_publisher<agnocast::gpu::PointCloud2>("/gpu_points", 4);
+
+    // Page-locked so the uploads are genuinely asynchronous; dispatch() warns
+    // when they are not. One buffer holds both parameter blocks.
+    cudaMallocHost(&parameters_, 8 * sizeof(float));
```
nishikawa-masaki, line 46: The host memory allocated here is not freed explicitly.
It will be freed at the termination of the main(), but It should be freed in the destructor of GpuTalker, considering that this is a sample code referenced by the users.

src/agnocast_sample_application/src/gpu_talker.cpp
```diff
@@ -0,0 +1,106 @@
...
+    cloud->is_dense = true;
+
+    const auto seq = static_cast<uint8_t>(seq_ % 251);
+
+    // The device buffers start null and are allocated on the dispatch stream, so
+    // no synchronous allocation happens inside the callback. Dropping the borrow
+    // returns its slot, so a failed fill costs one frame.
+    const bool filled = dispatch(
+      uploads(transform_, parameters_, 4, TransferOptions::kAllocateDeviceAsync),
+      uploads(calibration_, parameters_ + 4, 4, TransferOptions::kAllocateDeviceAsync),
```
nishikawa-masaki, line 79: The GPU device memory allocated to transform_ and calibration_ are not freed explicitly.
They will be freed at the termination of the main(), but Ithey should be freed in the destructor of GpuTalker, considering that this is a sample code referenced by the users.

src/agnocast_sample_application/src/gpu_talker.cpp
```diff
@@ -0,0 +1,106 @@
...
+  float * parameters_ = nullptr;
+  float * transform_ = nullptr;
+  float * calibration_ = nullptr;
+  int64_t seq_ = 0;
+};
+
+int main(int argc, char ** argv)
+{
+  agnocast::init(argc, argv);
+  agnocast::AgnocastOnlySingleThreadedExecutor executor;
```
nishikawa-masaki, line 101: Does the sample work correctly if AgnocastOnlyMultiThreadExecutor is used?
Claude Code says:
*  It still works with AgnocastOnlyMultiThreadExecutor, but that depends on the hidden implementation:
    * The default callback group serializes the timer.
    * dispatch() synchronizes before returning.
* As a sample, the codes need to be more explicit.
    * Use thread-local members to avoid race conditions at the cost of memory allocation for each thread.
    ```
    static thread_local float * transform_ = nullptr;
    static thread_local float * calibration_ = nullptr;
    ```
    * Or, add comments in the sample codes.

src/agnocast_gpu/include/agnocast/gpu/dispatch.hpp
```diff
@@ -0,0 +1,440 @@
...
+  // driver call it makes allocates host memory of its own.
+  auto drain = rcpputils::make_scope_exit([s]() noexcept {
+    const agnocast::internal::SuspendedBorrowWindow drain_suspended;
+    static_cast<void>(cudaStreamSynchronize(s));
+  });
+
+  // Folded so that every declaration is prepared even once one has failed:
+  // allocations made here are the caller's from then on.
+  bool ready = true;
+  ((ready = prepare(std::get<I>(parts), s) && ready), ...);
```
nishikawa-masaki, line 382: Another possible implementation that emphasizes propagation of failure.
```
((ready &= prepare(std::get<I>(parts), s)), ...);
```

src/agnocast_gpu/include/agnocast/gpu/dispatch.hpp
```diff
@@ -0,0 +1,440 @@
...
+  if (!ready) return false;
+
+  bool ok = true;
+  ((ok = issue_upload(std::get<I>(parts), s) && ok), ...);
```
nishikawa-masaki, line 386: Another possible implementation that emphasizes propagation of failure.
```
((ok &= issue_upload(std::get<I>(parts), s)), ...);
```
This pattern appears repeatedly.

src/agnocast_gpu/include/agnocast/gpu/dispatch.hpp
```diff
@@ -0,0 +1,440 @@
...
+inline bool prepare(const Download & d, cudaStream_t s)
+{
+  return allocate_if_requested(d.device_src, d.bytes, d.options, s);
+}
+
+template <typename T>
+bool issue_upload(const T &, cudaStream_t)
+{
+  return true;
+}
```
nishikawa-masaki, line 317: This "catch-all" template can be a trap for the future developers.
The overider "issue_upload(const Upload & u, ...)" doesn't work for the classes derived from "Upload".

The same applies for the template for issue_download().

src/agnocast_gpu/include/agnocast/gpu/dispatch.hpp
```diff
@@ -0,0 +1,440 @@
...
+  // The callable is the user's, and may well fill in message fields, so it runs
+  // with the window as it found it.
+  suspended.reset();
+  std::get<kWork>(parts)(s);
+  ok = check(cudaGetLastError(), "work submission") && ok;
+
+  // Reopened for the driver calls below, which allocate host memory of their own.
+  suspended.emplace();
+  ok = check(cudaEventRecord(done, s), "event record") && ok;
+  ok = check(cudaEventSynchronize(done), "event synchronize") && ok;
```
nishikawa-masaki, line 410: `cudaEventSynchronize()` makes the **CPU** thread wait for completion of the GPU kernel. The message is published after the finish of the GPU kernel.
I'm afraid that this is not what the perception team wants to have. They want to have "GPU works on-the-fly", the GPU can keep working on the shared memory in the message while the message is published by the publisher and received by the subscriber(s).
(Synchronization is done only on the GPU)

Is this intentional restriction for the time being?

I investegated on this. "GPU works on-the-fly" is possible, but not easy.
* cudaEvent is a "binary semaphore", therefore it can't be used if there are N subsribers.
* So-called "timeline semaphore" can be used for N subscribers.
  * CUDA can import "timeline semaphore" with `cudaImportExternalSemaphore()`, but CUDA itself can't create such a semaphore.
    * Need Vulkan or NvSciSync for creating it. It means extra dependency.
* Using CUDA driver API (`cuStreamWriteValue32()` and `cuStreamWaitValue32()`), it is possible to emulate "timeline semaphore.
    * CUDA driver API is used in addition to CUDA runtime API.
    * `cuStreamWaitValue32()` may fall into eternal waiting if the publisher dies.

PR #1510, opened 2026-08-07
feat(agnocastlib): add AsyncParametersClient to Node

bdm-k:
## Description

This PR introduces `agnocast::AsyncParametersClient`, which is essentially `rclcpp::AsyncParametersClient` for `agnocast::Node`. It lets an `agnocast::Node` get and set another node's parameters in coordination with `agnocast::ParameterService` over the Agnocast service layer.

In the constructor of `AsyncParametersClient`, six `Client` instances are created, one for each parameter service endpoint. The class provides utility methods to send requests to the parameter services.

### API

| Method | Purpose |
| ------ | ------- |
| `get_parameters(names, callback)` | Get parameters |
| `get_parameter_types(names, callback)` | Get parameter types |
| `set_parameters(params, callback)` | Set parameters one-by-one |
| `set_parameters_atomically(params, callback)` | Set parameters as an all-or-nothing operation |
| `describe_parameters(names, callback)` | Get parameter descriptors |
| `list_parameters(prefixes, depth, callback)` | List parameters |
| `service_is_ready()` | Check if all the 6 endpoints are available |
| `wait_for_service(timeout)` | Block until all endpoints are ready |

Some other methods that `rclcpp::AsyncParametersClient` provides are not yet supported. 

### Integration test

Two integration test suites (`ParameterServiceTest` and `AsyncParametersClientTest`) are added. `ParameterServiceTest` exercises all six service endpoints under certain configurations, using a real `AsyncParametersClient`. `AsyncParametersClientTest` has three tests, each asserting some aspect of `AsyncParametersClient`'s behavior:

1. An empty remote node name resolves to the owning node itself.
2. `wait_for_service()` times out when the target node has no parameter services running.
3. The completion callback is invoked once the response arrives.

## How was this PR tested?

- [ ] Autoware
- [x] sample application

## Notes for reviewers

Should we note `AsyncParametersClient` in `docs/agnocast_node_interface_comparison.md`?

## Post-merge checklist


src/agnocast_sample_application/src/no_rclcpp_client.cpp
```diff
@@ -18,6 +19,26 @@ int main(int argc, char * argv[])
...
+        node->get_logger(), "Interrupted while waiting for parameter services. Exiting.");
+      spin_thread.join();
+      agnocast::shutdown();
+      return 0;
+    }
+    RCLCPP_INFO(node->get_logger(), "Parameter services not available, waiting again...");
+  }
+  auto params_fut = params_client->get_parameters({"qos.depth"});
+  auto const & params = params_fut.get();
+  if (params.size() >= 1) {
```
Koichi98: If this is false, the sample prints nothing and moves on, so a failed fetch looks the same as a successful run. Could you add an `else` with a warning? Also `!params.empty()` reads better than `size() >= 1`.

src/agnocastlib/src/node/agnocast_parameter_client.cpp
```diff
@@ -0,0 +1,245 @@
...
+  auto request = get_parameters_client_->borrow_loaned_request();
+  request->names = names;
+
+  get_parameters_client_->async_send_request(
+    std::move(request),
+    [names, promise = std::move(promise), future, callback = std::move(callback)](
+      Client<rcl_interfaces::srv::GetParameters>::SharedFuture response_future) {
+      const auto & response = response_future.get();
+
+      const size_t count = std::min(names.size(), response->values.size());
```
Koichi98, line 53: The `std::min` is a necessary bound since the loop indexes both `names` and `values`. My question is what happens when it kicks in: the caller gets a shorter vector with no indication anything went wrong. This is reachable — the service returns an empty `values` when a name isn't declared. Would a warning log be worth it?

Koichi98:
> Should we note `AsyncParametersClient` in `docs/agnocast_node_interface_comparison.md`?

Yes, I think we should. §2.3 already has a `| Parameter Service | ✓ | **Full Support** | - | ... |` row, so adding an `AsyncParametersClient` row next to it would fit naturally.

src/agnocast_sample_application/src/no_rclcpp_client.cpp
```diff
@@ -18,6 +19,26 @@ int main(int argc, char * argv[])
...
+      RCLCPP_ERROR(
+        node->get_logger(), "Interrupted while waiting for parameter services. Exiting.");
+      spin_thread.join();
+      agnocast::shutdown();
+      return 0;
+    }
+    RCLCPP_INFO(node->get_logger(), "Parameter services not available, waiting again...");
+  }
+  auto params_fut = params_client->get_parameters({"qos.depth"});
+  auto const & params = params_fut.get();
```
Koichi98, line 36: I assume this follows the rclcpp sample, but `spin_until_future_complete()` isn't supported in agnocast yet, so that pattern isn't available here. Unlike it, `future.get()` can't be aborted on Ctrl-C — this hangs if the response never arrives.

Could you add a comment noting that, and that it should be switched over once `spin_until_future_complete()` is implemented?

src/agnocastlib/src/node/agnocast_parameter_client.cpp
bdm-k: There may be cases where `get_parameters` is used simply to check whether a parameter exists. In such cases, I think the log would just be unnecessary noise. The `rclcpp` implementation doesn’t do any logging here either, so I think this is something that the user should handle correctly.

src/agnocast_sample_application/src/no_rclcpp_client.cpp
bdm-k: I'm not really sure if I understand this. The executor is already spinning in another thread, so can we still use `spin_until_future_complete()` here?

src/agnocast_sample_application/src/no_rclcpp_client.cpp
bdm-k: fixed in [0923b42](https://github.com/autowarefoundation/agnocast/pull/1510/commits/0923b426f4f109c5622bc42b84795684cabeb89f)

src/agnocast_sample_application/src/no_rclcpp_client.cpp
Koichi98: My point was that with `spin_until_future_complete()`, this sample could drop the spin thread and follow the rclcpp tutorial's single-threaded shape. But for a service/client sample the current form is fine, and it's consistent with `minimal_client.cpp`. Please ignore — nothing to change.

src/agnocastlib/src/node/agnocast_parameter_client.cpp
Koichi98: Agreed, the logging isn't warranted. I was worried about a partial response misaligning `names[i]` with `values[i]`, but that can't happen — the service throws before pushing anything, so `values` is either full length or empty. Nothing to change.

src/agnocast_sample_application/src/no_rclcpp_client.cpp
```diff
@@ -18,6 +19,28 @@ int main(int argc, char * argv[])
...
+  if (!params.empty()) {
+    RCLCPP_INFO(node->get_logger(), "Server QoS depth: %ld", params[0].as_int());
+  } else {
+    RCLCPP_INTO(node->get_logger(), "Server QoS depth: N/A (parameter not declared)")
```
Koichi98: `RCLCPP_INTO` → `RCLCPP_INFO`. The trailing semicolon is missing too.

bdm-k: fixed

Koichi98 (dismissed):

Koichi98 (approved):

PR #1511, opened 2026-08-07
fix(agnocastlib): enable tracepoints by default

kobayu858:
## Description
`agnocast_tracepoint_wrapper.c` gated its LTTng provider on `TRACETOOLS_LTTNG_ENABLED`, which ros2_tracing removed in Jazzy and replaced with `TRACETOOLS_TRACEPOINTS_EXCLUDED`. The guard is therefore always false on Jazzy: no probes are generated and every `ros_trace_agnocast_*` function compiles down to an empty stub, so Agnocast tracepoints are silently never emitted and CARET cannot measure Agnocast paths.

This PR gates on `TRACETOOLS_TRACEPOINTS_EXCLUDED` instead, following the upstream `tracetools.c` pattern, so tracepoints are enabled by default and excluded only when explicitly requested. `LTTngUST` is already a `REQUIRED` dependency, so no CMake change is needed.

Switch behaviour after this change:

| | Humble | Jazzy |
|---|---|---|
| `TRACETOOLS_DISABLED` | honoured by the outer guard | honoured by the outer guard |
| `TRACETOOLS_TRACEPOINTS_EXCLUDED` | not defined by tracetools, so tracepoints are always compiled in | excludes tracepoints |

Always compiling tracepoints in on Humble is intentional. Agnocast defines its own LTTng provider and requires `LTTngUST` unconditionally, so it does not need to mirror tracetools' own LTTng detection result.

One known limitation: `TRACETOOLS_DISABLED` currently fails to build on both distributions for an unrelated pre-existing reason — `tracetools::get_symbol` is called without a guard, and it is only declared when tracing is compiled in. That is addressed in #1537.

## Related links
- #1537 — support building with `TRACETOOLS_DISABLED`

## How was this PR tested?

- [ ] Autoware
- [x] sample application
- E2E: ran `talker` + `no_rclcpp_take_listener` under an LTTng session and captured
  real events (`agnocast:agnocast_publish` ×79, `agnocast:agnocast_take` ×7, ...)

## Post-merge checklist

**Important notes:**


src/agnocastlib/src/agnocast_tracepoint_wrapper.c
```diff
@@ -4,7 +4,7 @@
 
 #ifndef TRACETOOLS_DISABLED
 
-#ifdef TRACETOOLS_LTTNG_ENABLED
+#ifndef TRACETOOLS_TRACEPOINTS_EXCLUDED
```
atsushi421, line 9: This repo also supports Humble, where tracetools never defines `TRACETOOLS_TRACEPOINTS_EXCLUDED` (its switch is still `TRACETOOLS_LTTNG_ENABLED`), so this `#ifndef` is always true there: the LTTng provider gets compiled unconditionally and Agnocast tracepoints fire even when tracetools/rclcpp tracepoints are compiled out, so "controlled by the same switches as rclcpp" only holds on Jazzy. Consider making the guard distro-aware (e.g. fall back to `#ifdef TRACETOOLS_LTTNG_ENABLED` when the Jazzy macro scheme is absent), or if always-on for Humble is intended, state that explicitly in the PR description and a code comment.

kobayu858: Thanks — you are right that Humble's tracetools has no `TRACETOOLS_TRACEPOINTS_EXCLUDED`, so this `#ifndef` is always true there and the provider is compiled unconditionally. That is intended, and I would rather not make the guard distro-aware.

`TRACETOOLS_LTTNG_ENABLED` is not a user-facing switch. tracetools' own CMake writes it into `tracetools/config.h` to record whether *tracetools* found lttng-ust at *its* build time.

The switch to standardise on for turning tracing off is `TRACETOOLS_DISABLED`, since it is the one knob that exists in both Humble and Jazzy and is already honoured by the outer guard in this file. Note it currently fails to build on *both* distributions for an unrelated pre-existing reason (`tracetools::get_symbol` is called without a guard, and it is only declared when tracing is compiled in); #1537 fixes that. Once #1537 is in, `TRACETOOLS_DISABLED` gives a single distro-independent way to disable Agnocast tracepoints.

I have added a code comment stating the intent, as you suggested, and a note in the PR description.

atsushi421 (approved):
Thanks for documenting the intent in the code comment and the PR description. The rationale makes sense: Agnocast has its own LTTng provider with LTTngUST as an unconditional dependency, so mirroring tracetools' own detection result is unnecessary, and TRACETOOLS_DISABLED (with #1537) remains the distro-independent off switch. LGTM.

PR #1513, opened 2026-08-09
refactor(kmod)[need-minor-update]: rename performance bridge manager to bridge manager

Koichi98:
## Description

The Standard Bridge is gone and `BridgeMode` is now just `{Off, On}`, so the "performance" qualifier no longer distinguishes anything. This drops it from the remaining names.

- ioctl fields: `is_performance_bridge_manager` → `is_bridge_manager`, `ret_performance_bridge_daemon_exist` → `ret_bridge_daemon_exist`
- C++ types: `PerformanceBridge{Manager,Loader,IpcEventLoop}` → `Bridge*`, `PerformancePubsubBridgeResult` → `PubsubBridgeResult`
- Flattened `bridge/performance/` into `bridge/` and dropped the `agnocast_performance_bridge_` filename prefix
- UDS status wire value: `{"type":"performance"}` → `{"type":"bridge"}`. The CLI reports both legacy values (`'standard'`, `'performance'`) with a dedicated "legacy bridge type, which is no longer supported" message rather than the generic unknown-type one.

## Related links

None.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Notes for reviewers

- `AGNOCAST_BRIDGE_MODE=performance` / `=standard` are intentionally kept — they are deprecated user-facing env values that still warn and fall back to ON.
- The bridge node name changes to `agnocast_bridge_node_<n>`, which is externally visible.

## Post-merge checklist

https://github.com/autowarefoundation/agnocast_doc/pull/51

`need-minor-update` (kmod/agnocastlib ioctl field names and the UDS wire value change). Label and title marker not applied yet.


agnocast_kmod/agnocast_kunit/agnocast_kunit_add_process.c
```diff
@@ -64,7 +64,7 @@ void test_case_add_process_twice(struct kunit * test)
   KUNIT_EXPECT_FALSE(test, agnocast_is_proc_exited(local_pid));
 }
 
-// A performance bridge manager is gated per-(ipc_ns, domain): a manager in one
+// A bridge manager is gated per-(ipc_ns, domain): a manager in one
```
atsushi421, line 67: This comment was updated to drop "performance", but the function it documents is still named `test_case_add_process_perf_manager_per_domain` (its declaration and KUNIT_CASE registration in `agnocast_kunit_add_process.h` are also untouched). Consider renaming it to `test_case_add_process_bridge_manager_per_domain` to complete the rename.

Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1513/commits/db370573a88eb0e5cab2133c408058a23fd3d6ee.

atsushi421 (approved):

sykwer (approved):

PR #1515, opened 2026-08-10
test(scripts): allow -p/--parallel in e2e_test_1to1.bash with the bridge on

Koichi98:
## Description

`e2e_test_1to1.bash` rejected `-p/--parallel` unless `AGNOCAST_BRIDGE_MODE=off`. The guard was added in #1402: the standard bridge created its ROS 2 endpoints inside the application process, so a worker's `ROS_DOMAIN_ID` also isolated bridged traffic, while the remaining bridge runs as a separate kmod-spawned process that was domain-unaware at the time.

Domain isolation has since landed (#1398, #1409, #1410, #1485). The kmod keys topics on `(topic_name, ipc_ns, domain_id)` and keys the bridge manager and the discovery agent on `(ipc_ns, domain_id)`, and `create_uds_addr_for_bridge()` appends a `_d<domain>` suffix. Each worker already gets its own `ROS_DOMAIN_ID` and its own topic name, so the guard is obsolete and this removes it.

## Related links

- #1092 introduced `-p/--parallel`
- #1402 added the guard

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

Additionally, `bash scripts/test/e2e_test_1to1.bash -p 32` exercises the path this PR unblocks, and `AGNOCAST_BRIDGE_MODE=off bash scripts/test/e2e_test_1to1.bash -p 32` covers the previously allowed one.

## Post-merge checklist


scripts/test/e2e_test_1to1.bash
```diff
@@ -78,12 +77,6 @@ else
     BRIDGE_OFF=false
 fi
 
-if [ "$PARALLEL" -gt 1 ] && [ "$BRIDGE_OFF" != true ]; then
```
atsushi421, line 81: With this guard removed, a parallel worker returns its domain-ID slot after only `sleep 2`, which can race the previous test's bridge-manager teardown on that domain: the manager keeps its abstract UDS address (`agnocast_bridge_manager_<ipcns>_d<domain>`) bound until its event loop and `rclcpp::shutdown()` complete, so the next test on the reused domain can fork a replacement manager that hits `bind()` EADDRINUSE and exits without retry, or deliver its one-shot bridge registrations to the dying manager, where they are discarded. The 2s margin was calibrated for sequential runs (expected max ~1s) and stretches under `-p 32`-level CPU contention, so this can become a flake source. Consider having the worker wait until the old manager for its domain is actually gone (e.g. poll `ss -x` for the domain-suffixed UDS address with a timeout) before returning the slot to the pool.

atsushi421 (commented):
Now that both arms of the bridge-mode check are reachable in parallel mode, the parallel branch's sweep setup (the `USE_AGNOCAST_PUB` selection, the five parameter arrays, and the `TOTAL_TESTS` computation) is an exact duplicate of the sequential branch's, so future sweep changes have to be made twice and can silently drift. Consider hoisting that setup above the single/parallel/sequential split so both modes share one copy.

scripts/test/e2e_test_1to1.bash
Koichi98: Thanks. I checked, and this can already happen on `main` — the kmod marks the bridge manager gone before it releases its socket. This PR just makes it more likely to show up. Fixing it properly isn't small, so I'll do it in a separate PR rather than patch around it here.

Koichi98:
Done — moved the setup above the split so both modes share one copy. No behavior change.

atsushi421 (approved):

PR #1516, opened 2026-08-10
docs(agnocast): correct message queue mentions left stale by earlier refactors

Koichi98:
## Description

Two message queue mentions no longer match the code:

- `docs/shared_memory.md` describes a per-process `/new_publisher@pid` message queue that notifies subscribers of a new publisher. That queue is gone; the kernel module now returns the new publisher's shm info on the subscriber's next `AGNOCAST_RECEIVE_MSG_CMD` / `AGNOCAST_TAKE_MSG_CMD`.
- `agnocast_ioctl.c` says the performance bridge manager is keyed by its MQ name. It has used a UDS address since #1433.

Documentation and comment only, no behavior change.

## Related links

- #1433 (bridge MQ replaced with UDS)

## How was this PR tested?

Not applicable: no executable code changed. `agnocast_kmod` still builds.

## Notes for reviewers

Both mentions were already stale before #1503; they are split out here to keep that PR scoped to what it makes obsolete itself.

## Post-merge checklist

`need-patch-update`

### DIFF
--- a/agnocast_kmod/agnocast_ioctl.c
+++ b/agnocast_kmod/agnocast_ioctl.c
@@ -627,8 +627,8 @@ int agnocast_ioctl_get_version(struct ioctl_get_version_args * ioctl_ret)
   return 0;
 }
 
-// A performance bridge manager is per-(ipc_ns, domain): its MQ name carries the
-// domain suffix, so each domain needs its own manager. Gate on the domain too,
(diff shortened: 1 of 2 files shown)


Koichi98:
Folded into #1503: strictly not caused by that PR, but closely enough related to keep in one place.

PR #1520, opened 2026-08-10
fix(kmod): read-lock get_exit_process phase 1 and stop the exit-flag copy from leaking shm

Koichi98:
## Description

Follow-up to #1503, which noted both of these as out of scope there because they are changes rather than dead-code removal. Both concern the same function, so they are handled together.

- **Phase 1 takes the read lock.** With the MQ-info buffer gone, `agnocast_ioctl_get_exit_process()` no longer writes anything: it scans `proc_info_htable` and copies `local_pid` / `global_pid` into caller-local storage. On `main` it still had one real mutation — `agnocast_free_exit_subscription_list()` on the no-buffer path — which is why the downgrade only becomes safe on top of #1503. Phase 2 (`agnocast_commit_exit_process`) keeps the write lock.
- **A failed `ret_daemon_should_exit` copy is no longer fatal.** `proc_info` is already freed by then, so `-EFAULT` made `poll_for_unlink()` exit while discarding the `ret_pid` it already held, leaking that process's shm. The flag is advisory and re-derived on the next poll.

## Related links

- Follows #1503, which is where these were split out from

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Notes for reviewers

After #1503, `get_exit_process` was the only `get_*` ioctl in the file still taking the write lock; every remaining `down_write` site mutates.

The copy failure is not covered by KUnit: the cases call `agnocast_ioctl_*` directly, not the `*_cmd` dispatch handlers where `copy_to_user` lives.

## Post-merge checklist

`need-patch-update` — no ABI or interface change; the lock taken internally and an error return.

### DIFF
--- a/agnocast_kmod/agnocast_ioctl.c
+++ b/agnocast_kmod/agnocast_ioctl.c
@@ -2707,8 +2707,9 @@ static long get_exit_process_cmd(struct ioctl_get_exit_process_args __user * arg
   bool daemon_should_exit = false;
   agnocast_commit_exit_process(ipc_ns, global_pid, &daemon_should_exit);
 
-  // Patch ret_daemon_should_exit. Not fatal: proc_info is already freed, so -EFAULT would kill the
-  // daemon while it holds the ret_pid whose shm needs unlinking; the flag is re-derived next poll.
+  // Patch ret_daemon_should_exit. Not fatal: when a pid was returned, its proc_info has already
+  // been committed, so -EFAULT would make the daemon exit while discarding the ret_pid whose shm
+  // needs unlinking; the flag is advisory and re-derived on the next poll.
   if (copy_to_user(&arg->ret_daemon_should_exit, &daemon_should_exit, sizeof(daemon_should_exit))) {
     dev_warn(agnocast_device, "Failed to report the daemon exit flag. (%s)\n", __func__);
   }
--- a/agnocast_kmod/agnocast_ioctl.c
+++ b/agnocast_kmod/agnocast_ioctl.c
@@ -1407,7 +1407,7 @@ pid_t agnocast_ioctl_get_exit_process(
   ioctl_ret->ret_daemon_should_exit = false;
   pid_t global_pid = -1;
 
-  down_write(&global_htables_rwsem);
+  down_read(&global_htables_rwsem);
 
   struct process_info * proc_info;
   int bkt;
@@ -1422,7 +1422,7 @@ pid_t agnocast_ioctl_get_exit_process(
     break;
   }
 
-  up_write(&global_htables_rwsem);
+  up_read(&global_htables_rwsem);
   return global_pid;
 }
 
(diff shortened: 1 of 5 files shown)


agnocast_kmod/agnocast_ioctl.c
```diff
@@ -2707,9 +2707,11 @@ static long get_exit_process_cmd(struct ioctl_get_exit_process_args __user * arg
   bool daemon_should_exit = false;
   agnocast_commit_exit_process(ipc_ns, global_pid, &daemon_should_exit);
 
-  // Patch ret_daemon_should_exit in user-space, after Phase 2 has committed.
-  if (copy_to_user(&arg->ret_daemon_should_exit, &daemon_should_exit, sizeof(daemon_should_exit)))
-    return -EFAULT;
+  // Patch ret_daemon_should_exit. Not fatal: proc_info is already freed, so -EFAULT would kill the
```
atsushi421: The "proc_info is already freed" justification only holds when a pid was actually returned; on an idle poll (`global_pid == -1`) nothing was freed and there is no held `ret_pid`, though returning 0 is still harmless there. Suggest rewording to something like: "when a pid was returned, its proc_info has already been committed, so -EFAULT would make the daemon exit while discarding the ret_pid whose shm needs unlinking; the flag is advisory and re-derived next poll."

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -2707,9 +2707,11 @@ static long get_exit_process_cmd(struct ioctl_get_exit_process_args __user * arg
...
+  // daemon while it holds the ret_pid whose shm needs unlinking; the flag is re-derived next poll.
+  if (copy_to_user(&arg->ret_daemon_should_exit, &daemon_should_exit, sizeof(daemon_should_exit))) {
```
atsushi421, line 2713: Phase 1 already sets `ret_daemon_should_exit = false` and the full-struct `copy_to_user` before the commit delivers it, so this patch copy only changes anything when `daemon_should_exit` is true; on every ordinary poll it rewrites a byte that is already false in user memory. Consider guarding with `daemon_should_exit &&` to skip the second user-space copy in the common case.

atsushi421 (commented):
The recovery story for the non-fatal copy relies on `agnocast_commit_exit_process()` computing `*ret_daemon_should_exit` even when `global_pid == -1`, but no KUnit case asserts that derivation (the `copy_to_user` failure itself is unreachable from KUnit as you note, while this property lives in the ioctl layer KUnit can call). Consider adding a case that calls `agnocast_commit_exit_process(ipc_ns, -1, &flag)` and asserts the flag is true on an empty namespace and false while a process remains, so a future refactor gating the computation behind `global_pid >= 0` cannot silently break the fallback.

---

This fixes the leak only for the patch copy. If the preceding full-struct `copy_to_user` fails, `poll_for_unlink()` still exits, and since `ret_unlink_daemon_exist` counts exited entries, no new process will ever fork a replacement daemon, so every pending exited process's shm leaks until module reload. Probably out of scope here, but worth recording as a follow-up (e.g. retry transient ioctl failures in `poll_for_unlink()`, or exclude exited entries from `ret_unlink_daemon_exist` so a replacement daemon can be forked).

agnocast_kmod/agnocast_ioctl.c
Koichi98: Confirmed — the patch copy only changes user memory when the flag is true. I'd rather keep it unguarded, though: the guard would make Phase 2 depend on Phase 1 initializing the field to `false`, and if that ever changes a stale `true` would silently make the daemon exit early. One byte on a path that sleeps a second per iteration isn't worth the coupling.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed as you proposed in https://github.com/autowarefoundation/agnocast/pull/1520/commits/9a3e31a83ea1ea77041ce8f302b78bf98ae8219d.

Koichi98:
> Consider adding a case that calls `agnocast_commit_exit_process(ipc_ns, -1, &flag)` and asserts the flag is true on an empty namespace and false while a process remains

Added in `9a3e31a8` as `agnocast_kunit_get_exit_process.c`, covering both cases plus a real Phase 1 → Phase 2 commit. The remaining uncovered branches (Phase 1's `break` with several exited entries, Phase 2's `!proc_info` guard) are split into #1527.

> Probably out of scope here, but worth recording as a follow-up

Agreed. Filed as autowarefoundation/agnocast#1526 with both suggested fixes recorded.

atsushi421 (approved):

PR #1522, opened 2026-08-10
feat(agnocastlib): unify Subscription and TakeSubscription

Koichi98:
## Description

`Subscription<MessageT>` can now be constructed without a callback and polled with `take()`, replacing the `Subscription` / `TakeSubscription` split. The mode is fixed at construction and the two are exclusive: `take()` throws `std::runtime_error` on a subscription that has a callback.

That exclusivity is a kernel constraint. `entry_node::referencing_subscribers` is a bitmap with **one bit per (subscriber, entry)**, so the callback path and the take path would each build their own `ipc_shared_ptr` on top of that single bit — whichever is destroyed first clears it while the other still points into the publisher's shared memory. `rclcpp::Subscription::take()` can coexist with a callback only because it copies into caller-provided storage; Agnocast hands out a reference.

`callback_info_id_` becomes `std::optional` and `is_take_sub` is derived from it rather than passed in. `TakeSubscription` becomes a deprecated shim over a callback-less `Subscription`, with its constructor signatures, `::SharedPtr` type, runtime behaviour and tracepoint arguments unchanged. **No kmod change.**

## Motivation

`autoware::agnocast_wrapper::AgnocastSubscription<MessageT>` holds a single `agnocast::Subscription<MessageT>::SharedPtr`. For one wrapper type to serve both callback delivery and polling, that member has to hold either kind — today the callback-less object is an unrelated `TakeSubscription` / `PollingSubscriber`, which would force a variant member or a second implementation class.

Trade-off: the two modes used to be exclusive at compile time; the check is now a runtime `throw`.

## Related links

- #1531 — switches `PollingSubscriber` to `Subscription`, stacked on this branch
- #1532 — removes `TakeSubscription` (needs a major version bump)
- autowarefoundation/autoware_core#1364 — the consumer

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

New integration test `test/integration/test_agnocast_subscription_take.cpp` (10 cases, `requires_kernel_module;requires_heaphook`) passes against a loaded kmod, covering both node overloads, the take paths, the throw on callback subscriptions (typed and type-erased), the callback-registry invariant on destruction, and the `TakeSubscription` / `PollingSubscriber` shims. Disabling the guard makes exactly the three throw cases fail. The e2e runs above are still to be done.

## Notes for reviewers

Spinning a callback and calling `take()` on one subscription now fails loudly instead of corrupting shared memory. No in-tree call site does this.

## Post-merge checklist

### DIFF
--- a/src/agnocastlib/CMakeLists.txt
+++ b/src/agnocastlib/CMakeLists.txt
@@ -278,6 +278,19 @@ if(BUILD_TESTING)
     LABELS "requires_kernel_module;requires_heaphook"
   )
 
+  # Subscription take() integration test (requires kernel module, no mock).
+  # Covers callback-less construction, the take() polling paths and the guard that keeps
+  # callback delivery and take() from sharing one kernel-side reference.
+  ament_add_gmock(test_integration_agnocast_subscription_take_${PROJECT_NAME}
+    test/integration/test_agnocast_subscription_take.cpp)
+  target_link_libraries(test_integration_agnocast_subscription_take_${PROJECT_NAME} agnocast)
+  ament_target_dependencies(test_integration_agnocast_subscription_take_${PROJECT_NAME} std_msgs)
+  set_tests_properties(test_integration_agnocast_subscription_take_${PROJECT_NAME} PROPERTIES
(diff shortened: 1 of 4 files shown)


Koichi98:
Closing: `Subscription` and `TakeSubscription` stay separate.

Merging them only moves the callback/take exclusivity from a compile-time property to a runtime throw. The single handle type is needed at `wrapper::Subscription<M>`, not in agnocast, so autowarefoundation/autoware_core#1363 backs the polling path with `TakeSubscription` instead.

PR #1524, opened 2026-08-13
fix(agnocastlib): honor NodeOptions::automatically_declare_parameters_from_overrides()

Koichi98:
## Description

`agnocast::Node` resolves its parameter overrides but never declares the ones that no explicit `declare_parameter()` call claimed, because the constructor does not pass `NodeOptions::automatically_declare_parameters_from_overrides()` down to `NodeParameters`. Every other flag on that path (`allow_undeclared_parameters`, `use_global_arguments`, `parameter_overrides`) is forwarded — this one was simply missed.

`get_parameters(prefix, map)`, `list_parameters()` and the parameter services all enumerate declared parameters, so an override that was resolved but not declared is invisible to them. Nothing throws and nothing is logged; the caller just gets an empty result.

## Changes

- `node_interfaces::NodeParameters` takes `automatically_declare_parameters_from_overrides` (defaulted to `false` and appended last, so no existing call site changes).
- `agnocast::Node`'s constructor forwards `options.automatically_declare_parameters_from_overrides()`.
- The declaration loop lives in `local_perform_automatically_declare_parameters_from_overrides()`, following the jazzy rclcpp shape. The constructor binds `NodeParameters::` directly, since the object is still under construction there; the new protected `perform_automatically_declare_parameters_from_overrides()` binds through `this`, so a subclass that overrides `declare_parameter()` can have its override respected by calling it after construction.
- `docs/agnocast_node_interface_comparison.md`: note in §3.2 that the option is honored.

## Related links

- The same logic in rclcpp (jazzy, 28.1.18): [`local_perform_automatically_declare_parameters_from_overrides()`](https://github.com/ros2/rclcpp/blob/4321b44ef1a3cc7b46e844c6f2bf148ae5ad2869/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L42-L65), and its two call sites — the [constructor](https://github.com/ros2/rclcpp/blob/4321b44ef1a3cc7b46e844c6f2bf148ae5ad2869/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L126-L144) and [`perform_automatically_declare_parameters_from_overrides()`](https://github.com/ros2/rclcpp/blob/4321b44ef1a3cc7b46e844c6f2bf148ae5ad2869/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L146-L165).

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

Integration tests are added to `test/integration/test_agnocast_node.cpp`, next to the existing per-`NodeOptions`-flag tests:

- with the option on, two overrides under a common prefix become declared parameters, keep their values, and are found by `get_parameters(prefix, map)` under the prefix-stripped keys;
- the declared descriptor has `dynamic_typing` set;
- with the option left at its default, the same override is neither declared nor visible to a prefix lookup;
- the constructor path declares the overrides without dispatching to a subclass override of `declare_parameter()`, while `perform_automatically_declare_parameters_from_overrides()` does.

## Post-merge checklist

### DIFF
--- a/src/agnocastlib/src/node/node_interfaces/node_parameters.cpp
+++ b/src/agnocastlib/src/node/node_interfaces/node_parameters.cpp
@@ -367,7 +367,7 @@ NodeParameters::NodeParameters(
   // but did not get declared explicitly by this point.
   if (automatically_declare_parameters_from_overrides) {
     local_perform_automatically_declare_parameters_from_overrides(
-      this->get_parameter_overrides(),
+      parameter_overrides_,
       [this](const std::string & name) { return NodeParameters::has_parameter(name); },
       [this](
         const std::string & name, const rclcpp::ParameterValue & default_value,
--- a/src/agnocastlib/src/node/node_interfaces/node_parameters.cpp
+++ b/src/agnocastlib/src/node/node_interfaces/node_parameters.cpp
@@ -8,6 +8,7 @@
 #include <algorithm>
 #include <cassert>
 #include <cmath>
+#include <functional>
 #include <limits>
 #include <sstream>
 #include <utility>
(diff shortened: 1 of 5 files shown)


src/agnocastlib/src/node/node_interfaces/node_parameters.cpp
```diff
@@ -344,6 +362,33 @@ NodeParameters::NodeParameters(
   std::string combined_name = node_base_->get_fully_qualified_name();
   parameter_overrides_ =
     resolve_parameter_overrides(combined_name, parameter_overrides, local_args, global_args);
+
+  // If asked, initialize any parameters that ended up in the initial parameter values,
+  // but did not get declared explcitily by this point.
```
kobayu858: typo: explicitly

Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1524/commits/dd905671e17031b3bdb205a877319a049ed824ac.

kobayu858 (dismissed):

kobayu858 (approved):

PR #1528, opened 2026-08-13
feat(agnocast_cie_thread_configurator): manage kernel threads and IRQ affinities

atsushi421:
## Description

The thread configurator exists to centrally manage every scheduling-related element of the application, but until now it only covered announcement-driven application threads (callback groups and non-ROS threads). This PR brings kernel threads and device-backed IRQs under its control as well, using the same prerun-then-configure flow.

### What is added

- **New `system_scan` module** (shared library): enumerates kernel threads from `/proc` (via the `PF_KTHREAD` flag in `/proc/<pid>/stat`, parsing after the last `)` because kthread comms contain spaces and parentheses) and device-backed IRQs from `/sys/kernel/irq` + `/proc/irq`. All observed values are file-based, so both scans are fully unit-tested against fake trees.
- **New YAML sections** emitted by `prerun_node` after `non_ros_threads`:

  ```yaml
  kernel_threads:
    - comm: agnocast_exit_worker
      policy: SCHED_OTHER        # observed value as the initial desired value
      priority: 0                # nice for OTHER/BATCH/IDLE, rt_priority for FIFO/RR
      affinity: [0, 1, 2, 3]
    - comm: ksoftirqd/0
      policy: SCHED_OTHER
      priority: 0
      affinity: UNMANAGEABLE     # PF_NO_SETAFFINITY: the kernel forbids changing it
  irqs:
    - irq: 103
      name: nvidia               # /sys/kernel/irq/103/actions, verified before applying
      affinity: [0, 1, 2, 3]
  ```

- **Apply pass in `thread_configurator_node`** (at startup and on every `~/reapply_config`): matches entries against a fresh `/proc` scan (a comm matches every live instance, e.g. an nfsd pool) and converges the system to the YAML.

### Key design decisions

- **Observed values become the initial desired values with compare-before-set.** The configurator ensures each engaged attribute matches the YAML and issues syscalls/writes only on mismatch. Unedited templates therefore stay no-ops, which also structurally avoids spurious errors from entries the kernel refuses to touch even with identical values (stop-class `migration/N`, kernel-managed MSI-X IRQs that return EIO on every write).
- **`UNMANAGEABLE` sentinel vs YAML null.** Null (or an absent key) keeps its existing meaning of "the user chose not to manage this attribute". The new uppercase sentinel marks attributes whose value cannot be provided or changed by the tool, so prerun output never contains `~` and the two meanings stay distinct in the template.
- **irqbalance validation.** If any IRQ entry is managed while an `irqbalance` process is running, the whole IRQ section is refused (reported as failed) with an instructive error, because applying would only fake success until irqbalance rewrites the masks. Kernel-thread configuration is unaffected.
- **Privileges.** Kernel-thread scheduling reuses the existing `CAP_SYS_NICE` syscall paths (`issue_syscalls` already takes an arbitrary tid). Writing `/proc/irq/<N>/smp_affinity_list` additionally needs `CAP_DAC_OVERRIDE` (unlike `/proc/sys`, plain procfs files honor it), so `setup_thread_configurator.bash` now grants `cap_sys_nice,cap_dac_override=eip`. This was verified empirically: a non-root process with only `CAP_DAC_OVERRIDE` can write the file. The idempotency check was updated so binaries set up by the older script get upgraded to the combined set.
- **kworker/workqueue are deliberately out of scope.** kworker comms mutate at runtime and per-CPU instances reject affinity changes (`PF_NO_SETAFFINITY`); unbound workqueue placement was judged not worth managing. kworker entries are excluded from the scan and rejected by the parser.
- **Breaking srv change**: `ReapplyConfig.srv` gains `applied/failed/skipped_kernel_threads` and `applied/failed_irqs` response arrays (no `skipped_irqs`: an IRQ has no announce-later state). The type hash changes, so out-of-tree service clients must be rebuilt. In-tree consumers are rebuilt together.

Backward compatibility is otherwise preserved: old YAMLs without the new sections parse to empty vectors, old configurators ignore the new sections, the existing `parse_yaml` signature and the announcement-driven accounting (`unapplied_num_` etc.) are untouched.

Verified beyond CI: on a real 16-CPU machine, prerun records 174 kernel-thread comms (including the kmod's `agnocast_exit_worker`) and 80 IRQs; the configurator startup reports all kernel-thread entries in-sync without needing `CAP_SYS_NICE`, and refuses the IRQ section with the instructive error while irqbalance is running.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

Additionally, `test_system_scan` (21 new unit tests) and `test_thread_config` (23 new unit tests) pass, and the thread-configurator launch tests (`reapply` with 4 new cases, `wildcard`, `restart`, `precedence`) all pass locally.

## Post-merge checklist

**Important notes:**

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -13,7 +13,7 @@ All scripts are intended to be invoked from the repository root unless noted oth
 | Script | Purpose |
 |---|---|
 | `dds_config.bash` | Apply CycloneDDS runtime settings (`net.core.rmem_max`, loopback multicast) required for Agnocast over CycloneDDS. Guarded by `/tmp/cycloneDDS_configured` so it runs only once per boot. |
-| `setup_thread_configurator.bash` | Grant `CAP_SYS_NICE` to `thread_configurator_node` and register library paths in `/etc/ld.so.conf.d/agnocast-cie.conf`. Required for Callback Isolated Executor. See the [integration guide](https://autowarefoundation.github.io/agnocast_doc/callback-isolated-executor/integration-guide/#step-2-set-up-the-thread-configurator). |
+| `setup_thread_configurator.bash` | Grant `CAP_SYS_NICE` and `CAP_DAC_OVERRIDE` (for writing `/proc/irq/<N>/smp_affinity_list`) to `thread_configurator_node` and register library paths in `/etc/ld.so.conf.d/agnocast-cie.conf`. Required for Callback Isolated Executor. See the [integration guide](https://autowarefoundation.github.io/agnocast_doc/callback-isolated-executor/integration-guide/#step-2-set-up-the-thread-configurator). |
 | `switch_kmod.bash` | Swap the host's `agnocast-kmod-v<ver>` to another version. For container-based setups where only the host-side kmod needs replacing; the kmod and in-container heaphook must share the same ioctl ABI version. |
 
 ### sample_application/
--- a/scripts/setup_thread_configurator.bash
+++ b/scripts/setup_thread_configurator.bash
@@ -9,7 +9,8 @@
 #   ./scripts/setup_thread_configurator.bash
 #
 # This script automates:
-#   1. Granting CAP_SYS_NICE to the thread_configurator_node binary.
+#   1. Granting CAP_SYS_NICE (scheduling syscalls) and CAP_DAC_OVERRIDE (writes
+#      to root-owned /proc/irq/<N>/smp_affinity_list) to thread_configurator_node.
 #   2. Registering required library paths in /etc/ld.so.conf.d/agnocast-cie.conf.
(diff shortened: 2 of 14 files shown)


atsushi421:
Closing in favor of a 4-PR stack that splits this change into reviewable units:

1. #1543 system_scan module (scan utilities + unit tests)
2. #1544 kernel_threads / irqs YAML parsers (+ unit tests)
3. #1545 prerun template generation
4. #1546 apply logic, reapply integration, srv fields, capability script, E2E test

Merge order: #1543 -> #1544 -> #1545 -> #1546 (each PR's base is the previous branch). The combined tree of the stack is identical to this branch.

PR #1531, opened 2026-08-14
refactor(agnocastlib): make PollingSubscriber wrap Subscription

Koichi98:
## Description

`PollingSubscriber<MessageT>` now holds a callback-less `Subscription<MessageT>` instead of `TakeSubscription<MessageT>`, so the last in-tree user of `TakeSubscription` outside the e2e test is gone. Pure refactor: the constructor arguments reaching the kmod, the `is_take_sub` flag and the emitted tracepoint are unchanged.

Stacked on #1522, which adds the callback-less `Subscription` constructors this depends on. Please review and merge that one first.

## Related links

- #1522

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

Only built so far (`colcon build --packages-select agnocastlib agnocast_e2e_test`). The e2e runs are still to be done.

## Notes for reviewers

`TakeSubscription` itself is untouched and still compiles; its removal is tracked separately and needs a major version bump.

## Post-merge checklist

### DIFF
--- a/src/agnocastlib/include/agnocast/agnocast_subscription.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_subscription.hpp
@@ -460,7 +460,7 @@ public:
 /**
  * @brief Agnocast polling subscriber for a compile-time known message type.
  *
- * Wraps TakeSubscription<MessageT> and exposes a simple take_data() API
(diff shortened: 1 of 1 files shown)


Koichi98:
Closing: depends on the callback-less `Subscription` constructor from #1522, which is not going ahead.

Issue #1489, opened 2026-07-31
Enforce single service provider


bdm-k:
If there are several service servers for the same service name, each of them will respond to service requests, making the system's state inconsistent and unpredictable. ROS 2 does not prevent it because ROS 2 is a distributed system and cannot detect duplicated servers effectively. Agnocast, however, is capable of precisely detecting duplicated servers thanks to its centralized kernel module. We should throw exceptions when constructing the second service server for that service name.

This will be a major change because the contract of service factory APIs will be updated (they never throw now).

PR #1533, opened 2026-08-16
fix(agnocastlib): demand-gate R2A service bridges on external ROS 2 clients

sykwer:
### Problem

  Every `agnocast::Node` auto-creates six ROS 2 parameter services. Each is `ServiceRole::Default`, so each requested an R2A service bridge, which the manager created eagerly. Each R2A
  bridge owns callback groups, which `CallbackIsolatedAgnocastExecutor` backs with a thread and an epoll apiece — O(N_nodes x 6) threads and fds even when nobody touches parameters. 
  At ~175 nodes the manager hit `epoll_create1: EMFILE` and died, taking every bridged topic with it. Pub/sub bridges are already demand-gated; services never got the same gate.

 ### Change

  - `ros2_client_exists()`: the demand signal, via `count_clients()`.
  - `handle_request()` no longer starts R2A eagerly; like A2R it only enters PENDING.
  - `check_and_update_pending()` starts R2A only once a client exists; PENDING itself costs nothing.
  - `check_and_update_r2a()` reaps once the last client leaves.
  - Adds the missing `rclcpp/version.h`, without which every `RCLCPP_VERSION_MAJOR >= 28` guard here silently took its pre-Jazzy branch, making the gate a no-op.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Post-merge checklist

**Important notes:**


bdm-k (dismissed):

src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
```diff
@@ -386,17 +425,20 @@ void ServiceBridgeItem::check_and_update_a2r(const ServiceBridgeDeps & deps)
   shadow_node_ = nullptr;
 }
 
+// Stays in PENDING, or takes arrow (5), (3) or (2), in that order of precedence.
 void ServiceBridgeItem::check_and_update_pending(const ServiceBridgeDeps & deps)
 {
+  // Arrow (5). The two conditions below also exclude (3) and (2), hence the unconditional return.
   if (may_start_r2a_bridge_ && agno_service_exists()) {
-    if (start_r2a_bridge(deps) != 0) {
+    if (ros2_client_exists(deps) && start_r2a_bridge(deps) != 0) {
```
Koichi98, line 453: On Jazzy I think this makes an `agnocast::Node` unreachable from `ros2 param`: the shadow node is the node's only ROS 2 graph presence (`NodeBase::get_rcl_node_handle()` throws, `node_base.cpp:129`) and is created only inside `start_r2a_bridge()` (`:261-266`), while `ros2 param get` returns `Node not found` before creating any client (`ros2param/verb/get.py:51-53`), so the client this gate waits for never appears. The shadow node hosts nothing -- `shadow_node_` is never read and the bridge's ROS 2 service goes on `container_node` (`agnocast_bridge_loader.cpp:343`) -- so would tying its lifetime to the item instead of the bridge work, creating it in `check_and_update_pending()` and releasing it at arrow (2)?

src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
```diff
@@ -330,9 +367,10 @@ void ServiceBridgeItem::update_configuration(const BridgeMsgServicePayload & pay
   }
 }
 
+// Stays in R2A, or takes arrow (6) back to PENDING.
 void ServiceBridgeItem::check_and_update_r2a(const ServiceBridgeDeps & deps)
 {
-  if (agno_service_exists()) {
+  if (agno_service_exists() && ros2_client_exists(deps)) {
```
Koichi98, line 385: Separately, I think this also affects cross-IPC-namespace service bridging: the service side waits here for a ROS 2 client that only the peer namespace's A2R bridge creates, while that A2R waits at `:442` for the ROS 2 service that only this R2A creates. That looks like it needs a forced lease from the discovery agent rather than anything in this PR -- should I pick it up in a separate one?

src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
```diff
@@ -173,8 +174,44 @@ bool ServiceBridgeItem::ros2_service_exists(const ServiceBridgeDeps & deps)
...
+{
+  try {
+#if RCLCPP_VERSION_MAJOR >= 28
+    const bool exists = deps.container_node->count_clients(this->service_name_) > 0;
+    if (!exists) {
+      set_error_string("No external ROS 2 client found");
+    }
+    return exists;
+#else
+    // Pre-Jazzy rclcpp has no service-client count in the graph API, so demand cannot be observed.
```
Koichi98, line 218: Is `so demand cannot be observed` right? There's no count API pre-Jazzy, but [`get_client_names_and_types_by_node`](https://github.com/ros2/rclcpp/blob/16.0.19/rclcpp/include/rclcpp/node_interfaces/node_graph_interface.hpp#L210) looks like it gives the same answer, the way `ros2_service_exists()` already does at `:152-160`.

src/agnocastlib/include/agnocast/bridge/agnocast_service_bridge.hpp
```diff
@@ -22,6 +22,62 @@ struct ServiceBridgeDeps
...
+// is none. The item then waits in Pending, which is where most items spend their whole life: a
+// parameter service nobody calls is never bridged. (5) is taken on the first tick after a client
+// appears, and (6) returns the item here once the last one leaves.
+//
+// That wait is stable rather than transient, because may_start_r2a_bridge_ && agno_service_exists()
+// on their own -- client or no client -- already stop (3) and (2) from being evaluated. This is
+// deliberate. The Agnocast service answers every Agnocast client directly, and an A2R bridge would
+// add a second Agnocast service to the same request topic, so each request would be answered twice.
+// A same-named ROS 2 service is therefore ignored, and the item kept alive, until that Agnocast
+// service goes away.
```
Koichi98, line 79: This doesn't seem to hold once the item is already in A2R: `check_and_update_a2r()` only re-checks `ros2_service_exists()` (`:404`), so if a same-named Agnocast service registers after the A2R bridge is up, the item stays A2R and the request topic ends up with two Agnocast services. It predates this PR so it needn't block here, but I'd like to fix it in a separate PR soon.

src/agnocastlib/src/bridge/agnocast_bridge_manager.cpp
```diff
@@ -245,7 +244,7 @@ void BridgeManager::dispatch_bridge_message(const BridgeMsg & msg)
         active_service_bridges_.erase(it);
       }
 
-      sb_item.handle_request(payload, deps);
+      sb_item.handle_request(payload);
       if (sb_item.state() != ServiceBridgeState::NONE) {
```
Koichi98: `handle_request()` now always leaves the item in PENDING, so this is always true.

src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
bdm-k: Confirmed. `ros2 param` does not work on Jazzy. The proposed solution looks natural and reasonable to me.

src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
bdm-k: `get_client_names_and_types_by_node()` would work, but calling it on every node is very high overhead, especially considering that it is called in every maintenance tick. Unlike `ros2_service_exists()`, `ros2_client_exists()` is allowed to be best-effort, so I think it's fine not using `get_client_names_and_types_by_node()`. But the comment needs updating as you pointed out.

> `ros2_service_exits()` must be precise because a premature decision of creating the A2R service bridge would conflict with the Agnocast service created shortly afterward.

src/agnocastlib/include/agnocast/bridge/agnocast_service_bridge.hpp
bdm-k: My assumption when I designed the service bridge state transition was that it is not allowed for two services (whether ROS 2 or Agnocast) with the same name to exist simultaneously. So, a same-named Agnocast service registering while in the A2R state is already undefined behavior in my opinion. Plus, once we enforce a single service provider (#1489), we don't need to worry about such case.

src/agnocastlib/src/bridge/agnocast_bridge_manager.cpp
bdm-k: agreed.

src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
bdm-k: I would appreciate it if you could fix that in a separate PR.

src/agnocastlib/src/bridge/agnocast_bridge_manager.cpp
Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1533/commits/6dc37a002b0b3b09fa2229185607f5058ca03690.

src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1533/commits/0bc76f20c492bf05a893f6bf67ce7d69b7872b03.

src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
Koichi98: Opened #1552 for this (draft). It extends the existing daemon forced-lease mechanism to services: `BridgeMsgType::DaemonService`, a per-direction lease that both ROS 2 side conditions accept in place of a real DDS endpoint, and service awareness in `bridge_decider.py` derived from the `/AGNOCAST_SRV_REQUEST` topics it already gossips.

Based on this PR's branch, so it should merge after this one.

src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
Koichi98: Ok. I've noted it in https://github.com/autowarefoundation/agnocast/issues/1554.

src/agnocastlib/include/agnocast/bridge/agnocast_service_bridge.hpp
Koichi98: Agreed it's UB and #1489 removes the case structurally. But #1489 is a major change, so I can't tell when it lands, and until then the guard is one-directional -- enter A2R first and every request is silently answered twice.

I was going to send a separate PR, but the alternative here was to qualify the paragraph above to say the guard only holds one way, and re-checking the same condition in `check_and_update_a2r()` turned out shorter than that caveat. `may_start_r2a_bridge_` short-circuits, so a normal A2R item pays only a bool test. Pushed it here.

Koichi98 (approved):

PR #1534, opened 2026-08-16
feat(agnocastlib): add get_actual_qos() to subscriptions and publishers

Koichi98:
## Description

Adds `get_actual_qos()` to `SubscriptionBase`, `PollingSubscriber` and `PublisherBase`, the counterpart of the same accessor on `rclcpp::SubscriptionBase` / `rclcpp::PublisherBase`. Both a subscription and a publisher already resolve their effective QoS at construction — `qos_overriding_options` can replace the QoS passed by the caller with a declared parameter — but that result was discarded after `init_base()`, so callers had no way to read it back.

`init_base()` now stores the resolved QoS in `actual_qos_` and returns `void`; the constructors read the member instead of the former return value. No behavior change beyond the new accessors.

Unlike the rclcpp counterparts, the returned QoS is not RMW-resolved: Agnocast has no DDS entity to query, so `SystemDefault` and the policies Agnocast ignores are reported as requested. This is documented on the accessors.

## Related links

None.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

Nine cases added: `test/unit/test_agnocast_subscription.cpp` covers `Subscription`, `TakeSubscription`, `PollingSubscriber`, the `qos_overriding_options` path, and that the resolved durability still reaches `register_callback()`. `test/unit/test_agnocast_publisher.cpp` covers `Publisher`, `GenericPublisher`, the override path, and that the resolved QoS still reaches `initialize_publisher()`. `ctest -R test_unit_agnocastlib` passes.

## Notes for reviewers

None.

## Post-merge checklist

### DIFF
--- a/src/agnocastlib/include/agnocast/agnocast_subscription.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_subscription.hpp
@@ -128,11 +128,14 @@ public:
   /**
    * @brief Return the QoS passed at construction with any `qos_overriding_options` applied.
    *
-   * Counterpart of `rclcpp::SubscriptionBase::get_actual_qos()`.
+   * Unlike `rclcpp::SubscriptionBase::get_actual_qos()`, the value is not RMW-resolved:
+   * there is no DDS entity to query, so `SystemDefault` and the policies Agnocast ignores are
+   * reported as requested.
+   *
    * @return Effective QoS of this subscription.
    */
   AGNOCAST_PUBLIC
-  const rclcpp::QoS & get_actual_qos() const { return actual_qos_; }
+  rclcpp::QoS get_actual_qos() const { return actual_qos_; }
 
   virtual ~SubscriptionBase()
   {
--- a/src/agnocastlib/include/agnocast/agnocast_subscription.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_subscription.hpp
@@ -183,7 +183,7 @@ class Subscription : public SubscriptionBase
   template <typename NodeT, typename Func>
   void constructor_impl(
     NodeT * node, const std::string & type_name, const rclcpp::QoS & qos, Func && callback,
-    agnocast::SubscriptionOptions options, SubscriptionRole role)
+    const agnocast::SubscriptionOptions & options, SubscriptionRole role)
   {
     rclcpp::CallbackGroup::SharedPtr callback_group = get_valid_callback_group(node, options);
 
@@ -289,7 +289,7 @@ private:
 
   template <typename NodeT>
   void constructor_impl(
-    NodeT * node, const rclcpp::QoS & qos, agnocast::SubscriptionOptions options,
+    NodeT * node, const rclcpp::QoS & qos, const agnocast::SubscriptionOptions & options,
     SubscriptionRole role)
   {
     // Gated to message types — service types pulled in by
--- a/src/agnocastlib/include/agnocast/agnocast_subscription.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_subscription.hpp
@@ -101,12 +101,13 @@ protected:
   topic_local_id_t id_{-1};
   const std::string topic_name_;
   int notify_eventfd_ = -1;  // publish-notification eventfd (-1 for take subscriptions)
+  rclcpp::QoS actual_qos_{1};
   void initialize(
     const rclcpp::QoS & qos, const bool is_take_sub, const bool ignore_local_publications,
     SubscriptionRole role, const std::string & node_name, const std::string & type_name);
 
   template <typename NodeT>
-  rclcpp::QoS init_base(
+  void init_base(
     NodeT * node, const rclcpp::QoS & qos, const std::string & type_name, bool is_take_sub,
     const SubscriptionOptions & options, SubscriptionRole role);
 
@@ -122,6 +123,15 @@ public:
   const char * get_topic_name() const { return topic_name_.c_str(); }
 
   uint32_t get_publisher_count() const { return get_publisher_count_core(topic_name_); }
+
+  /**
+   * @brief Return the QoS passed at construction with any `qos_overriding_options` applied.
+   *
+   * Counterpart of `rclcpp::SubscriptionBase::get_actual_qos()`.
+   * @return Effective QoS of this subscription.
+   */
+  AGNOCAST_PUBLIC
+  const rclcpp::QoS & get_actual_qos() const { return actual_qos_; }
 
   virtual ~SubscriptionBase()
   {
@@ -179,10 +189,10 @@ class Subscription : public SubscriptionBase
     const void * callback_addr = static_cast<const void *>(&callback);
     const char * callback_symbol = tracetools::get_symbol(callback);
 
-    const rclcpp::QoS actual_qos = init_base(node, qos, type_name, false, options, role);
+    init_base(node, qos, type_name, false, options, role);
 
     const bool is_transient_local =
-      actual_qos.durability() == rclcpp::DurabilityPolicy::TransientLocal;
+      actual_qos_.durability() == rclcpp::DurabilityPolicy::TransientLocal;
     callback_info_id_ = agnocast::register_callback<MessageT>(
       std::forward<Func>(callback), topic_name_, id_, is_transient_local, notify_eventfd_,
       callback_group);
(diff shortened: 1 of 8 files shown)


atsushi421 (commented):
`agnocast_publisher.cpp` has the same pattern this PR fixes on the subscription side: the QoS resolved via `qos_overriding_options` is discarded after construction, and rclcpp also offers `PublisherBase::get_actual_qos()`. If the publisher-side counterpart is intentionally out of scope, it would be worth stating that in the PR description (ideally with a tracking issue); otherwise it could be added in the same pass.

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
```diff
@@ -123,6 +124,15 @@ class SubscriptionBase
...
   uint32_t get_publisher_count() const { return get_publisher_count_core(topic_name_); }
 
+  /**
+   * @brief Return the QoS passed at construction with any `qos_overriding_options` applied.
+   *
+   * Counterpart of `rclcpp::SubscriptionBase::get_actual_qos()`.
+   * @return Effective QoS of this subscription.
+   */
+  AGNOCAST_PUBLIC
+  const rclcpp::QoS & get_actual_qos() const { return actual_qos_; }
```
atsushi421: This accessor is not reachable from `PollingSubscriber`, which holds its `TakeSubscription` as a private member and only forwards `takeData()`/`take_data()`, so subscriptions obtained through the polling `create_subscription` overloads cannot read their QoS. Consider adding a forwarding `get_actual_qos()` to `PollingSubscriber` as well.

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
```diff
@@ -123,6 +124,15 @@ class SubscriptionBase
 
   uint32_t get_publisher_count() const { return get_publisher_count_core(topic_name_); }
 
+  /**
+   * @brief Return the QoS passed at construction with any `qos_overriding_options` applied.
+   *
+   * Counterpart of `rclcpp::SubscriptionBase::get_actual_qos()`.
```
atsushi421: The rclcpp counterpart returns `rclcpp::QoS` by value and reflects the RMW-resolved QoS (e.g. SystemDefault policies come back as concrete values), while this method returns a reference that is only valid for the subscription's lifetime and returns the post-override constructor QoS verbatim. Consider noting these differences here so callers migrating from rclcpp do not assume the same contract.

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
```diff
@@ -101,12 +101,13 @@ class SubscriptionBase
   topic_local_id_t id_{-1};
   const std::string topic_name_;
   int notify_eventfd_ = -1;  // publish-notification eventfd (-1 for take subscriptions)
+  rclcpp::QoS actual_qos_{1};
```
atsushi421, line 105: The `{1}` is just a placeholder forced by `rclcpp::QoS` having no default constructor and is always overwritten by `init_base()`, but that is not visible at the declaration site. A short comment like the one on `notify_eventfd_` above would prevent readers from assuming depth 1 is meaningful.

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
```diff
@@ -277,7 +287,7 @@ class TakeSubscription : public SubscriptionBase
   std::mutex last_taken_ptr_mtx_;
 
   template <typename NodeT>
-  rclcpp::QoS constructor_impl(
+  void constructor_impl(
```
atsushi421, line 295: Since this signature is being touched anyway: `options` is taken by value but only forwarded to `init_base()`, which takes a `const SubscriptionOptions &`, so the copy (a `shared_ptr` refcount bump plus a `QosOverridingOptions` copy with its vector and `std::function`) is pure overhead. Consider changing the parameter to `const agnocast::SubscriptionOptions & options`.

src/agnocastlib/test/unit/test_agnocast_subscription.cpp
```diff
@@ -1,7 +1,28 @@
+#include "agnocast/agnocast.hpp"
+#include "agnocast/agnocast_callback_info.hpp"
 #include "agnocast/agnocast_subscription.hpp"
 
+#include "std_msgs/msg/string.hpp"
+
 #include <gtest/gtest.h>
 
+#include <algorithm>
```
atsushi421, line 9: The new code calls `setenv`/`unsetenv` but relies on transitive includes for their declarations. Please add `#include <cstdlib>`.

src/agnocastlib/test/unit/test_agnocast_subscription.cpp
```diff
@@ -47,3 +68,92 @@ TEST_F(GetValidCallbackGroupTest, get_valid_callback_group_nullptr)
...
+    // the variable is unset. It cannot be stubbed here because test_agnocast_utils.cpp tests the
+    // real one in this same binary, so satisfy it instead -- it only inspects the variable.
+    setenv("LD_PRELOAD", "libagnocast_heaphook.so", 1);
+    rclcpp::init(0, nullptr);
+  }
+
+  void TearDown() override
+  {
+    rclcpp::shutdown();
+    unsetenv("LD_PRELOAD");
```
atsushi421: `TearDown()` unconditionally unsets `LD_PRELOAD`, so a value present when the binary was launched (real heaphook, a sanitizer runtime) is lost for the rest of the process, including the threadsafe death tests that re-exec the binary. Consider saving the previous value in `SetUp()` and restoring it here, as test_daemon_bridge_uds.cpp does.

src/agnocastlib/test/unit/test_agnocast_subscription.cpp
Koichi98: Fixed in [1646c165](https://github.com/autowarefoundation/agnocast/commit/1646c16557fc02c6bae38b715d2c0ea5f50da6da).

src/agnocastlib/test/unit/test_agnocast_subscription.cpp
Koichi98: Fixed in [677bff98](https://github.com/autowarefoundation/agnocast/commit/677bff98f2e72346dbfae196d4b04e49bd05e214).

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
Koichi98: Fixed in [8a0cc331](https://github.com/autowarefoundation/agnocast/commit/8a0cc3319024651975fce2e78a9a3c26415dd8f4).

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
Koichi98: Fixed in [db63e736](https://github.com/autowarefoundation/agnocast/commit/db63e73671a8ba6e9d037feeffa84ff395c7a0d6).

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
Koichi98: Fixed in [29131320](https://github.com/autowarefoundation/agnocast/commit/29131320d0fd9d0786971a9142e12e12b74fff9e). Applied to `Subscription::constructor_impl` as well, which had the same signature.

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
Koichi98: Fixed in [a34ba6cf](https://github.com/autowarefoundation/agnocast/commit/a34ba6cf0ab9f07aefed439d7a7c0649e42c9692). The by-value part turned out to be alignable, so instead of only documenting it I changed the return type to `rclcpp::QoS`. The RMW-resolved part is not: there is no DDS entity to query, so that difference is documented.

Koichi98:
Fixed in [cf02a88a](https://github.com/autowarefoundation/agnocast/commit/cf02a88a9a7f57893a5ac97ceb026db47a6f33da). Added `PublisherBase::get_actual_qos()` in the same shape, with unit tests in test_agnocast_publisher.cpp.

atsushi421 (approved):

PR #1535, opened 2026-08-17
feat(agnocastlib): add client GID to service request metadata

bdm-k:
## Description

This PR adds the `client_gid` field to the service request's correlation metadata, allowing the server to uniquely identify the calling client. This is required to implement the service introspection support on the server side; service event message types have the `client_gid` field and the server must populate it with the calling client's GID.

> [!note]
> Actually, the server-side `client_gid` field had not been the GID of the calling client. It had depended on the RMW implementations before the issue ros2/rmw#357 was submitted. According to the issue timeline, efforts have been made to standardize the `client_gid` field across RMW implementations so that it reliably represents the calling client's identity. **Apparently**, the current status is that rmw_connextdds (in fact correct from the beginning) and the latest version of rmw_fastrtps follow the requested semantics, while rmw_cyclonedds still requires a corresponding fix. I think Agnocast should also follow the semantics, so I decided to add the `client_gid` field to the service request's correlation metadata.

### Client GID

The client's GID is derived directly from its underlying distinct publisher. My reasoning here is that, because this internal publisher is not observable to users, it is safe to reuse its GID to represent the client's identity. But maybe that's not the case.

## How was this PR tested?

- [ ] Autoware
- [x] sample application

## Notes for reviewers

Do you think it is safe to use the underlying publisher's GID as-is to present the client's identity?

## Post-merge checklist

### DIFF
--- a/src/agnocastlib/include/agnocast/internal/service_wire_type.hpp
+++ b/src/agnocastlib/include/agnocast/internal/service_wire_type.hpp
@@ -1,16 +1,19 @@
 // On-the-wire layout of Agnocast service request/response messages.
 //
-// ┌───────────┬────────────────────────┐
-// │           │  correlation metadata  │
-// │  payload  ├─────────┬──────────────┤
-// │           │  seqno  │  node_name   │
-// └───────────┼─────────┴──────────────┘
+// ┌───────────┬────────────────────────────────────┐
+// │           │        correlation metadata        │
+// │  payload  ├─────────┬──────────────┬───────────┤
+// │           │  seqno  │  client_gid  │ node_name │
+// └───────────┼─────────┴──────────────┴───────────┘
 //             └ 16-byte aligned
 //
 // payload: request/response payload (ServiceT::Request or ServiceT::Response).
 //
 // seqno: A sequence number to identify each service call (int64_t). To avoid nastiness that may be
 // caused by compiler-inserted padding, this field is conservatively aligned to 16 bytes.
+//
+// client_gid: The client GID (uint8_t array of length RMW_GID_STORAGE_SIZE). This field only exists
+// in the request.
 //
 // node_name: The node name of the caller (std::string). This field only exists in the request.
 
@@ -44,6 +47,7 @@ template <typename ServiceT>
 struct ServiceRequestWrapper : public ServiceT::Request
 {
   alignas(16) int64_t seqno;
+  uint8_t client_gid[RMW_GID_STORAGE_SIZE];
   std::string node_name;
 };
 
@@ -58,6 +62,7 @@ class GenericRequestWrapper
   struct Meta
   {
     alignas(16) int64_t seqno;
+    uint8_t client_gid[RMW_GID_STORAGE_SIZE];
     std::string node_name;
   };
 
@@ -87,6 +92,7 @@ public:
   }
 
   int64_t & seqno() { return meta_ptr_->seqno; }
+  auto client_gid() -> uint8_t (&)[RMW_GID_STORAGE_SIZE] { return meta_ptr_->client_gid; }
   std::string & node_name() { return meta_ptr_->node_name; }
 
   ipc_shared_ptr<void> && take_request() && { return std::move(request_); }
@@ -94,8 +100,8 @@ public:
   /// @brief Allocate a wire-format request buffer in shared memory.
   ///
   /// The payload buffer is initialized via the introspection init function with
-  /// MessageInitialization::SKIP. The seqno number is uninitialized (garbage), and the node_name
-  /// string is default-constructed via placement new.
+  /// MessageInitialization::SKIP. The seqno number and client_gid are uninitialized (garbage),
+  /// and the node_name string is default-constructed via placement new.
   ///
   /// @param request_members The introspection message members for the request type.
   /// @param borrow_loaned_message A callable that allocates a buffer of the given size in shared
--- a/src/agnocastlib/include/agnocast/internal/service_wire_type.hpp
+++ b/src/agnocastlib/include/agnocast/internal/service_wire_type.hpp
@@ -60,51 +60,52 @@ struct DummyService
 namespace agnocast
 {
 
-template <typename ServiceT>
-struct ServiceRequestWrapper : public ServiceT::Request
+struct RequestMeta
 {
   alignas(16) int64_t seqno;
   uint8_t client_gid[RMW_GID_STORAGE_SIZE];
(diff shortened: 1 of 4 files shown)


Koichi98:
Unrelated to this PR: are services assumed to be 1 client per node? The response topic is keyed on the node name (`agnocast_utils.cpp:116-120`) and the client matches by `seqno` alone (`agnocast_client.hpp:191`), which restarts at 0 per client — so two clients for the same service on one node would collide. Am I right?

src/agnocastlib/include/agnocast/internal/service_wire_type.hpp
```diff
@@ -44,6 +47,7 @@ template <typename ServiceT>
 struct ServiceRequestWrapper : public ServiceT::Request
 {
   alignas(16) int64_t seqno;
+  uint8_t client_gid[RMW_GID_STORAGE_SIZE];
```
Koichi98, line 66: This file now uses `RMW_GID_STORAGE_SIZE`, but only gets it transitively via `agnocast_smart_pointer.hpp` → `agnocast_utils.hpp` → `rclcpp/rclcpp.hpp`. Could you add `#include <rmw/types.h>`?

src/agnocastlib/include/agnocast/internal/service_wire_type.hpp
```diff
@@ -58,6 +62,7 @@ class GenericRequestWrapper
   struct Meta
   {
     alignas(16) int64_t seqno;
+    uint8_t client_gid[RMW_GID_STORAGE_SIZE];
```
Koichi98: `ServiceRequestWrapper` and `GenericRequestWrapper::Meta` have to stay byte-identical, but the new field is duplicated by hand. Could we add a `static_assert` on a sample srv type so a one-sided edit fails at compile time?

bdm-k:
> are services assumed to be 1 client per node?

Yes, we are making that assumption. The reasoning is that if a client for a particular service already exists within the node, we can simply use it instead of creating a second one.

Koichi98:
> The reasoning is that if a client for a particular service already exists within the node, we can simply use it instead of creating a second one.

Okay. But is the constraint actually needed? It seems like rclcpp allows N clients per node, and aligning with it would mostly mean keying the response topic on the client instead of the node. If we do keep it, could we at least make it fail loudly rather than silently resolving the wrong future?

src/agnocastlib/include/agnocast/internal/service_wire_type.hpp
bdm-k: fixed

src/agnocastlib/include/agnocast/internal/service_wire_type.hpp
bdm-k: fixed

bdm-k:
I think you're absolutely right. It looks like we can use the `pub_id` of the underlying publisher as the key instead of the node name. I'll note this enhancement as an issue and will give it a try when I'm available.

bdm-k:
Or should I work on it now so that we don't need to have `client_gid` in the service request?

Koichi98:
> Or should I work on it now so that we don't need to have client_gid in the service request?

I don't think it would remove the need for it — the server still has to read the client identity out of the request, whether the key is the GID itself or the `pub_id` embedded in it at `[12..15]`. So I'd merge this one first.

src/agnocastlib/include/agnocast/internal/service_wire_type.hpp
Koichi98: Thanks. One follow-up: the assert only compares sizes, so a one-sided reordering still passes — e.g. swapping `client_gid` and `node_name` keeps `sizeof` at 64 while moving `client_gid` from offset 8 to 40. Could we define the metadata once and inherit it from both wrappers instead?

bdm-k: defined `RequestMeta` and `ResponseMeta` at the top level and made the template wrappers inherit them as well as `ServiceT::Request`/`ServiceT::Response`.

[adb9ad1](https://github.com/autowarefoundation/agnocast/pull/1535/commits/adb9ad163e538418d7889ff5ed2542ac482c5783)

Koichi98: Thanks. `GenericResponseWrapper` still refers to `RequestMeta` rather than `ResponseMeta` — is that intended?

src/agnocastlib/include/agnocast/internal/service_wire_type.hpp
```diff
@@ -35,48 +40,72 @@ constexpr size_t offsetof_meta(size_t base_size)
...
 
-template <typename ServiceT>
-struct ServiceResponseWrapper : public ServiceT::Response
+struct ResponseMeta
 {
   alignas(16) int64_t seqno;
 };
 
+template <typename ServiceT>
+struct ServiceRequestWrapper : public ServiceT::Request, public RequestMeta
```
Koichi98, line 79: Sorry, the inheritance I suggested has a side effect: a payload field with the same name as a metadata field is now ambiguous instead of hidden. `composition_interfaces/srv/LoadNode` has `node_name` in its Request and no longer compiles. `seqno`, `node_name`, `client_gid` are all exposed. Qualifying (`request->RequestMeta::node_name`) may fix it.

bdm-k: Ah, yes. I didn't notice.

bdm-k: fixed

Koichi98: Thanks. `agnocast_service.hpp` looks unqualified still.

bdm-k: sorry, I forgot to include the file for the commit. It should be fixed now.

src/agnocastlib/include/agnocast/internal/service_wire_type.hpp
bdm-k: fixed in [5b01ba5](https://github.com/autowarefoundation/agnocast/pull/1535/commits/5b01ba5bb6d5f0b8821ace6561ad1c688261ca5c)

Koichi98 (dismissed):

Koichi98 (approved):

PR #1537, opened 2026-08-17
fix(agnocastlib): support building with TRACETOOLS_DISABLED

kobayu858:
## Description
Fix `agnocastlib` so that it builds with `TRACETOOLS_DISABLED`. `tracetools::get_symbol` is only declared when tracing is compiled in ([humble](https://github.com/ros2/ros2_tracing/blob/humble/tracetools/include/tracetools/utils.hpp#L26), [jazzy](https://github.com/ros2/ros2_tracing/blob/jazzy/tracetools/include/tracetools/utils.hpp#L23)), so calling it without a guard broke the build on both distributions:

```
error: 'tracetools' has not been declared
```

The four `tracetools::get_symbol` call sites therefore now go through a new `agnocast::get_callback_symbol` wrapper. Once tracing is compiled out, variables that exist only to be passed to tracepoints become unused, so they are marked `[[maybe_unused]]`.

While doing this, an ownership difference between the distributions came to light:

| | return type | ownership |
|---|---|---|
| Humble | [`const char *`](https://github.com/ros2/ros2_tracing/blob/humble/tracetools/include/tracetools/utils.hpp#L57) | must not be freed |
| Jazzy | [`char *`](https://github.com/ros2/ros2_tracing/blob/jazzy/tracetools/include/tracetools/utils.hpp#L56) | caller must `free()` it, and it may be nullptr |

The wrapper handles this as well. Agnocast never freed the string, so it leaked on every subscription and timer creation on Jazzy (rclcpp [frees it explicitly](https://github.com/ros2/rclcpp/blob/jazzy/rclcpp/include/rclcpp/timer.hpp#L254-L259)). Returning `std::string` fixes that and guarantees a non-null argument for `ctf_string()`.

## Related links

- #1511 — enable tracepoints by default

## How was this PR tested?

- [ ] Autoware
- [x] sample application

Built on Humble with and without `TRACETOOLS_DISABLED`, both with 0 errors and 0 warnings. It did not compile at all before this change:

| build | LTTng linked | `ros_trace_agnocast_*` | probes |
|---|---|---|---|
| default | yes | 13 | 13 |
| `TRACETOOLS_DISABLED` | no | 0 | 0 |

## Notes for reviewers

The distribution is detected from the return type of `get_symbol` rather than a distro macro, following the existing `#ifndef DECLARE_TRACEPOINT` compatibility shim in `agnocast_tracepoint_wrapper.h`.

`get_callback_symbol` is placed in `agnocast_tracepoint_wrapper.h` because it exists solely to produce a tracepoint argument and that header already owns the tracetools dependency. It needs a `#ifdef __cplusplus` block since the header is also included from C. Happy to move it to a dedicated C++-only header if you prefer keeping that header free of C++.

## Post-merge checklist

**Important notes:**

### DIFF
--- a/src/agnocastlib/include/agnocast/agnocast.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast.hpp
@@ -200,6 +200,10 @@ typename Subscription<MessageT>::SharedPtr create_subscription(
 /// @return Shared pointer to the created polling subscription.
 AGNOCAST_PUBLIC
 template <typename MessageT, typename NodeT>
+[[deprecated(
+  "agnocast::PollingSubscriber is planned to move to autoware_agnocast_wrapper and be removed from "
+  "agnocast. Obtain a polling subscriber from the wrapper, or use agnocast::TakeSubscription "
+  "directly.")]]
 typename PollingSubscriber<MessageT>::SharedPtr create_subscription(
   NodeT * node, const std::string & topic_name, const size_t qos_history_depth)
 {
@@ -219,6 +223,10 @@ typename PollingSubscriber<MessageT>::SharedPtr create_subscription(
 /// @return Shared pointer to the created polling subscription.
 AGNOCAST_PUBLIC
 template <typename MessageT, typename NodeT>
+[[deprecated(
+  "agnocast::PollingSubscriber is planned to move to autoware_agnocast_wrapper and be removed from "
+  "agnocast. Obtain a polling subscriber from the wrapper, or use agnocast::TakeSubscription "
+  "directly.")]]
 typename PollingSubscriber<MessageT>::SharedPtr create_subscription(
   NodeT * node, const std::string & topic_name, const rclcpp::QoS & qos)
 {
--- a/src/agnocastlib/include/agnocast/agnocast.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast.hpp
@@ -191,6 +191,44 @@ typename Subscription<MessageT>::SharedPtr create_subscription(
     std::forward<Func>(callback), options);
 }
 
+/// @brief Create an Agnocast take-subscription (Stage 1 free function, QoS overload).
+/// @tparam MessageT ROS message type.
+/// @tparam NodeT Node type (rclcpp::Node or agnocast::Node).
+/// @param node Pointer to the node.
+/// @param topic_name Topic name.
+/// @param qos Quality of service profile.
+/// @param options Subscription options.
+/// @return Shared pointer to the created take-subscription.
+AGNOCAST_PUBLIC
+template <typename MessageT, typename NodeT>
+typename TakeSubscription<MessageT>::SharedPtr create_take_subscription(
+  NodeT * node, const std::string & topic_name, const rclcpp::QoS & qos,
+  agnocast::SubscriptionOptions options = agnocast::SubscriptionOptions{})
+{
+  static_assert(
+    std::is_base_of_v<rclcpp::Node, NodeT> || std::is_base_of_v<agnocast::Node, NodeT>,
+    "NodeT must be rclcpp::Node or agnocast::Node (or derived from them)");
+  return std::make_shared<TakeSubscription<MessageT>>(node, topic_name, qos, std::move(options));
+}
+
+/// @brief Create an Agnocast take-subscription (Stage 1 free function, history-depth overload).
+/// @tparam MessageT ROS message type.
+/// @tparam NodeT Node type (rclcpp::Node or agnocast::Node).
+/// @param node Pointer to the node.
+/// @param topic_name Topic name.
+/// @param qos_history_depth History depth for the QoS profile.
+/// @param options Subscription options.
+/// @return Shared pointer to the created take-subscription.
+AGNOCAST_PUBLIC
+template <typename MessageT, typename NodeT>
+typename TakeSubscription<MessageT>::SharedPtr create_take_subscription(
+  NodeT * node, const std::string & topic_name, const size_t qos_history_depth,
+  agnocast::SubscriptionOptions options = agnocast::SubscriptionOptions{})
+{
+  return create_take_subscription<MessageT>(
+    node, topic_name, rclcpp::QoS(rclcpp::KeepLast(qos_history_depth)), std::move(options));
+}
+
 /// @brief Create an Agnocast polling subscription (Stage 1 free function, history-depth overload).
 /// @tparam MessageT ROS message type.
 /// @tparam NodeT Node type (rclcpp::Node or agnocast::Node).
@@ -202,8 +240,8 @@ AGNOCAST_PUBLIC
 template <typename MessageT, typename NodeT>
 [[deprecated(
   "agnocast::PollingSubscriber is planned to move to autoware_agnocast_wrapper and be removed from "
-  "agnocast. Obtain a polling subscriber from the wrapper, or use agnocast::TakeSubscription "
-  "directly.")]]
+  "agnocast. Obtain a polling subscriber from the wrapper, or use "
+  "agnocast::create_take_subscription().")]]
 typename PollingSubscriber<MessageT>::SharedPtr create_subscription(
   NodeT * node, const std::string & topic_name, const size_t qos_history_depth)
 {
@@ -225,8 +263,8 @@ AGNOCAST_PUBLIC
 template <typename MessageT, typename NodeT>
 [[deprecated(
   "agnocast::PollingSubscriber is planned to move to autoware_agnocast_wrapper and be removed from "
-  "agnocast. Obtain a polling subscriber from the wrapper, or use agnocast::TakeSubscription "
-  "directly.")]]
+  "agnocast. Obtain a polling subscriber from the wrapper, or use "
+  "agnocast::create_take_subscription().")]]
 typename PollingSubscriber<MessageT>::SharedPtr create_subscription(
   NodeT * node, const std::string & topic_name, const rclcpp::QoS & qos)
 {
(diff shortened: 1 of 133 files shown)


src/agnocastlib/include/agnocast/agnocast.hpp
```diff
@@ -384,8 +384,8 @@ TimerBase::SharedPtr create_timer(
   const uint32_t timer_id = allocate_timer_id();
   const auto period_ns = period.to_chrono<std::chrono::nanoseconds>();
 
-  const void * callback_addr = static_cast<const void *>(&callback);
-  const char * callback_symbol = tracetools::get_symbol(callback);
+  [[maybe_unused]] const void * callback_addr = static_cast<const void *>(&callback);
```
atsushi421, line 425: `node_handle` a few lines below is assigned in both `if constexpr` branches but consumed only by the `TRACEPOINT`, so under `TRACETOOLS_DISABLED` Clang emits `-Wunused-but-set-variable` for it (GCC suppresses this warning inside templates, which is why the Humble/GCC build showed no warnings). Consider marking it `[[maybe_unused]]` like `callback_addr`.

src/agnocastlib/include/agnocast/agnocast_tracepoint_wrapper.h
```diff
@@ -96,3 +96,52 @@ DECLARE_TRACEPOINT(
...
+ */
+template <typename CallbackT>
+inline std::string get_callback_symbol([[maybe_unused]] const CallbackT & callback)
+{
+#ifndef TRACETOOLS_DISABLED
+  auto * symbol = tracetools::get_symbol(callback);
+  if (symbol == nullptr) {
+    return {};
+  }
+  std::string result{symbol};
```
atsushi421: On Jazzy, if this `std::string` construction throws `std::bad_alloc`, the function exits before `std::free(symbol)` and the demangled buffer leaks. Consider holding the owned pointer in a scope guard (e.g. `std::unique_ptr<char, decltype(&std::free)>`) right after the null check so the free runs on all exit paths.

src/agnocastlib/include/agnocast/agnocast_tracepoint_wrapper.h
```diff
@@ -96,3 +96,52 @@ DECLARE_TRACEPOINT(
...
+ * \param[in] callback the callback to resolve
+ * \return the symbol, or an empty string if it is unavailable
+ */
+template <typename CallbackT>
+inline std::string get_callback_symbol([[maybe_unused]] const CallbackT & callback)
+{
+#ifndef TRACETOOLS_DISABLED
+  auto * symbol = tracetools::get_symbol(callback);
+  if (symbol == nullptr) {
+    return {};
```
atsushi421: On Humble, a failed resolution yields a string starting with `TRACETOOLS_SYMBOL_UNKNOWN` ("UNKNOWN"), while this maps Jazzy's nullptr to an empty string, so trace consumers see inconsistent `function_symbol` values across distros. Consider returning `TRACETOOLS_SYMBOL_UNKNOWN` here and updating the `\return` doc accordingly.

src/agnocastlib/include/agnocast/agnocast_tracepoint_wrapper.h
```diff
@@ -96,3 +96,52 @@ DECLARE_TRACEPOINT(
...
+  }
+  std::string result{symbol};
+  if constexpr (!std::is_const_v<std::remove_pointer_t<decltype(symbol)>>) {
```
atsushi421, line 144: The constness check encodes "Jazzy returns `char *`" indirectly, leaving the reader to reconstruct why constness implies ownership. `if constexpr (std::is_same_v<decltype(symbol), char *>)` would state the discriminator directly; alternatively, keep the check and extend the comment to say the non-const return type is what marks the caller-owned contract.

src/agnocastlib/include/agnocast/node/agnocast_node.hpp
```diff
@@ -565,8 +565,8 @@ class Node
     const uint32_t timer_id = allocate_timer_id();
     const auto period_ns = std::chrono::duration_cast<std::chrono::nanoseconds>(period);
 
-    const void * callback_addr = static_cast<const void *>(&callback);
-    const char * callback_symbol = tracetools::get_symbol(callback);
+    [[maybe_unused]] const void * callback_addr = static_cast<const void *>(&callback);
+    const std::string callback_symbol = agnocast::get_callback_symbol(callback);
```
atsushi421, line 610: This header does not include `agnocast/agnocast_tracepoint_wrapper.h` directly, so `agnocast::get_callback_symbol` resolves only through the transitive include from `agnocast_subscription.hpp`. Consider adding the direct include so a future include cleanup there does not silently break this header.

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
```diff
@@ -319,7 +320,7 @@ class TakeSubscription : public SubscriptionBase
 
     {
       auto default_cbg = node->get_node_base_interface()->get_default_callback_group();
-      auto dummy_cb = []() {};
+      [[maybe_unused]] auto dummy_cb = []() {};
```
atsushi421: An alternative to the `[[maybe_unused]]` annotations: wrapping the tracepoint-only blocks in `#ifndef TRACETOOLS_DISABLED` (the pattern rclcpp uses) would compile out the remaining runtime work in disabled builds, such as the `get_default_callback_group` lookups, the `"dummy_take" + topic_name_` concatenation, and the `get_callback_symbol` calls, and would make most of the annotations unnecessary.

atsushi421 (commented):
`src/agnocastlib/src/agnocast_tracepoint_wrapper.c` still says "Building with TRACETOOLS_DISABLED will be supported later." Since this PR is that support, please update or drop that sentence here as well.

---

In `src/agnocastlib/src/agnocast_timer_info.cpp` (around lines 452 and 546), lambdas capture `callable_ptr` only to feed tracepoints, so under `TRACETOOLS_DISABLED` Clang's `-Wunused-lambda-capture` fires at sites this PR does not touch (GCC has no equivalent warning, which is why the GCC build was clean). If a warning-clean Clang build is a goal, those captures need the same treatment.

src/agnocastlib/include/agnocast/agnocast.hpp
kobayu858: Confirmed with Clang and fixed, though by wrapping the block in `#ifndef TRACETOOLS_DISABLED` rather than annotating it. https://github.com/autowarefoundation/agnocast/pull/1537/commits/fcfb515821e4091b028ae88789b5c78e0441ccbb

src/agnocastlib/include/agnocast/agnocast_tracepoint_wrapper.h
kobayu858: Good catch — fixed with the scope guard you suggested; the free now happens on the throwing path too. https://github.com/autowarefoundation/agnocast/pull/1537/commits/d6622680264b76691916d826cba9f493e851eff9

src/agnocastlib/include/agnocast/agnocast_tracepoint_wrapper.h
kobayu858: Fixed by returning `TRACETOOLS_SYMBOL_UNKNOWN` in both cases, with a fallback definition because Jazzy dropped the macro. https://github.com/autowarefoundation/agnocast/pull/1537/commits/d6622680264b76691916d826cba9f493e851eff9

src/agnocastlib/include/agnocast/agnocast_tracepoint_wrapper.h
kobayu858: Kept the constness check, since it survives a change of pointee type, and extended the comment to say the non-const return type is what marks the buffer as caller-owned. https://github.com/autowarefoundation/agnocast/pull/1537/commits/fcfb515821e4091b028ae88789b5c78e0441ccbb

src/agnocastlib/include/agnocast/node/agnocast_node.hpp
kobayu858: Added the direct include. https://github.com/autowarefoundation/agnocast/pull/1537/commits/c519886d0b31664eae2c9dfe056e9979219fb70b

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
kobayu858: Adopted it where the tracepoint block is self-contained, which removed most annotations and the runtime work you listed; kept `[[maybe_unused]]` only where the variable is also read outside the block.
https://github.com/autowarefoundation/agnocast/pull/1537/commits/fcfb515821e4091b028ae88789b5c78e0441ccbb

kobayu858:
> src/agnocastlib/src/agnocast_tracepoint_wrapper.c still says "Building with TRACETOOLS_DISABLED will be supported later." Since this PR is that support, please update or drop that sentence here as well.

Good catch — that sentence was written when the support was still pending, and this PR is that support, so I dropped it. https://github.com/autowarefoundation/agnocast/pull/1537/commits/f7fc3069726a137297c8da729781e37c37ed985c

> In src/agnocastlib/src/agnocast_timer_info.cpp (around lines 452 and 546), lambdas capture callable_ptr only to feed tracepoints, so under TRACETOOLS_DISABLED Clang's -Wunused-lambda-capture fires at sites this PR does not touch (GCC has no equivalent warning, which is why the GCC build was clean). If a warning-clean Clang build is a goal, those captures need the same treatment.

Confirmed: `-Wunused-lambda-capture` fires on Clang with -Wall -Wextra and GCC has no equivalent, which is why my GCC-only check looked clean. `[[maybe_unused]]` cannot be applied to a lambda capture (rejected in C++17 and C++20 alike), so I marked the capture as used with `(void)callable_ptr;` in both lambdas instead. https://github.com/autowarefoundation/agnocast/pull/1537/commits/e74981fe8cf5654400eebf34d5055d122fb0252e

atsushi421 (approved):

Koichi98 (approved):

PR #1539, opened 2026-08-18
refactor(agnocast_kmod): split the domain bridge rule add path into helpers

Koichi98:
## Description

`agnocast_ioctl_add_domain_bridge()` did three things in one body: validate and canonicalize the
request, decide whether the pair may be registered, and allocate the rule. This splits the last two
into `add_domain_rule()` and `insert_domain_rule()`, leaving the ioctl entry point with only the
argument handling and the lock. Canonicalization now happens in `insert_domain_rule()`, where the
storage layout is what demands it.

Delivery, return codes and lock scope are unchanged. The one deliberate behavior change is that the
two rejection paths now log which invariant was violated, which they previously did not.

## Related links

None.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application
- [x] `bash scripts/test/e2e_test_domain_bridge.bash`

## Notes for reviewers

None.

## Post-merge checklist

`need-patch-update`

### DIFF
--- a/agnocast_kmod/agnocast_ioctl.c
+++ b/agnocast_kmod/agnocast_ioctl.c
@@ -2354,10 +2354,10 @@ static int add_domain_rule(
       r_from->b_to_a = true;
     }
 
-    // Unlike the new-pair path (the fallthrough, which delegates to insert_domain_rule), this
-    // one runs after endpoints may have registered, and the direction just enabled was denied
-    // when their notify lists were built. Both cells share one topic_struct once grouped, so
-    // either wrapper reaches every publisher.
+    // Unlike the new-pair path that insert_domain_rule handles, this one runs after endpoints
+    // may have registered, and the direction just enabled was denied when their notify lists
+    // were built. Both cells share one topic_struct once grouped, so either wrapper reaches
+    // every publisher.
     struct topic_wrapper * wrapper = find_topic(topic_name_from, ipc_ns, from_domain);
     if (!wrapper) wrapper = find_topic(topic_name_to, ipc_ns, to_domain);
     if (wrapper) rebuild_all_notify_lists(wrapper);
--- a/agnocast_kmod/agnocast_ioctl.c
+++ b/agnocast_kmod/agnocast_ioctl.c
@@ -2337,9 +2337,10 @@ static int add_domain_rule(
     r_a->a_to_b |= from_is_a;
     r_a->b_to_a |= !from_is_a;
 
-    // Unlike the paths below, this one runs after endpoints may have registered, and the
-    // direction just enabled was denied when their notify lists were built. Both cells share one
-    // topic_struct once grouped, so either wrapper reaches every publisher.
+    // Unlike the new-pair path (the fallthrough, which delegates to insert_domain_rule), this
+    // one runs after endpoints may have registered, and the direction just enabled was denied
+    // when their notify lists were built. Both cells share one topic_struct once grouped, so
+    // either wrapper reaches every publisher.
     struct topic_wrapper * wrapper = find_topic(name_a, ipc_ns, domain_a);
     if (!wrapper) wrapper = find_topic(name_b, ipc_ns, domain_b);
     if (wrapper) rebuild_all_notify_lists(wrapper);
--- a/agnocast_kmod/agnocast_ioctl.c
+++ b/agnocast_kmod/agnocast_ioctl.c
@@ -2285,8 +2285,8 @@ static struct domain_bridge_rule * find_domain_rule(
   return NULL;
 }
 
-// The caller holds global_htables_rwsem for write and has already established that neither cell
-// is taken.
+// The caller holds global_htables_rwsem for write and has verified, for both cells, that no
+// domain_bridge_rule covers it (find_domain_rule) and that no endpoint has joined it (find_topic).
 static int insert_domain_rule(
   const char * name_a, const char * name_b, const struct ipc_namespace * ipc_ns,
   const uint32_t domain_a, const uint32_t domain_b, const bool from_is_a)
--- a/agnocast_kmod/agnocast_ioctl.c
+++ b/agnocast_kmod/agnocast_ioctl.c
@@ -2288,9 +2288,17 @@ static struct domain_bridge_rule * find_domain_rule(
 // The caller holds global_htables_rwsem for write and has verified, for both cells, that no
 // domain_bridge_rule covers it (find_domain_rule) and that no endpoint has joined it (find_topic).
 static int insert_domain_rule(
-  const char * name_a, const char * name_b, const struct ipc_namespace * ipc_ns,
-  const uint32_t domain_a, const uint32_t domain_b, const bool from_is_a)
+  const char * topic_name_from, const char * topic_name_to, const uint32_t from_domain,
+  const uint32_t to_domain, const struct ipc_namespace * ipc_ns)
 {
+  // Store the pair canonically (domain_a < domain_b), each domain keeping its own name
+  // (they differ on rename). Direction lives only in the a_to_b / b_to_a flags; grouping
+  // (find_grouped_topic_struct) pairs the two cells and delivery direction is enforced by
+  // domain_delivery_allowed.
+  const bool from_is_a = from_domain < to_domain;
+  const char * name_a = from_is_a ? topic_name_from : topic_name_to;
+  const char * name_b = from_is_a ? topic_name_to : topic_name_from;
+
   struct domain_bridge_rule * rule = kmalloc(sizeof(*rule), GFP_KERNEL);
   if (!rule) return -ENOMEM;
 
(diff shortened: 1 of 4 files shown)


agnocast_kmod/agnocast_ioctl.c
```diff
@@ -2285,71 +2285,22 @@ static struct domain_bridge_rule * find_domain_rule(
   return NULL;
 }
 
-int agnocast_ioctl_add_domain_bridge(
-  const char * topic_name_from, const char * topic_name_to, const uint32_t from_domain,
-  const uint32_t to_domain, const struct ipc_namespace * ipc_ns)
+// The caller holds global_htables_rwsem for write and has already established that neither cell
+// is taken.
+static int insert_domain_rule(
```
atsushi421, line 2290: All 19 kunit cases in agnocast_kunit_domain_bridge.c make their first add_domain_bridge call with from_domain < to_domain (the `(2, 1)` calls are all re-declarations that stop at the r_a == r_b path), so insert_domain_rule never runs with from_is_a == false. That leaves the `a_to_b = false` / `b_to_a = true` initialization and the new dev_info order reconstruction untested, and the reconstruction is the only non-mechanical transformation in this PR: a transposed ternary would silently log the reversed direction, and the flag half would deliver in the wrong direction.

Consider adding a kunit case that declares `(TOPIC_NAME, TOPIC_NAME, 2, 1)` as the first rule and asserts via agnocast_get_domain_rule that domain_a == 1, domain_b == 2, a_to_b == false, b_to_a == true.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -2285,71 +2285,22 @@ static struct domain_bridge_rule * find_domain_rule(
   return NULL;
 }
 
-int agnocast_ioctl_add_domain_bridge(
-  const char * topic_name_from, const char * topic_name_to, const uint32_t from_domain,
-  const uint32_t to_domain, const struct ipc_namespace * ipc_ns)
+// The caller holds global_htables_rwsem for write and has already established that neither cell
```
atsushi421: The caller establishes two distinct preconditions here: no domain_bridge_rule covers either cell (find_domain_rule), and no endpoint has joined either cell (find_topic). "Neither cell is taken" reads as either one on its own, and a future direct caller satisfying only the first would hash_add a second rule for an already-paired cell, making find_domain_rule's full-bucket scan return whichever duplicate it hits first.

Suggest naming both checks: "has verified, for both cells, that no domain_bridge_rule covers it (find_domain_rule) and that no endpoint has joined it (find_topic)".

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -2360,11 +2311,66 @@ int agnocast_ioctl_add_domain_bridge(
   // find_domain_rule scans every bucket, so the hash key is only for even distribution.
   hash_add(domain_rule_htable, &rule->node, full_name_hash(NULL, name_a, strlen(name_a)));
 
+  // Report it the way it was declared, not in the canonical order it is stored in.
   dev_info(
-    agnocast_device, "Domain bridge rule added (%s@%u -> %s@%u).\n", topic_name_from, from_domain,
-    topic_name_to, to_domain);
+    agnocast_device, "Domain bridge rule added (%s@%u -> %s@%u).\n", from_is_a ? name_a : name_b,
```
atsushi421: The entry point encodes the declared direction into canonical values plus a from_is_a bit, and this dev_info decodes it back with four ternaries. Nothing below the entry point actually needs pre-canonicalized values: find_domain_rule and find_topic are direction-symmetric, and from_is_a is derivable from `from_domain < to_domain`.

Moving the canonicalization into insert_domain_rule and passing the original five arguments through would let this log print its parameters verbatim, drop both helpers from six parameters to five, and align the argument order with the entry point. The current order also separates each name from its domain, unlike every other per-cell function in this file (find_topic, find_domain_rule, find_grouped_topic_struct all use `name, ipc_ns, domain`), so two same-typed arguments can be swapped at a call site without a compiler error.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -2360,11 +2311,66 @@ int agnocast_ioctl_add_domain_bridge(
...
+  // TODO: support >2 domains per topic by storing a domain group instead of a fixed pair.
+  struct domain_bridge_rule * r_a = find_domain_rule(name_a, ipc_ns, domain_a);
+  struct domain_bridge_rule * r_b = find_domain_rule(name_b, ipc_ns, domain_b);
+  if (r_a || r_b) {
+    if (r_a != r_b) return -EBUSY;
+
+    r_a->a_to_b |= from_is_a;
+    r_a->b_to_a |= !from_is_a;
+
+    // Unlike the paths below, this one runs after endpoints may have registered, and the
```
atsushi421: The allocation path this contrasts against now lives in insert_domain_rule, which is defined above this comment, so a reader following "below" top-to-bottom finds only a call, not the kmalloc/hash_add/dev_info logic being contrasted. Consider wording that does not depend on physical position, e.g. "Unlike the new-pair path (the fallthrough, which delegates to insert_domain_rule), this one runs after endpoints may have registered".

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -2360,11 +2311,66 @@ int agnocast_ioctl_add_domain_bridge(
...
+  // Invariant: each cell (name, domain) belongs to at most one rule, and a rule pairs
+  // exactly two cells. r_a == r_b (non-NULL) means an existing rule already pairs exactly
+  // these two cells -- a re-declaration or the reverse direction, so just OR in the
+  // direction. Any other overlap (a cell already paired with a different cell) is a
+  // fan-out and is rejected. This is the one place that enforces one pair per cell.
+  // TODO: support >2 domains per topic by storing a domain group instead of a fixed pair.
+  struct domain_bridge_rule * r_a = find_domain_rule(name_a, ipc_ns, domain_a);
+  struct domain_bridge_rule * r_b = find_domain_rule(name_b, ipc_ns, domain_b);
+  if (r_a || r_b) {
+    if (r_a != r_b) return -EBUSY;
```
atsushi421: This rejection and the endpoint-already-joined one below both return a bare -EBUSY with no log, so userspace cannot tell which invariant it violated; e2e_test_domain_bridge.bash line 75 has to guess ("fresh kmod? cell already has an endpoint?"). agnocast_ioctl_remove_bridge already dev_warns on its rejection path.

Since this PR gives each of these paths a dedicated helper, it is a natural moment to add a dev_warn on each naming the topics, domains, and cause, leaving the return codes unchanged.

atsushi421 (commented):
Verified the "pure refactor, no behavior change" claim holds: the four ternaries in the new dev_info produce the same string as the old call for both values of from_is_a, `if (r_a != r_b) return -EBUSY;` is equivalent to the old if/else across all four r_a/r_b combinations, and lock scope, error codes, and log firing conditions are all preserved. The comments below are about contracts the split newly exposed and about coverage of the branches it touched.

---

agnocast_internal.h line 232 says "See agnocast_ioctl_add_domain_bridge for the rules on adding one." Those rules (the one-pair-per-cell invariant, the fan-out rejection, the endpoint-already-joined rejection) moved into add_domain_rule, a static helper that is not visible from the header and not reachable by grepping the referenced name. A reader following the pointer now lands in a function containing only canonicalization and the lock. Consider repointing it at add_domain_rule in agnocast_ioctl.c.

---

The checklist marks `scripts/test/run_requires_kernel_module_tests.bash` as required but leaves it unchecked. This PR changes kernel ioctl code, and that suite is the labeled integration gate for kmod changes. Running it before merge, plus `e2e_test_domain_bridge.bash` with `E2E_REMAP` set (the only e2e mode that exercises the swapped canonical orientation), would also cover the from_is_a == false gap noted inline.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in [d401dee1](https://github.com/autowarefoundation/agnocast/commit/d401dee18c08c18760d87d63181a74149d65c0c5).

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in [ce01699d](https://github.com/autowarefoundation/agnocast/commit/ce01699d202c84c2c2f01218d04c0d61ca461830).

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in [47e5546f](https://github.com/autowarefoundation/agnocast/commit/47e5546fbfcdf595a4c780c6accdf45536ce63ce). I kept the argument order aligned with the entry point rather than moving to `name, ipc_ns, domain`, since the entry point's own signature is what separates each name from its domain.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in [5aaa4ba4](https://github.com/autowarefoundation/agnocast/commit/5aaa4ba4e40e5fbad8501838fde9aadaa4c46dbf).

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in [07f15c90](https://github.com/autowarefoundation/agnocast/commit/07f15c90e6ff056bf228272c26274e5e82b56238).

Koichi98:
Fixed in [4e126eb0](https://github.com/autowarefoundation/agnocast/commit/4e126eb01cab18b022aff172d545fd8afc5036b4).

`run_requires_kernel_module_tests.bash` and `e2e_test_domain_bridge.bash` with `E2E_REMAP` both pass.

atsushi421 (approved):

PR #1540, opened 2026-08-18
feat(agnocastlib): key the service response topic by the calling client

Koichi98:
## Description

A service response rides on a topic named after the service and the calling node. Both sides derived that name independently, so the client and the server had to agree on the rule. The bridge plugins are separately built shared libraries carrying their own copy of it, which makes the agreement a versioning constraint.

Send the name the caller already computed for its own subscription in the request instead, and have the server publish to whatever it is told. `RequestMeta::node_name` becomes `response_topic_name`.

Agnocast assumes fully qualified node names are unique, so the node name was enough to tell callers apart. A domain bridge breaks that: two ROS domains may legitimately hold the same node name, and the bridge merges the two response topics into one. Append the caller's id of its request publisher so the name stays per-client. The bridge merges the request topic's id space across both domains, which is what makes that id unique there.

## Related links

Builds on #1535, which introduced `RequestMeta` and the `client_gid` field beside the renamed one. Groundwork for the service domain bridge (#1547).

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Notes for reviewers

The field keeps its type and position in `RequestMeta`, so the wire layout is unchanged; only its meaning moves from "who is calling" to "where the answer goes". Mixed-vintage binaries therefore fail silently rather than loudly: an older client writes a node name where the newer server reads a topic name, and the call hangs with no error.

## Post-merge checklist

`need-patch-update`

### DIFF
--- a/src/agnocastlib/include/agnocast/agnocast_service.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_service.hpp
@@ -75,20 +75,19 @@ private:
   typename ServiceRequestSubscriber::SharedPtr subscriber_;
 
   typename ServiceResponsePublisher::SharedPtr get_or_create_publisher_for(
-    const std::string & node_name)
+    const std::string & response_topic_name)
   {
     typename ServiceResponsePublisher::SharedPtr pub;
     {
       std::lock_guard<std::mutex> lock(publishers_mtx_);
-      auto it = publishers_.find(node_name);
+      auto it = publishers_.find(response_topic_name);
       if (it == publishers_.end()) {
         std::visit(
-          [this, &pub, &node_name](auto * node) {
-            std::string topic_name = create_service_response_topic_name(service_name_, node_name);
+          [this, &pub, &response_topic_name](auto * node) {
             agnocast::PublisherOptions pub_options;
             pub = std::make_shared<ServiceResponsePublisher>(
-              node, topic_name, qos_, pub_options, to_publisher_role(role_));
-            publishers_[node_name] = pub;
+              node, response_topic_name, qos_, pub_options, to_publisher_role(role_));
+            publishers_[response_topic_name] = pub;
           },
           node_);
       } else {
@@ -102,7 +101,7 @@ private:
   auto wrap_basic_service_callback_for_subscriber(Func && callback)
   {
     return [this, callback = std::forward<Func>(callback)](ipc_shared_ptr<RequestT> && request) {
-      auto publisher = this->get_or_create_publisher_for(request->RequestMeta::node_name);
+      auto publisher = this->get_or_create_publisher_for(request->RequestMeta::response_topic_name);
 
       ipc_shared_ptr<ResponseT> response = publisher->borrow_loaned_message();
       response->ResponseMeta::seqno = request->RequestMeta::seqno;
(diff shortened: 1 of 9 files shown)


src/agnocastlib/include/agnocast/agnocast_utils.hpp
```diff
@@ -89,8 +89,12 @@ std::string create_shm_name(const pid_t pid);
 // `${AGNOCAST_TMPFS_DIR:-/dev/shm}/agnocast_type_registry/<ipc_ns_inode>/`.
 uint64_t get_self_ipc_ns_inode();
 std::string create_service_request_topic_name(const std::string & service_name);
+// A domain bridge merges two domains, where the same fully qualified node name may legitimately
+// appear twice, into one response topic. The publisher id separates such clients: it comes from a
+// counter the kernel module keeps per request topic, and the bridge merges that counter too.
```
kobayu858: This holds only while the request topic's rule is in place, and nothing checks that. The two rules go in as separate ioctls, so a run after the server is up registers the response prefix but gets `-EBUSY` for the request rule — then both domains' counters start at 0 and two `/talker` nodes compute the same response topic name, merged by the prefix rule. The tool exits non-zero so it's noticeable, and stopping every node and re-running fixes it, but the precondition isn't stated anywhere and the client says nothing when it doesn't hold.

src/agnocastlib/include/agnocast/agnocast_service.hpp
```diff
@@ -75,20 +75,19 @@ class BasicService : public std::enable_shared_from_this<BasicService<ServiceT>>
...
       if (it == publishers_.end()) {
         std::visit(
-          [this, &pub, &node_name](auto * node) {
-            std::string topic_name = create_service_response_topic_name(service_name_, node_name);
+          [this, &pub, &response_topic_name](auto * node) {
             agnocast::PublisherOptions pub_options;
             pub = std::make_shared<ServiceResponsePublisher>(
-              node, topic_name, qos_, pub_options, to_publisher_role(role_));
-            publishers_[node_name] = pub;
+              node, response_topic_name, qos_, pub_options, to_publisher_role(role_));
```
kobayu858, line 94: nit: the server used to derive this from its own `service_name_`, so the topic it published on was structurally guaranteed to be `/AGNOCAST_SRV_RESPONSE<my service>_SEP_…`. It now takes whatever the request carries, with no check — not even the prefix. A malformed name throws `InvalidTopicNameError` out of the request callback rather than out of the client that caused it. One prefix comparison would keep the invariant. Same shape in `agnocast_service.cpp:21`.

kobayu858:
nit: the description says "Append the caller's pid and the id of its request publisher" — the code appends the node name and the publisher id, no pid. The publisher id carries the uniqueness on its own, so the code looks right and it's the description that's stale.

src/agnocastlib/include/agnocast/agnocast_utils.hpp
Koichi98: You're right, and the comment was mine to get right. Fixed in eb2aca0b.

The comment now states the dependency, what breaks without it, and that nothing checks it. I left out the two-ioctl mechanics deliberately — the plan is a new ioctl that applies both rules under one `global_htables_rwsem` write, so the partial state can't arise; those details would go stale. That fix belongs in the registration path, not in this PR.

src/agnocastlib/include/agnocast/agnocast_service.hpp
Koichi98: Good catch. Fixed in b9a9cb55, but by catching rather than a prefix check — worth your read since it differs from what you suggested.

A prefix check leaves a hole (`/AGNOCAST_SRV_RESPONSE//bad` passes it and still throws). Catching closes that, and fixes something worse: `execute_agnocast_executable()` restores `can_be_taken_from()` *after* the callable, so an escaping exception leaves a MutuallyExclusive callback group permanently taken — silencing every callback in it, not just this service. Both request callbacks now log the offending name and drop the request.

Koichi98:
@kobayu858 
> The publisher id carries the uniqueness on its own, so the code looks right and it's the description that's stale.

Fixed! Thank you.

kobayu858 (approved):

PR #1542, opened 2026-08-18
feat(agnocast_kmod)[need-minor-update]: count subscribers reachable across a domain bridge

Koichi98:
## Description

`get_subscription_count()` reports only the caller's own domain, matching ROS 2. A publisher whose topic is covered by a domain bridge rule therefore cannot tell that its publications reach a subscriber at all: the peer's endpoints sit in the same `topic_struct` but are filtered out. A service client asking whether its server is up is the first caller that needs the other answer.

`AGNOCAST_GET_SUBSCRIBER_NUM_CMD` now also reports `ret_other_domain_subscriber_num`: the subscribers in the paired domain that this domain's publications reach. It is counted in the walk the ioctl already does, with `domain_delivery_allowed()` -- the predicate the publish path uses -- so a rule declared only in the opposite direction contributes nothing and the count cannot disagree with what delivery does. The existing fields keep their own-domain meaning, so `get_subscription_count()` is unchanged.

## Related links

None.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Post-merge checklist

`need-minor-update`

### DIFF
--- a/agnocast_kmod/agnocast.h
+++ b/agnocast_kmod/agnocast.h
@@ -155,6 +155,9 @@ union ioctl_get_subscriber_num_args {
     uint32_t ret_other_process_subscriber_num;
     uint32_t ret_same_process_subscriber_num;
     uint32_t ret_ros2_subscriber_num;
+    // Subscribers in the domain a bridge rule pairs this one with, counted only where the rule
+    // delivers this way round. Disjoint from the own-domain counts above.
+    uint32_t ret_other_domain_subscriber_num;
     bool ret_a2r_bridge_exist;
     bool ret_r2a_bridge_exist;
   };
--- a/agnocast_kmod/agnocast_ioctl.c
+++ b/agnocast_kmod/agnocast_ioctl.c
@@ -1381,6 +1381,7 @@ int agnocast_ioctl_get_subscriber_num(
   ioctl_ret->ret_other_process_subscriber_num = 0;
   ioctl_ret->ret_same_process_subscriber_num = 0;
   ioctl_ret->ret_ros2_subscriber_num = 0;
+  ioctl_ret->ret_other_domain_subscriber_num = 0;
   ioctl_ret->ret_a2r_bridge_exist = false;
   ioctl_ret->ret_r2a_bridge_exist = false;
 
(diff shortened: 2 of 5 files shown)


agnocast_kmod/agnocast.h
```diff
@@ -155,6 +155,9 @@ union ioctl_get_subscriber_num_args {
     uint32_t ret_other_process_subscriber_num;
     uint32_t ret_same_process_subscriber_num;
     uint32_t ret_ros2_subscriber_num;
+    // Subscribers in the domain a bridge rule pairs this one with, counted only where the rule
+    // delivers this way round. Disjoint from the own-domain counts above.
+    uint32_t ret_other_domain_subscriber_num;
     bool ret_a2r_bridge_exist;
     bool ret_r2a_bridge_exist;
   };
```
kobayu858, line 163: Needs the same field in the union in `agnocast_ioctl.hpp` — otherwise `_IOWR` encodes a different size on each side and the ioctl fails with EINVAL.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -1397,16 +1398,20 @@ int agnocast_ioctl_get_subscriber_num(
...
+  // Match ROS 2's get_subscription_count: the same/other-process counts report only same-domain
+  // subscribers. A bridge rule still delivers cross-domain (see the publish/receive paths), but a
+  // publisher does not count subscribers in another domain. Ungrouped topics hold only one domain,
+  // so this is a no-op for them.
   struct subscriber_info * sub_info;
   int bkt_sub;
   hash_for_each(wrapper->topic->sub_info_htable, bkt_sub, sub_info, node)
   {
     if (sub_info->domain_id != wrapper->domain_id) {
+      if (domain_delivery_allowed(wrapper->topic, wrapper->domain_id, sub_info->domain_id)) {
```
kobayu858, line 1412: Should this exclude `is_bridge` subscribers, or expose a flag like `ret_a2r_bridge_exist` so callers can? The same-domain branch below lets callers subtract the synthetic subscriber the A2R bridge registers; this count has no equivalent. With only an A2R bridge in the peer domain and no real subscriber, "is anything on the other side receiving me" answers 1.

kobayu858 (changes requested):

agnocast_kmod/agnocast_ioctl.c
Koichi98: Thank you for pointing out.
I've opened a new PR to address this
https://github.com/autowarefoundation/agnocast/pull/1555.

agnocast_kmod/agnocast.h
Koichi98: Fixed in https://github.com/autowarefoundation/agnocast/pull/1542/commits/138bbe7147fef9b1fb4490869654cbdefd7aac3d.

kobayu858 (approved):

PR #1544, opened 2026-08-18
feat(agnocast_cie_thread_configurator): parse kernel_threads and irqs config sections

atsushi421:
## Description

Part 2 of 4 splitting #1528 into reviewable units.

Adds parsers for two new optional sections in the thread configurator YAML:

- `kernel_threads` (`KernelThreadConfig`, `parse_kernel_threads()`): desired policy, nice/priority, and affinity per kernel thread comm. The schema follows `callback_groups`: CFS policies (`SCHED_OTHER`/`BATCH`/`IDLE`) take a `nice` key validated against [-20, 19], while `SCHED_FIFO`/`SCHED_RR` take a `priority` key validated against [1, 99], reusing the same helpers and rationale (the syscalls silently clamp or fail only at apply time). Validation: a non-empty `comm` of at most 15 characters is required (the kernel truncates `task->comm`, so a longer comm could never match a scanned thread), `kworker/*` comms are rejected, `nice`/`priority` require `policy` (and vice versa per policy class), `SCHED_DEADLINE` requires `runtime`/`period`/`deadline`, and duplicate comms are rejected.
- `irqs` (`IrqConfig`, `parse_irqs()`): desired affinity per IRQ number, with an optional expected `name` (the `/sys/kernel/irq/<N>/actions` content) that the apply step uses to detect IRQ renumbering across boots. A non-negative integer `irq` is required and duplicates are rejected.
- `UNMANAGEABLE` sentinel (`k_unmanageable`): marks attribute values the kernel or this tool cannot manage, as opposed to YAML null which is the user's opt-out. Both parse to "not applied". The sentinel applies only to attribute values (`policy`/`nice`/`priority`/`affinity`/`runtime`/`period`/`deadline`), not to the identity keys `comm` and `name`.

Both sections share `parse_affinity()` with the existing parsers, so affinity values must be a list, are bounds-checked against the machine CPU count, and come out sorted and deduplicated; a scalar value (e.g. a cpu-list string copied from a scan) is rejected instead of silently meaning "leave alone". A section that is present but not a list is rejected rather than parsed as empty, and error messages identify the failing entry by index when no comm/irq is available yet.

Missing or null sections yield empty vectors, so pre-existing YAML files keep working unchanged. Unit tests cover the validation rules.

The parsers gain callers in the follow-up PRs of the stack (prerun template generation and apply logic).

## Related links

- Split from #1528
- Stacked on #1543

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Notes for reviewers

The base branch is #1543 only to keep the stack linear; this PR does not use the system_scan module and can be reviewed independently.

## Post-merge checklist

**Important notes:**

### DIFF
--- a/src/agnocast_cie_thread_configurator/src/thread_config.cpp
+++ b/src/agnocast_cie_thread_configurator/src/thread_config.cpp
@@ -17,6 +17,16 @@ namespace agnocast_cie_thread_configurator
 namespace
 {
 
+// Unset = the attribute must not be applied: YAML null / absent key (the
+// user's opt-out) or the UNMANAGEABLE sentinel (a kernel/tool constraint).
+bool is_unset(const YAML::Node & node)
+{
+  if (!node || node.IsNull()) {
+    return true;
+  }
+  return node.IsScalar() && node.Scalar() == k_unmanageable;
+}
+
 constexpr int k_nice_min = -20;
 constexpr int k_nice_max = 19;
 constexpr int k_rt_priority_min = 1;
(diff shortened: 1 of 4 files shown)


src/agnocast_cie_thread_configurator/src/thread_config.cpp
```diff
@@ -83,8 +107,8 @@ std::vector<int> parse_affinity(const YAML::Node & entry, const std::string & en
 {
   const YAML::Node affinity = entry["affinity"];
   std::vector<int> cpus;
-  // Absent or null means "do not manage affinity".
-  if (!affinity || affinity.IsNull()) {
+  // Unset (absent, null, or UNMANAGEABLE) means "do not manage affinity".
+  if (is_unset(affinity)) {
```
kobayu858: Routing `parse_affinity()` through `is_unset()` extends the `UNMANAGEABLE` sentinel to `callback_groups` and `non_ros_threads`, which the header scopes to "kernel_threads/irqs attribute values". Before this change `affinity: UNMANAGEABLE` on a callback group threw `'affinity' must be a list of CPU numbers`; now it silently means "leave affinity alone", which is the silently-dead configuration this PR rejects elsewhere.
No existing valid YAML changes meaning, so this is only a widening, but the sentinel is documented as something the tool writes, and the tool never writes into these two sections. Passing a flag to the shared helpers would keep the scope the header states. Minor, same cause: `nice: UNMANAGEABLE` now reports `requires 'nice'` even though the key is present.

src/agnocast_cie_thread_configurator/src/thread_config.cpp
```diff
@@ -246,4 +276,170 @@ void parse_yaml(
...
+    if (cfg.comm.empty()) {
+      throw std::runtime_error(entry_pos + " is missing a non-empty 'comm'");
+    }
+    if (is_kworker_comm(cfg.comm)) {
+      throw std::runtime_error(
+        "kernel_threads entry '" + cfg.comm +
+        "' is not manageable: kworker comms are ephemeral and mutate at runtime, so they cannot "
+        "be matched reliably");
+    }
+    if (cfg.comm.size() > k_max_comm_len) {
```
kobayu858: Kernel threads are not truncated to 15 characters, so this check rejects comms that the merged `scan_kernel_threads()` actually returns. Measured on Ubuntu 22.04.5 / kernel 6.5.0-44: 31 of the 184 non-kworker kthreads have comms longer than 15 characters, including `agnocast_exit_worker` (20). `/proc/<pid>/stat` and `/proc/<pid>/comm` agree on all of them, so the scanner reports the full name and this parser refuses to accept configuration for it. Userspace comms are truncated to 15 as the comment says, but kernel threads are not, so the premise holds only for the former.
Could you share which kernel this was verified on?

src/agnocast_cie_thread_configurator/src/thread_config.cpp
```diff
@@ -246,4 +276,170 @@ void parse_yaml(
...
+        "comms to " +
+        std::to_string(k_max_comm_len) + " characters");
+    }
+
+    cfg.affinity = parse_affinity(kt, "comm=" + cfg.comm);
+
+    const bool has_policy = !is_unset(kt["policy"]);
+    const bool has_nice = !is_unset(kt["nice"]);
+    const bool has_priority = !is_unset(kt["priority"]);
+    if (!has_policy) {
```
kobayu858: The guard covers `nice` and `priority` but not `runtime`/`period`/`deadline`, so an entry carrying the full SCHED_DEADLINE triple with `policy` forgotten parses cleanly and yields `is_managed() == false`. That is the same mistake as a forgotten `policy` next to `priority`, which is rejected two lines below, and it is the silently dead configuration the comm-length and kworker checks were added to prevent. The comment at line 364 notes these fields are always hand-written and easy to forget, which is exactly when `policy` goes missing too.

src/agnocast_cie_thread_configurator/src/thread_config.cpp
```diff
@@ -246,4 +276,170 @@ void parse_yaml(
...
+    const std::string entry_pos = "irqs entry #" + std::to_string(i);
+
+    if (!iq.IsMap()) {
+      throw std::runtime_error(entry_pos + " must be a mapping (e.g. '- irq: ...')");
+    }
+    if (!iq["irq"] || iq["irq"].IsNull()) {
+      throw std::runtime_error(entry_pos + " is missing a non-negative integer 'irq'");
+    }
+    try {
+      cfg.irq = iq["irq"].as<int>();
```
kobayu858: `as<int>()` in yaml-cpp auto-detects the base, so `irq: 010` parses as IRQ 8 and `irq: 0x10` as 16 (verified against yaml-cpp 0.7). `irq: 09` throws, so zero-padding a column fails loudly for some numbers and silently targets a different interrupt for others. IRQ numbers are the one field users copy by hand and are likely to align, and the merged `system_scan.cpp` reads them with `is_all_digits()` + `from_chars()`, so scanner and parser disagree on the same text. Reusing the base-10 path, or rejecting a leading zero with a message, would close it.

Same applies to `runtime`/`period`/`deadline`.

src/agnocast_cie_thread_configurator/src/thread_config.cpp
atsushi421: Fixed in e7baa291: is_unset() and the shared helpers now take an allow_unmanageable flag, so the sentinel is recognized only in the kernel_threads/irqs sections; in callback_groups/non_ros_threads it fails validation as before. The 'requires nice/priority' message now appends "(UNMANAGEABLE counts as unset)" when the key is present with the sentinel.

src/agnocast_cie_thread_configurator/src/thread_config.cpp
atsushi421: You are right, thanks. The check was derived from the task->comm limit (TASK_COMM_LEN) and had not been verified against /proc output on a specific kernel. Since Linux 5.17 (commit d26270061ae6) /proc/<pid>/comm and stat report a kthread's full name untruncated, and I reproduced this on 6.5.0-44 (e.g. rcu_tasks_kthread, 17 chars, appears in full in stat). Since scan_kernel_threads() reads stat, longer comms do match at apply time, so the check was rejecting valid configuration. Removed in e7baa291, with the test replaced by one that accepts a 20-char comm.

src/agnocast_cie_thread_configurator/src/thread_config.cpp
atsushi421: Fixed in e7baa291: the no-policy guard now covers nice, priority, runtime, period and deadline, so a SCHED_DEADLINE triple with 'policy' forgotten is rejected instead of parsing as unmanaged.

src/agnocast_cie_thread_configurator/src/thread_config.cpp
atsushi421: Fixed in e7baa291: 'irq' and the runtime/period/deadline fields are now parsed with a digits-only base-10 helper matching the scanners' is_all_digits() + from_chars() path, so 010 means decimal 10 and 0x10 is rejected with a clear message. nice/priority/affinity share the parsing of the pre-existing callback_groups/non_ros_threads sections, so I left them unchanged here to avoid altering those sections' behavior; happy to align them in a follow-up.

kobayu858 (approved):

PR #1545, opened 2026-08-18
feat(agnocast_cie_thread_configurator): emit kernel_threads and irqs sections in prerun template

atsushi421:
## Description

Part 3 of 4 splitting #1528 into reviewable units.

`prerun_node` now appends `kernel_threads` and `irqs` sections to the generated YAML template:

- Kernel threads are scanned via the system_scan module (#1543) and deduplicated by comm, since one template entry manages every thread sharing that comm. The scan is sorted by (comm, tid), so the kept observed values are the lowest tid's.
- Observed values become the initial desired values. Combined with compare-before-set in the apply step (part 4), unedited template entries stay no-ops.
- Values that cannot be represented or changed are emitted as `UNMANAGEABLE` (#1544): for policy/priority, unmapped policies and `SCHED_DEADLINE` (whose runtime/period/deadline cannot be recovered from `/proc`); for affinity, `PF_NO_SETAFFINITY` per-CPU kthreads and unreadable CPU lists. YAML null remains the user's explicit opt-out.

## Related links

- Split from #1528
- Stacked on #1544 (depends on both #1543 and #1544)

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Post-merge checklist

**Important notes:**

### DIFF
--- a/src/agnocast_cie_thread_configurator/src/prerun_node.cpp
+++ b/src/agnocast_cie_thread_configurator/src/prerun_node.cpp
@@ -8,9 +8,14 @@
 
 #include "agnocast_cie_config_msgs/msg/callback_group_info.hpp"
 
+#include <sched.h>
+#include <unistd.h>
+
+#include <algorithm>
 #include <filesystem>
 #include <fstream>
 #include <iostream>
+#include <optional>
 #include <set>
 #include <string>
 #include <utility>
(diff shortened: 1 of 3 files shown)


src/agnocast_cie_thread_configurator/src/prerun_node.cpp
```diff
@@ -181,6 +228,85 @@ void PrerunNode::dump_yaml_config(std::filesystem::path path)
     out << YAML::Newline;
   }
 
+  out << YAML::EndSeq;
+
+  // Add kernel_threads and irqs sections: observed values become the initial
+  // desired values (compare-before-set keeps unedited entries no-ops), and
+  // UNMANAGEABLE marks values that cannot be provided or changed (YAML null
+  // stays the user's opt-out).
```
kobayu858, line 236: Every entry whose policy is representable is emitted with a concrete policy, so it comes back is_managed() == true, while the two sections directly above deliberately emit `affinity: []` for "do not manage".
The generated file therefore declares desired values for every scanned kernel thread and IRQ before the user has edited anything, which is the opposite default to the pre-existing sections in the same function.
"Compare-before-set keeps unedited entries no-ops" holds only while nothing changes between generation and apply.

src/agnocast_cie_thread_configurator/src/prerun_node.cpp
```diff
@@ -181,6 +228,85 @@ void PrerunNode::dump_yaml_config(std::filesystem::path path)
...
+  }
+
+  out << YAML::EndSeq;
+
+  // Add irqs section
+  out << YAML::Key << "irqs";
+  out << YAML::Value << YAML::BeginSeq;
+
+  for (const auto & info : irqs) {
+    const auto cpus = parse_manageable_cpu_list(info.affinity);
```
kobayu858: There is no IRQ equivalent of the PF_NO_SETAFFINITY -> UNMANAGEABLE mapping used for kernel threads.
Kernel-managed interrupts (IRQD_AFFINITY_MANAGED, which covers nvme queues and most MSI-X) reject writes to smp_affinity_list; write_irq_affinity() returns -EIO for them.
Any host with an NVMe disk hits this: its queue interrupts (nvme0q0, nvme0q1, ...) are managed, and the template emits them as ordinary editable entries with a concrete affinity.
So a user who edits one of those lines gets an apply-time EIO with nothing in the template hinting the value was never settable.

src/agnocast_cie_thread_configurator/src/prerun_node.cpp
atsushi421: Thanks for the careful review. This asymmetry is intentional.

Concrete defaults are already the convention in this template: callback_groups and non_ros_threads emit `policy: SCHED_OTHER` and `nice: 0` for every entry (their parser makes policy mandatory), and only affinity defaults to "do not manage" there. For kernel threads a fixed default such as SCHED_OTHER would actively rewrite RT kthreads (e.g. migration/*, irq/* threads) on apply, so the only concrete default that is safe to emit is the observed value itself.

In addition, the generated template is not only an input for the apply step: CIE also uses it to capture the scheduling attributes of every scheduling element on the system. Emitting null placeholders would drop exactly the information the template exists to record, so each entry should carry concrete values.

Regarding drift between generation and apply: the config is declarative, so the system converging to what the YAML states is the intended contract rather than a hazard. Every emitted value originates from the machine itself, so applying an unedited template can only take the system back to a state it actually had, and compare-before-set makes the unchanged case a no-op. Individual entries can still be opted out by hand with `policy: ~` / `affinity: []`, which the parser accepts for these sections.

kobayu858: Understood, that all makes sense.

kobayu858 (approved):

PR #1546, opened 2026-08-18
feat(agnocast_cie_thread_configurator): apply kernel thread and IRQ configs with reapply support

atsushi421:
## Description

Part 4 of 4 splitting #1528 into reviewable units.

`thread_configurator_node` now applies the `kernel_threads` and `irqs` sections, both at startup and on every `~/reapply_config` request:

- Both sections are matched against a fresh `/proc` scan on every apply pass, with no carry-over: kernel threads and IRQs never announce themselves, so they live outside the announcement-driven accounting used for the other sections.
- Compare-before-set: an entry already in the desired state counts as applied without any syscall, so unedited template entries (part 3) stay no-ops and kernel-managed IRQs (which reject every write with EIO) see no spurious errors. Both sides of the comparison share the manageable-CPU bound (`parse_manageable_cpu_list`, moved from prerun into `system_scan`), so this holds even where the kernel's possible-CPU mask exceeds the present CPUs. One exception, documented in the srv: `SCHED_DEADLINE` parameters cannot be read back, so such entries are re-applied on every pass.
- Kernel threads: every running thread whose comm matches is configured through the existing syscall path (affinity factored out as `issue_affinity_syscalls` for affinity-only entries). An affinity request differing from a `PF_NO_SETAFFINITY` kthread's fixed one fails pointing at `UNMANAGEABLE`; an equal one is satisfied with only the policy applied. An unmatched comm is reported as skipped and can be retried via reapply.
- IRQs: the expected `actions` name is verified first to guard against IRQ renumbering across boots. If irqbalance is running, entries that need a write are failed instead of written, since irqbalance would silently rewrite them; in-sync entries still count as applied, so an unedited template stays a no-op even where irqbalance is enabled by default. The write uses raw open/write so errno-specific guidance is emitted (EIO: kernel-managed IRQ, EACCES/EPERM: missing capability, ENOENT: vanished IRQ).
- `ReapplyConfig.srv` gains `applied`/`failed`/`skipped` arrays for kernel threads and `applied`/`failed` for IRQs (an IRQ has no announce-later state, so a missing one is failed, not skipped).
- `setup_thread_configurator.bash` additionally grants `CAP_DAC_OVERRIDE`, required for writing `/proc/irq/<N>/smp_affinity_list`; an install by an older script version (only `cap_sys_nice`) is upgraded.
- E2E launch tests cover the reapply path for both sections.

## Related links

- Split from #1528
- Stacked on #1545 (merged)

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Notes for reviewers

The three main commits build up sequentially and each builds on its own: the `parse_manageable_cpu_list` refactor, the apply feature, then the capability grant. Within the feature commit, `apply_kernel_thread_configs()` and `apply_irq_configs()` are independent passes and can be reviewed function by function.

## Post-merge checklist

**Important notes:**

### DIFF
--- a/src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
+++ b/src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
@@ -2,6 +2,7 @@
 
 #include "agnocast_cie_thread_configurator/cie_thread_configurator.hpp"
 #include "agnocast_cie_thread_configurator/sched_deadline.hpp"
+#include "agnocast_cie_thread_configurator/system_scan.hpp"
 #include "agnocast_cie_thread_configurator/thread_config.hpp"
 #include "rclcpp/rclcpp.hpp"
 #include "yaml-cpp/yaml.h"
@@ -10,10 +11,13 @@
 #include "agnocast_cie_config_msgs/srv/reapply_config.hpp"
 
 #include <error.h>
+#include <fcntl.h>
 #include <sys/resource.h>
 #include <sys/time.h>
 #include <unistd.h>
 
+#include <algorithm>
+#include <cerrno>
 #include <filesystem>
 #include <fstream>
 #include <optional>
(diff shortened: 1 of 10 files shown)


src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
```diff
@@ -379,26 +469,33 @@ bool ThreadConfiguratorNode::issue_syscalls(const ThreadConfig & config, int64_t
...
+  return issue_affinity_syscalls(config.thread_str, config.policy, config.affinity, thread_id);
+}
+
+bool ThreadConfiguratorNode::issue_affinity_syscalls(
+  const std::string & thread_str, const std::string & policy, const std::vector<int> & affinity,
+  int64_t thread_id)
+{
+  if (affinity.size() > 0) {
+    if (policy == "SCHED_DEADLINE") {
+      if (!set_affinity_by_cgroup(thread_id, affinity)) {
```
kobayu858, line 487: On a cgroup v2 host this graceful path is unreachable: set_affinity_by_cgroup starts with std::filesystem::create_directory("/sys/fs/cgroup/cpuset/<n>"), the throwing overload, and with no cpuset controller the parent is missing, so ENOENT raises filesystem_error instead of returning false.
Ubuntu 22.04 and later default to the unified hierarchy, where /sys/fs/cgroup is cgroup2fs and /sys/fs/cgroup/cpuset does not exist, so the "Please disable cgroup v2 if used" message never reaches the user.

src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
```diff
@@ -407,6 +504,241 @@ bool ThreadConfiguratorNode::issue_syscalls(const ThreadConfig & config, int64_t
...
+    irq_configs_.begin(), irq_configs_.end(),
+    [](const IrqConfig & config) { return config.is_managed(); });
+  if (!any_managed) {
+    return outcome;
+  }
+
+  // A running irqbalance would periodically rewrite smp_affinity, silently
+  // undoing whatever is applied here, so applying would only fake success.
+  // Checked immediately before the writes to keep the detection fresh.
+  if (agnocast_cie_thread_configurator::process_with_comm_exists("irqbalance")) {
```
kobayu858: The irqbalance check sits before the per-entry compare-before-set, so entries that need no write at all are reported failed.
irqbalance is enabled by default on Ubuntu, and every entry in an unedited prerun template carries the observed affinity as the desired one, so the first startup after generating a template logs this ERROR and puts every device-backed IRQ into failed_irqs while issuing zero writes.

src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
atsushi421: Fixed in 4c82961e. set_affinity_by_cgroup now uses the non-throwing create_directory overload and returns false on ec || !created, so on a cgroup v2 host the ENOENT lands on the existing "Please disable cgroup v2 if used" message instead of a filesystem_error escaping into the constructor or a callback.

src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
atsushi421: Fixed in 4c82961e. The irqbalance gate now sits after the per-entry compare-before-set: in-sync entries are reported applied without any write, and only out-of-sync entries are failed (with a single aggregate ERROR). A first startup on an unedited template no longer logs an error or fills failed_irqs when irqbalance is running. The E2E test now asserts the in-sync-to-applied path directly instead of branching on the irqbalance state.

kobayu858 (approved):

PR #1547, opened 2026-08-18
feat(agnocast_kmod)[need-minor-update]: add prefix domain bridge rules

Koichi98:
## Description

A domain bridge rule names both cells it pairs, so every bridged topic has to be known when the rule is declared. That does not hold for a topic family whose members appear as processes start: an Agnocast service gives each caller its own response topic, named after the caller, and no configuration can list them in advance. Declaring the rule later does not work either, because grouping merges the two domains' id spaces and is rejected once an endpoint has joined.

`AGNOCAST_ADD_DOMAIN_BRIDGE_PREFIX_CMD` covers such a family with one declaration: every topic whose name starts with the prefix is paired with the identical name in the other domain. A prefix rule cannot rename, which is what lets one rule stand for names it has never seen.

Overlaps are rejected rather than resolved by precedence — in both directions against exact rules, and against nesting prefixes — so a lookup never has to choose between two candidates.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Notes for reviewers

Nothing on the publish or receive path changes: `domain_delivery_allowed()` reads the rule cached on the shared `topic_struct` and never looks at whether it is a prefix rule.

### Alternatives considered

**One response topic per service, callers correlating by client GID.** Naming the response topic after the service alone makes its rule exact, and prefix rules stop being needed at all; ROS 2 works this way, correlating a reply through `rmw_request_id_t` (client GUID and sequence number). Rejected here as reaching beyond bridging a service across domains: it changes the service wire format and the request/response path this PR does not touch, so it can regress existing behaviour, and it wakes every caller of a service on every response. Still an option to take later for the simpler design it gives.

**Each caller declaring the rule for its own response topic.** A caller knows the name, and could declare an exact rule synchronously before creating its subscription, keeping prefix matching out of the kmod. Rejected over what that does to rules: nothing removes a rule before module unload, so one accumulates per caller node name; a rejected declaration surfaces inside each caller's constructor instead of once at boot; and rule registration splits between the configuration and the callers.

**A domain group with an explicit name matcher.** Replacing the fixed pair with a set of domains plus a matcher (exact or prefix) subsumes this PR and lifts the two-domain limit `add_domain_rule()` carries a TODO for. Rejected as more than this use case needs -- grouping N domains divides one id space N ways and turns the direction check into a table -- but it remains the design to reach for if more than two domains ever have to be bridged.

## Post-merge checklist

`need-minor-update`

### DIFF
--- a/agnocast_kmod/agnocast.h
+++ b/agnocast_kmod/agnocast.h
@@ -297,6 +297,16 @@ struct ioctl_add_domain_bridge_args
   uint32_t to_domain;
 };
 
+// Bridges every topic whose name starts with topic_name_prefix, pairing it with the identical
+// name in the other domain. For topic families whose full names appear only at runtime and so
+// cannot be listed in a config -- an Agnocast service's per-client response topics.
+struct ioctl_add_domain_bridge_prefix_args
+{
+  struct name_info topic_name_prefix;
+  uint32_t from_domain;
+  uint32_t to_domain;
+};
+
 #define AGNOCAST_GET_VERSION_CMD _IOR(0xA6, 1, struct ioctl_get_version_args)
 #define AGNOCAST_ADD_PROCESS_CMD _IOWR(0xA6, 2, union ioctl_add_process_args)
 #define AGNOCAST_ADD_SUBSCRIBER_CMD _IOWR(0xA6, 3, union ioctl_add_subscriber_args)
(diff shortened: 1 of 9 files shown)


kobayu858:
> The lookup keeps scanning after a prefix match so that an exact rule for the same name would win. Registration rejects that overlap, so the precedence is a property the lookup does not depend on.

The description still says "The lookup keeps scanning after a prefix match so that an exact rule for the same name would win." The lookup now returns on the first prefix hit; the in-code comment states the new contract correctly, so it
is just the description that is stale. Worth fixing since that sentence is offered as the rationale for a property the code no longer has.

kobayu858 (approved):

Koichi98:
@kobayu858 
I've deleted the sentence! Thank you.

PR #1548, opened 2026-08-18
feat(agnocast): register domain bridge rules for services

Koichi98:
## Description

Agnocast's domain bridge relays topics between ROS domains within one IPC namespace, with no DDS and no bridge process. It covered pub/sub only. This PR extends it to services: a `services:` entry in the `domain_bridge` config, registered with the kmod at boot.

A service is not a kernel-module concept, so the config names the *service* and the kmod expands it into the two rules Agnocast needs:

| rule | topic | direction |
| --- | --- | --- |
| request | `/AGNOCAST_SRV_REQUEST<svc>` (exact) | clients → server |
| response | `/AGNOCAST_SRV_RESPONSE<svc>%` (prefix) | server → clients |

The response rule has to be a prefix: each client appends its node name and request-publisher id, so those full names exist only at runtime and cannot be listed in advance.

A `remap` renames the request topic only. The response prefix keeps the client-side name on both sides, because the client dictates that name and the server publishes to whatever the request asked for.

No kernel module change: the ioctl this calls arrives via the base branch.

## Related links

Stacked on #1569. Builds on #1533, #1535, #1539, #1542, #1547, #1555, #1557, #1564 and #1566, all merged.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

`colcon test` for `ros2agnocast_discovery_agent` covers the config parsing and the registration. The kmod is not touched, so the kunit box does not apply.

Both new scripts pass:

- `e2e_test_domain_bridge_service.bash` — the plain and the renamed service in turn, three callers each with the sums matched and none unmatched.
- `e2e_test_service_via_ros2_domain_bridge.bash` — the service reaches both domains and both answers come back.

They were run against a kmod predating #1569, which is sound because nothing here depends on the separator: the config names the service and the kmod builds the topic names.

## Notes for reviewers

`e2e_test_service_via_ros2_domain_bridge.bash` covers a different path — the Agnocast service bridge (A2R / R2A) plus an external relay over DDS, for cross-IPC-namespace and cross-ECU services. That path needs no change to Agnocast. The script skips unless a `domain_bridge` with generic service support happens to be installed, and nothing here should be added to `package.xml`.

A `services:` block cannot be shared with the external `domain_bridge` node: that node puts the server in `from_domain` and renames the callers' side, the reverse of the rules here. The systemd README says so.

### Known gaps, tracked separately

- A server keeps one response publisher per caller and never drops it, so the map, the kernel-side registration and the shared memory each publisher reserves grow with every caller ever served. Pre-existing.
- `topic_local_id` is never reclaimed, so a long-lived server on a merged request topic exhausts the 4096-id space sooner.

## Post-merge checklist

`need-patch-update`

### DIFF
--- a/scripts/test/e2e_test_domain_bridge_service.bash
+++ b/scripts/test/e2e_test_domain_bridge_service.bash
@@ -1,0 +1,179 @@
+#!/bin/bash
+# Cross-domain zero-copy e2e for a *service*: sample server in one domain, sample clients in the
+# other, one IPC namespace, no DDS and no bridge node. One `services:` entry becomes an exact rule
+# for /AGNOCAST_SRV_REQUEST<svc> and a prefix rule for /AGNOCAST_SRV_RESPONSE<svc>%, whose
+# per-client names exist only at runtime.
+#
+# The cross-domain client and a same-named client in the server's domain run CONCURRENTLY: only
+# while both are live can a shared response topic show itself, as "bad entry id" in both logs.
+#
+# Two exchanges run in turn, plain and renamed, with different service names so neither rule
+# claims a cell of the other. Run on a freshly loaded kmod: a rule must precede its endpoints.
+
+set -uo pipefail
+
+if ! grep -q "^agnocast " /proc/modules; then
+    echo "ERROR: agnocast kernel module is not loaded." >&2
+    echo "Load it first: sudo insmod agnocast_kmod/agnocast.ko" >&2
+    exit 1
+fi
+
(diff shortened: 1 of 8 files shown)


Koichi98:
Closing this: with these rules in place, a client and a server in the same IPC namespace reach each other over both the kmod zero-copy rule and the A2R/R2A service bridge over DDS at once. Tracked in #1579; this work can be reopened once the domain bridge has a userspace side.

PR #1550, opened 2026-08-18
refactor(agnocastlib): consolidate service typesupport loading

bdm-k:
## Description

This PR extracts duplicated service type support loading logic into a single `load_service_typesupport()` utility function, reducing code duplication. `ServiceTsBundle` is the struct to encapsulate all service typesupport data and is the return type of `load_service_typesupport()`. Four classes are refactored to use the function:

- `agnocast::vendor_rclcpp::GenericService`
- `agnocast::vendor_rclcpp::GenericClient`
- `agnocast::GenericService`
- `agnocast::GenericClient`

## How was this PR tested?

- [ ] Autoware
- [ ] sample application
- [x] service E2E test

## Post-merge checklist

### DIFF
--- a/src/agnocastlib/include/agnocast/agnocast_utils.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_utils.hpp
@@ -3,12 +3,6 @@
 #include "agnocast/agnocast_ioctl.hpp"
 #include "rclcpp/rclcpp.hpp"
 
-#include <rcpputils/shared_library.hpp>
-#include <rosidl_typesupport_introspection_cpp/message_introspection.hpp>
-
-#include <rosidl_runtime_c/service_type_support_struct.h>
-
-#include <memory>
 #include <string>
 
 namespace agnocast
@@ -102,16 +96,5 @@ uint64_t agnocast_get_timestamp();
 // Returns a pointer to the inner node handle that can be used for the TRACEPOINT macro.
 const void * get_node_base_address(agnocast::Node * node);
 const void * get_node_base_address(rclcpp::Node * node);
-
-struct ServiceTsBundle
-{
-  std::shared_ptr<rcpputils::SharedLibrary> ts_lib;
-  std::shared_ptr<rcpputils::SharedLibrary> ts_lib_introspection;
-  const rosidl_service_type_support_t * service_ts{nullptr};
-  const rosidl_typesupport_introspection_cpp::MessageMembers * request_members{nullptr};
-  const rosidl_typesupport_introspection_cpp::MessageMembers * response_members{nullptr};
-};
-
-ServiceTsBundle load_service_typesupport(const std::string & service_type);
 
 }  // namespace agnocast
--- a/src/agnocastlib/include/agnocast/agnocast_utils.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_utils.hpp
@@ -108,7 +108,6 @@ struct ServiceTsBundle
   std::shared_ptr<rcpputils::SharedLibrary> ts_lib;
   std::shared_ptr<rcpputils::SharedLibrary> ts_lib_introspection;
   const rosidl_service_type_support_t * service_ts{nullptr};
-  const rosidl_service_type_support_t * service_ts_introspection{nullptr};
   const rosidl_typesupport_introspection_cpp::MessageMembers * request_members{nullptr};
   const rosidl_typesupport_introspection_cpp::MessageMembers * response_members{nullptr};
 };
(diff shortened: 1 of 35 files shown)


src/agnocastlib/include/agnocast/agnocast_utils.hpp
```diff
@@ -97,4 +103,16 @@ uint64_t agnocast_get_timestamp();
 const void * get_node_base_address(agnocast::Node * node);
 const void * get_node_base_address(rclcpp::Node * node);
 
+struct ServiceTsBundle
+{
+  std::shared_ptr<rcpputils::SharedLibrary> ts_lib;
+  std::shared_ptr<rcpputils::SharedLibrary> ts_lib_introspection;
+  const rosidl_service_type_support_t * service_ts{nullptr};
+  const rosidl_service_type_support_t * service_ts_introspection{nullptr};
```
Koichi98: `service_ts_introspection` is never read outside `load_service_typesupport()` itself, so it can stay a local instead of a public struct field.

src/agnocastlib/src/agnocast_utils.cpp
```diff
@@ -136,4 +146,153 @@ const void * get_node_base_address(rclcpp::Node * node)
...
+  bundle.service_ts_introspection = get_service_typesupport_handle(
+    service_type, ts_identifier_introspection, *bundle.ts_lib_introspection);
+#endif
+
+  const auto * service_members =
+    static_cast<const rosidl_typesupport_introspection_cpp::ServiceMembers *>(
+      bundle.service_ts_introspection->data);
+
+  bundle.request_members = service_members->request_members_;
+  bundle.response_members = service_members->response_members_;
```
Koichi98: I'd drop the second commit. Going through the service introspection handle buys one fewer `dlsym` and lets you skip the `+ "_Request"` concatenation, but that naming is asserted by rosidl itself (`rosidl_parser/definition.py:586-594`), and the generated `get_service_type_support_handle<T>()` fills its deliberately non-const `ServiceMembers` on first access with no synchronization, which the per-message handles (all `static const`) avoid entirely.

src/agnocastlib/include/agnocast/agnocast_utils.hpp
```diff
@@ -97,4 +103,16 @@ uint64_t agnocast_get_timestamp();
 const void * get_node_base_address(agnocast::Node * node);
 const void * get_node_base_address(rclcpp::Node * node);
 
+struct ServiceTsBundle
```
Koichi98: The repo already has `include/agnocast/internal/` for non-public API and `src/vendor/rclcpp/` for vendored code, and `agnocast_utils.hpp` is the most widely included header here, so this puts an implementation detail on the public `agnocast::` surface. A dedicated `internal/service_typesupport.{hpp,cpp}` pair keeps it off that surface.

src/agnocastlib/src/vendor/rclcpp/generic_client.cpp
```diff
@@ -92,10 +68,11 @@ GenericClient::SharedPtr GenericClient::create_generic_client(
...
 std::shared_ptr<void> GenericClient::create_response()
 {
-  void * response = new uint8_t[response_members_->size_of_];
-  response_members_->init_function(response, rosidl_runtime_cpp::MessageInitialization::ZERO);
-  return {response, [this](void * p) {
-            response_members_->fini_function(p);
+  const auto * response_members = service_ts_bundle_.response_members;
+  void * response = new uint8_t[response_members->size_of_];
+  response_members->init_function(response, rosidl_runtime_cpp::MessageInitialization::ZERO);
+  return {response, [response_members](void * p) {
```
Koichi98: Capturing the members pointer instead of `this` is the right direction, but `fini_function` lives in `ts_lib_introspection`, whose only owner is `service_ts_bundle_` in the object, so the deleter still needs the endpoint alive. Capturing the bundle by value makes it actually self-contained. Same in `generic_service.cpp:29` and `:56`.

Koichi98 (commented):
A few notes on the typesupport consolidation.

src/agnocastlib/include/agnocast/agnocast_utils.hpp
Koichi98: Fixed in 011f3467.

src/agnocastlib/src/agnocast_utils.cpp
Koichi98: Fixed in cbc178e3.

src/agnocastlib/include/agnocast/agnocast_utils.hpp
Koichi98: Fixed in 90baa646.

src/agnocastlib/src/vendor/rclcpp/generic_client.cpp
Koichi98: Fixed in 65f937d8.

Koichi98 (approved):

PR #1551, opened 2026-08-19
feat(agnocastlib): support service introspection on the server side

bdm-k:
## Description

Publishes `<service>/_service_event` for `agnocast::Service<T>`, so a service call can be observed with `ros2 service echo`. The server-side events `REQUEST_RECEIVED` and `RESPONSE_SENT` are published; `REQUEST_SENT` and `RESPONSE_RECEIVED` come from the client and are left to a follow-up. `GenericService` is not covered (#1577).

Introspection stays off until `agnocast::Service<T>::configure_introspection(clock, qos, state)` is called, mirroring `rclcpp::Service`: the clock and the QoS take effect when the event publisher is created, and the state can move between `OFF`, `METADATA` and `CONTENTS` at any time.

The whole feature is gated on rclcpp 21+, since `service_msgs` and the event typesupport hooks in `rosidl_service_type_support_t` do not exist on Humble.

`ros2 agnocast generate-bridge-plugins` also learns about the `<Srv>_Event` messages, so a typed bridge plugin can be generated for the event topic. The generic bridge handles it without one.

## How was this PR tested?

- [ ] Autoware
- [] kunit tests (required when modifying the kernel module)
- [ ] sample application

## Post-merge checklist

### DIFF
--- a/src/agnocastlib/CMakeLists.txt
+++ b/src/agnocastlib/CMakeLists.txt
@@ -45,7 +45,7 @@ add_library(agnocast SHARED
   src/agnocast_smart_pointer.cpp src/agnocast_callback_info.cpp src/agnocast_epoll.cpp src/agnocast_epoll_update_dispatcher.cpp src/agnocast_timer_info.cpp src/agnocast_timer.cpp src/agnocast_executor.cpp
   src/agnocast_single_threaded_executor.cpp src/agnocast_multi_threaded_executor.cpp
   src/agnocast_callback_isolated_executor.cpp
-  src/agnocast_tracepoint_wrapper.c src/agnocast_client.cpp src/agnocast_service.cpp
+  src/agnocast_tracepoint_wrapper.c src/agnocast_client.cpp src/agnocast_service.cpp src/agnocast_service_event_publisher.cpp
   src/node/agnocast_node.cpp src/node/agnocast_arguments.cpp src/node/agnocast_context.cpp src/node/agnocast_signal_handler.cpp
   src/node/agnocast_only_executor.cpp src/node/agnocast_only_single_threaded_executor.cpp src/node/agnocast_only_multi_threaded_executor.cpp
   src/node/agnocast_only_callback_isolated_executor.cpp src/node/agnocast_parameter_service.cpp src/node/agnocast_parameter_client.cpp
--- a/src/agnocastlib/CMakeLists.txt
+++ b/src/agnocastlib/CMakeLists.txt
@@ -168,7 +168,8 @@ if(BUILD_TESTING)
     test/unit/test_daemon_bridge_uds.cpp
     test/unit/test_agnocast_bridge_uds.cpp
     test/unit/test_discovery_agent_auto_fork.cpp
-    test/unit/test_bridge_node_name.cpp)
+    test/unit/test_bridge_node_name.cpp
+    test/unit/test_agnocast_service_event_publisher.cpp)
   target_include_directories(test_unit_${PROJECT_NAME} PRIVATE include src)
   target_link_libraries(test_unit_${PROJECT_NAME} agnocast)
   ament_target_dependencies(test_unit_${PROJECT_NAME} std_msgs diagnostic_msgs diagnostic_updater rosidl_typesupport_cpp)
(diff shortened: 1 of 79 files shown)


Koichi98 (commented):
Note that `change_state()` has no caller yet — `configure_introspection()` at `agnocast_service.hpp:325` is a TODO — so `state_` never leaves `Off` and none of this code runs today; most of my comments below are about what breaks once it is wired up, not about current behaviour.

src/agnocastlib/CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ find_package(rclcpp_components REQUIRED)
 find_package(rcl_yaml_param_parser REQUIRED)
 find_package(rcpputils REQUIRED)
 find_package(rmw REQUIRED)
+find_package(service_msgs REQUIRED)
```
Koichi98: `service_msgs` first ships in Iron, so this breaks the Humble half of the build matrix declared at `.github/workflows/build-and-test-agnocastlib-components-heaphook.yaml:22` — I reproduced it as `CMake Error at CMakeLists.txt:26`. Humble also has neither `event_message_create_handle_function` in `rosidl_service_type_support_t` nor `rcl/service_introspection.h`, so the whole feature needs an `RCLCPP_VERSION_MAJOR >= 21` gate rather than just a conditional dependency.

src/agnocastlib/include/agnocast/agnocast_service.hpp
```diff
@@ -172,6 +224,9 @@ class BasicService : public std::enable_shared_from_this<BasicService<ServiceT>>
         subscriber_role);
     }
 
+    event_publisher_ = std::make_shared<ServiceEventPublisher>(
```
Koichi98: An Agnocast subscription callback is runnable as soon as it is registered, per the comment at `src/agnocastlib/src/agnocast_executor.cpp:64-66`, so on a node already added to a spinning executor a request can reach `publish_request_received_event()` at line 115 while `event_publisher_` is still null. The same ordering is wrong on teardown: `event_publisher_` is declared after `subscriber_` at `:78-79`, so it is destroyed first while the callback that dereferences it is still registered. Creating it before the subscription and declaring it before `subscriber_` closes both.

src/agnocastlib/include/agnocast/agnocast_service_event_publisher.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include <memory>
+#include <mutex>
+#include <string>
+#include <utility>
+#include <variant>
+
+namespace agnocast
+{
+
+enum class ServiceIntrospectionState : uint8_t {
```
Koichi98: `ServiceIntrospectionState` duplicates `rcl_service_introspection_state_t` — same three values in the same order — which the public `configure_introspection()` already takes at `agnocast_service.hpp:323`, and nothing converts between the two. Nothing in the repo includes `rcl/service_introspection.h` either, so that type only resolves today through `rclcpp/rclcpp.hpp` → `rclcpp/service.hpp`. Including the header and using the rcl type throughout removes both.

src/agnocastlib/src/agnocast_service_event_publisher.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+ServiceEventPublisher::ServiceEventPublisher(
+  std::variant<rclcpp::Node *, agnocast::Node *> node, const std::string & service_name,
+  const std::string & service_type, const rclcpp::QoS & qos, const rclcpp::Clock::SharedPtr & clock)
+: node_(std::move(node)),
+  event_topic_name_(service_name + "/_service_event"),
+  event_topic_type_(service_type + "_Event"),
+  event_publisher_qos_(qos),
+  clock_(clock),
+  ts_bundle_(load_service_typesupport(service_type))
```
Koichi98: Loading the typesupport here makes every typed service's construction depend on the event typesupport being resolvable, and `load_service_typesupport()` throws per `internal/service_typesupport.hpp:28-29`, so a service that never enables introspection can now fail to construct for a reason unrelated to it. Deferring the load to the first transition out of OFF also puts it where `clock` and `qos_service_event_pub` belong: rcl applies both when the event publisher is created, so once `configure_introspection()` is implemented they have to move out of the constructor anyway.

src/agnocastlib/CMakeLists.txt
Koichi98: Fixed in e80b4364.

src/agnocastlib/include/agnocast/agnocast_service.hpp
Koichi98: Fixed in 0d5b1463.

src/agnocastlib/include/agnocast/agnocast_service_event_publisher.hpp
Koichi98: Fixed in 009b2514.

src/agnocastlib/src/agnocast_service_event_publisher.cpp
Koichi98: Fixed in 3a9261d9.

Koichi98:
**Plan:** merge this once introspection actually runs, so I've wired up `configure_introspection()` and added tests on top of your event publisher.

`change_state()` becomes `configure(clock, qos, state)`, so the clock and the QoS arrive where rcl applies them instead of being fixed in the constructor. The integration test drives a real call and reads the events off `<service>/_service_event`; it is labelled `requires_kernel_module` and I ran it against a Jazzy build with the module loaded.

Koichi98 (approved):

PR #1552, opened 2026-08-19
fix(agnocastlib): break the cross-IPC-namespace service bridge deadlock

Koichi98:
## Description

Two IPC namespaces sharing one Agnocast service deadlock: the service side needs an **R2A** bridge but waits for a ROS 2 client that only the client side's A2R creates, and the client side needs **A2R** but waits for the ROS 2 service that only R2A creates. Pub/sub breaks the same cycle with a TTL'd forced lease from the per-namespace discovery agent (`DAEMON_FORCE_TTL`); services never got one.

### Change

- **`BridgeMsgType::DaemonService`** — the service counterpart of `DaemonPubSub`, carrying a service name and nothing else (260 bytes, size-checked on both sides). The type comes from the local endpoint's own `BridgeMsgType::Service` request, since an Agnocast service registers none.
- **Only the R2A half is leased.** That is enough: the ROS 2 service it publishes is the term the client side's (3) was missing, so A2R follows on the ordinary demand rule. Leasing A2R too would let it run before its peer's ROS 2 service exists, and a request dropped in that window hangs the caller (#1576).
- **A2R bridges are now demand-gated on a local Agnocast client**, finishing on that side what #1533 started on R2A. Without it a cross-namespace pair holds itself up, each bridge counting the other's DDS endpoint as demand.

Drive-by: `bridge_decider.py` now skips `/AGNOCAST_SRV_*` in the pub/sub path — defensive only, since those topics carry no type name and `decide_bridges()` already dropped them.

## Related links

- #1576 — leasing the A2R half too becomes safe once an undeliverable request is retried instead of dropped.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application
- [x] `bash scripts/test/e2e_test_cross_ns_service.bash` 

## Notes for reviewers

**Only Jazzy and newer deadlock** — pre-Jazzy rclcpp has no service-client count, so `ros2_client_exists()` is unconditionally `true` and R2A is ungated.

**The A2R demand gate changes pre-existing behaviour** — an A2R bridge used to outlive its Agnocast clients while a same-named ROS 2 service stayed on the graph, and now goes with them; a client created after the last one left can meet the same window the first client already meets at startup (#1576).

## Post-merge checklist

`need-patch-update`

### DIFF
--- a/src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
+++ b/src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
@@ -517,7 +517,7 @@ void ServiceBridgeItem::handle_request(const BridgeMsgServicePayload & payload)
   }
 }
 
-void ServiceBridgeItem::handle_daemon_request(const BridgeMsgDaemonServicePayload &)
+void ServiceBridgeItem::handle_daemon_request()
 {
   r2a_forced_until_ = daemon_force_deadline(std::chrono::steady_clock::now());
 }
--- a/src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
+++ b/src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
@@ -230,6 +230,12 @@ bool ServiceBridgeItem::ros2_client_exists(const ServiceBridgeDeps & deps)
   }
 }
 
+bool ServiceBridgeItem::r2a_forced() const
+{
+  return r2a_forced_until_.has_value() &&
+         is_daemon_force_active(*r2a_forced_until_, std::chrono::steady_clock::now());
+}
+
 // Returns false if the target Agnocast service does not exist or if an error occurs while checking
 // it (the reason will be set in the error string). "Two or more services share this name" is one
 // such error: the probe below reads at most one entry, so a name collision reads as absence rather
@@ -382,7 +388,7 @@ void ServiceBridgeItem::update_configuration(const BridgeMsgServicePayload & pay
 // Stays in R2A, or takes arrow (6) back to PENDING.
 void ServiceBridgeItem::check_and_update_r2a(const ServiceBridgeDeps & deps)
 {
-  if (agno_service_exists() && ros2_client_exists(deps)) {
+  if (agno_service_exists() && (ros2_client_exists(deps) || r2a_forced())) {
     return;
   }
 
@@ -412,7 +418,7 @@ void ServiceBridgeItem::check_and_update_a2r(const ServiceBridgeDeps & deps)
 {
   if (may_start_r2a_bridge_ && agno_service_exists()) {
     set_error_string("An Agnocast service now owns this name");
-  } else if (ros2_service_exists(deps)) {
+  } else if (ros2_service_exists(deps) && agno_client_exists()) {
     return;
   }
 
@@ -450,7 +456,7 @@ void ServiceBridgeItem::check_and_update_pending(const ServiceBridgeDeps & deps)
     }
 
     // Arrow (5).
-    if (ros2_client_exists(deps) && start_r2a_bridge(deps) != 0) {
+    if ((ros2_client_exists(deps) || r2a_forced()) && start_r2a_bridge(deps) != 0) {
       RCLCPP_WARN(
         deps.logger, "Failed to start R2A service bridge for '%s': %s", service_name_.c_str(),
         get_error_string());
@@ -462,7 +468,7 @@ void ServiceBridgeItem::check_and_update_pending(const ServiceBridgeDeps & deps)
   release_shadow_node();
 
   // Arrow (3).
-  if (may_start_a2r_bridge_ && ros2_service_exists(deps)) {
+  if (may_start_a2r_bridge_ && ros2_service_exists(deps) && agno_client_exists()) {
     if (start_a2r_bridge(deps) != 0) {
       RCLCPP_WARN(
         deps.logger, "Failed to start A2R service bridge for '%s': %s", service_name_.c_str(),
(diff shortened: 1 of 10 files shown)


src/ros2agnocast_discovery_agent/test/test_bridge_decider.py
```diff
@@ -450,3 +460,107 @@ def test_dispatch_routes_to_per_domain_uds(monkeypatch):
...
+
+def test_service_decide_skips_cross_domain_match():
+    local = _state(topics=[_srv_topic('/add', subs=[_endpoint('/svc')], domain=0)])
+    remote = _state(host_uuid='OTHER', ipc_ns=222,
+                    topics=[_srv_topic('/add', pubs=[_endpoint('/cli')], domain=5)])
+
+    assert decide_service_bridges(local, {('OTHER', 222): remote}) == []
+
+
+def test_service_decide_emits_one_lease_when_both_roles_local():
```
kobayu858: Both sides carry a publisher and a subscriber here, so neither claim in the docstring is actually pinned. I ran the mutations against your branch: reversing the roles to `local_pubs and remote_subs` leaves this test green (the two neighbouring tests catch it), and emitting a client-side request as well leaves it green too, since requests collapse on `(service, domain)` and `ServiceBridgeRequest` carries no direction.

What it does pin is the behaviour in my other comment -- its remote holds `/rsvc`, so it asserts that a lease goes out to a namespace that is already self-sufficient. Adding the `remote_subs` guard fails this test and only this test, out of 125.

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/bridge_decider.py
```diff
@@ -160,6 +204,48 @@ def decide_bridges(local_state, remote_states) -> list:
...
+            if local_topic is None:
+                continue
+
+            domain_id = local_topic.domain_id
+
+            # Counting a bridge's own endpoints would keep every lease alive off its own bridge.
+            local_subs = [s for s in local_topic.subscribers if not s.is_bridge]
+            remote_pubs = [p for p in remote_topic.publishers if not p.is_bridge]
+
+            if local_subs and remote_pubs:
```
kobayu858, line 241: This only asks whether the remote holds a client, never whether it already has a service of its own. If the remote namespace runs both -- the same launch tree replicated into two IPC namespaces would do it -- its client is already served locally by name, so nothing needs to cross, but a lease still goes out and an R2A bridge starts here that nothing will ever call. 

That is the cost #1533 added `ros2_client_exists()` to bound, and the lease walks around it. Would `and not [s for s in remote_topic.subscribers if not s.is_bridge]` be right? In the topology this PR targets the remote has no service, so the guard does not block the case you are fixing.

src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
```diff
@@ -462,7 +468,7 @@ void ServiceBridgeItem::check_and_update_pending(const ServiceBridgeDeps & deps)
   release_shadow_node();
 
   // Arrow (3).
-  if (may_start_a2r_bridge_ && ros2_service_exists(deps)) {
+  if (may_start_a2r_bridge_ && ros2_service_exists(deps) && agno_client_exists()) {
```
kobayu858, line 471: The gate looks right -- without it the cross-namespace pair pins itself. The reviewer note frames it as a lifetime change, though, and it is a little more than that: once the last local Agnocast client goes, the bridge is torn down within a tick, and a client created afterwards that skips `wait_for_service()` publishes onto a request topic with no subscriber -- dropped, future never resolves, nothing logged. 

A client that does wait is fine, and arguably better off, because `service_is_ready_core()` counts the bridge's own subscriber. Worth saying in the note, since the window was previously startup-only and is now every turnover.

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/bridge_decider.py
Koichi98: Fair reading, and the test sent you there: its remote held both `/rsvc` and `/rcli`, so it asserted a topology we don't support. A service name has at most one provider per domain, as node names do -- `agno_service_exists()` reports absence once a second entry appears, and `service_is_ready_core()` warns. So the test goes rather than the condition; see the other thread.

Chasing it did turn up a real problem next door, though. The lease was also setting `may_start_r2a_bridge_` and creating items. That flag means "a `ServiceRole::Default` service registered here", which the daemon cannot see, and an item conjured from a lease carries no `service_type_`, so `start_r2a_bridge()` failed every tick and arrow (2) could not reap it. A lease now only renews `r2a_forced_until_` on an item a local registration created, so it stands in for `ros2_client_exists()` and nothing else. https://github.com/autowarefoundation/agnocast/pull/1552/commits/2fa116f96e5d88b32291d6f8fab6b608a36b75bd

src/ros2agnocast_discovery_agent/test/test_bridge_decider.py
Koichi98: You're right that it pinned neither claim, and its setup was worse than useless: a remote holding both a service and a client of one name is a configuration we don't support, so the test was asserting it as intended behaviour -- which is what sent you to the other thread. Replaced with `test_service_decide_collapses_clients_in_several_remotes`: two remotes each holding a client, one request out, pinning the `setdefault` you noted was uncovered. Role direction stays with the two neighbouring tests. https://github.com/autowarefoundation/agnocast/pull/1552/commits/0410a40777dbb895579d43e72fc5d4ecd49eccf4

src/agnocastlib/src/bridge/agnocast_service_bridge.cpp
Koichi98: Both true, but bounded on two sides. A live client keeps the bridge up -- `agno_client_exists()` counts its request-topic publisher, and the sweep only notices once it is gone -- so no client loses its subscriber mid-flight; only one created after the last has left can meet the gap. And that gap is the startup window recurring rather than a new one: `may_start_a2r_bridge_` is set by a client's own registration, so the bridge cannot exist before the first client, and today's first client already meets it.

"Never resolves" is rclcpp's behaviour too -- `async_send_request()` to an absent service leaves the future pending there as well, which is why it ships `remove_pending_request()`. What Agnocast lacks is that cleanup, i.e. #1576. So I have noted it in one line rather than as a new failure mode.

kobayu858 (dismissed):

kobayu858 (approved):

PR #1555, opened 2026-08-19
feat(agnocastlib): mark bridge-owned services and clients as bridges in kmod

Koichi98:
## Description

`ServiceRole` and `ClientRole` only had `Default` and `AgnocastOnly`, and the service implementation hardcoded `SubscriptionRole::AgnocastOnly` for its internal subscriber. A bridge-owned service was therefore indistinguishable from a real one on the request topic the two share -- `create_service_request_topic_name()` is a pure function of the service name, so both land in the same table with nothing to tell them apart. The pub/sub bridge has always marked its own endpoints; the service path did not.

Rename the enumerator to `BridgeInternal` -- the only non-default users are the A2R and R2A service bridges, so it already meant "the bridge's own entity" -- and derive the internal subscriber's and publisher's roles from it instead of hardcoding them, in the generated plugin template as well.

The mark is what #1557 keys on to keep an endpoint inside its own domain, and what `decide_bridges()` keys on to leave a bridge's own endpoints out of the demand it counts.

Because the kmod now guarantees that a bridge endpoint does not communicate with an endpoint in another domain, that fact no longer has to be carried out to user space: a peer-domain counterpart of `ret_a2r_bridge_exist` would describe something no endpoint ever has to handle. Since the communication does not happen in the first place, an endpoint has no occasion to deal with another domain's bridge subscriber, so `GET_SUBSCRIBER_NUM` does not report it at all rather than reporting it for the caller to subtract. The same-domain counts are unchanged, because delivery to a bridge in one's own domain does happen: it is still counted, and `ret_a2r_bridge_exist` still tells a caller to separate it from the ROS 2 subscribers it stands for.

## Related links

Answers https://github.com/autowarefoundation/agnocast/pull/1542#discussion_r3812254853.

Was stacked on #1557, which is merged, so this now sits on `main`.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Post-merge checklist

`need-patch-update` -- no ioctl struct or command changes, so kmod and agnocastlib stay compatible.

### DIFF
--- a/src/agnocastlib/include/agnocast/agnocast_service.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_service.hpp
@@ -22,10 +22,24 @@ namespace agnocast
 enum class ServiceRole : uint8_t {
   /// User-created service; issues an R2A bridge request.
   Default,
-  /// Used by the bridge plugin's own service; no bridge request is issued.
+  /// Used by the bridge implementation itself; marks the endpoints it creates as bridges in kmod
+  /// and issues no bridge request. A bridge service shares the request topic with a real one, so
+  /// that mark is what tells them apart.
   /// Not intended for direct use by application code.
-  AgnocastOnly,
+  BridgeInternal,
 };
+
+constexpr SubscriptionRole to_subscription_role(const ServiceRole role)
+{
+  return role == ServiceRole::BridgeInternal ? SubscriptionRole::BridgeInternal
+                                             : SubscriptionRole::AgnocastOnly;
+}
+
+constexpr PublisherRole to_publisher_role(const ServiceRole role)
+{
+  return role == ServiceRole::BridgeInternal ? PublisherRole::BridgeInternal
+                                             : PublisherRole::AgnocastOnly;
+}
 
 // Internal implementation - users should use agnocast::Service<ServiceT> instead.
 template <typename ServiceT>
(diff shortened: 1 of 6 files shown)


src/agnocastlib/include/agnocast/agnocast_service.hpp
```diff
@@ -22,11 +22,25 @@ namespace agnocast
 enum class ServiceRole : uint8_t {
   /// User-created service; issues an R2A bridge request.
   Default,
-  /// Used by the bridge plugin's own service; no bridge request is issued.
+  /// Used by the bridge implementation itself; marks the endpoints it creates as bridges in kmod
+  /// and issues no bridge request. A bridge service shares the request topic with a real one, so
+  /// that mark is what tells them apart.
   /// Not intended for direct use by application code.
-  AgnocastOnly,
+  BridgeInternal,
```
kobayu858, line 29: The label rationale covers kmod/agnocastlib compatibility, but the rename also
breaks source compatibility: `generate-bridge-plugins` writes  agnocast::ClientRole::AgnocastOnly` into a user-owned package, and nothing
prompts regeneration, so a patch upgrade fails their `colcon build`. `CONTRIBUTING.md` puts renaming user-facing APIs under `need-major-update`.

Keeping the patch label would work too, with an alias: `{ Default, BridgeInternal, AgnocastOnly = BridgeInternal }`. Either way is fine by me — your call.

kobayu858 (approved):

src/agnocastlib/include/agnocast/agnocast_service.hpp
Koichi98: Plugin packages are regenerated per version -- their `package.xml` is stamped with the generator's version -- so they are not what the version labels protect. No hand-written code names the enumerator, so `need-patch-update` stays.

PR #1557, opened 2026-08-20
fix(agnocast_kmod): keep bridge endpoints out of cross-domain delivery

Koichi98:
## Description

A bridge endpoint stands for its own domain's ROS 2 side: an A2R subscriber carries that domain's Agnocast traffic out to DDS, an R2A publisher carries DDS traffic back in. Delivering to one across a domain bridge rule therefore has the kmod carry Agnocast in one domain to ROS 2 in another -- and that is the external `domain_bridge` node's job, not the kmod's, which is what `decide_bridges()` already says.

`domain_delivery_allowed()` looked only at the two domains, so a rule reached those endpoints and a publication then crossed twice: once by the rule, once through the `domain_bridge` node the same YAML entry configures.

The predicate now rejects a crossing where either endpoint carries `is_bridge`. Same-domain delivery is untouched: a bridge still takes its own domain's traffic, grouped or not.

## Related links

None.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Notes for reviewers

`ret_other_domain_subscriber_num` from #1542 now excludes bridges by construction, so the count and delivery agree again. `GET_SUBSCRIBER_NUM` names a topic, not a publisher, so the predicate is asked with `pub_is_bridge = false`; every caller today is a `ClientRole::Default` client.

## Post-merge checklist

`need-patch-update` -- the ioctl structs and commands are unchanged, so kmod and agnocastlib stay compatible.

### DIFF
--- a/agnocast_kmod/agnocast_kunit/agnocast_kunit_publish_msg.c
+++ b/agnocast_kmod/agnocast_kunit/agnocast_kunit_publish_msg.c
@@ -723,8 +723,10 @@ void test_case_publish_msg_bridge_subscriber_in_other_domain_not_notified(struct
   topic_local_id_t publisher_id;
   uint64_t ret_addr;
   setup_publisher_in_domain(test, current->tgid, 1, is_bridge, &publisher_id, &ret_addr);
-  const int eventfd = 0;
-  setup_one_subscriber_in_domain_with_eventfd(test, 2, true, eventfd, false);
+  const int bridge_eventfd = 0;
+  const int plain_eventfd = 1;
+  setup_one_subscriber_in_domain_with_eventfd(test, 2, true, bridge_eventfd, false);
+  setup_one_subscriber_in_domain_with_eventfd(test, 2, is_bridge, plain_eventfd, false);
   union ioctl_publish_msg_args ioctl_publish_msg_ret;
 
   // Act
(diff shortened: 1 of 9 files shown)


agnocast_kmod/agnocast_kunit/agnocast_kunit_publish_msg.c
```diff
@@ -719,3 +710,75 @@ void test_case_publish_msg_no_process(struct kunit * test)
   // Assert
   KUNIT_EXPECT_EQ(test, ret, -EINVAL);
 }
+
+// PUBLISH resolves the wrapper by the caller's domain, so the publisher of a cross-domain case
+// has to be current->tgid.
+void test_case_publish_msg_bridge_subscriber_in_other_domain_not_notified(struct kunit * test)
```
kobayu858, line 716: nit: this passes even if grouping breaks — the endpoints would sit in separate structs and the count would be 0 for the wrong reason. A positive case (non-bridge subscriber in domain 2 *is* notified) would pin it down. The receive/take equivalents self-guard, since publish fails without grouping there.

Koichi98: Thanks for pointing out. Fixed in https://github.com/autowarefoundation/agnocast/pull/1557/commits/c62b68fdddfac7ef8d3eefcc33d8aadb25812312.

kobayu858 (approved):

PR #1559, opened 2026-08-20
fix(agnocast_kmod): reject a domain bridge direction added after an endpoint joined

Koichi98:
## Description

Declaring a domain bridge direction on an existing rule was accepted even after endpoints had joined, but nothing brought them up to date. A subscriber that joined while the direction was denied was kept out of its publishers' notify lists and skipped by `set_publisher_shm_info()`, so it never mapped the opposite-domain publisher's mempool. Once the direction was enabled it could be woken and handed entry addresses in that pool, and dereferencing them faults.

Refuse the declaration instead. This is the guarantee creating a rule already requires, so the two paths now agree, and the inconsistent state becomes unreachable rather than something to repair after the fact. A re-declaration that enables nothing new still succeeds, so re-running the registration tool with an unchanged config keeps working.

## Related links

None.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Post-merge checklist

`need-patch-update`

### DIFF
--- a/agnocast_kmod/agnocast_kunit/agnocast_kunit_domain_bridge.c
+++ b/agnocast_kmod/agnocast_kunit/agnocast_kunit_domain_bridge.c
@@ -236,34 +236,43 @@ void test_case_domain_bridge_direction_respected(struct kunit * test)
   KUNIT_EXPECT_EQ(test, signal_count_of(eventfd), 0);
 }
 
-void test_case_domain_bridge_late_reverse_direction_delivers(struct kunit * test)
+// Enabling a direction once endpoints exist cannot be made safe here: they were registered while
+// it was denied, so neither their notify lists nor their shm mappings account for it.
+void test_case_domain_bridge_late_reverse_direction_rejected(struct kunit * test)
 {
-  // Same setup as above, but 2 -> 1 is declared after both endpoints have registered, so the
-  // publisher's notify list was built while that direction was still denied.
+  // Arrange: 1 -> 2 declared, then both endpoints join.
   KUNIT_ASSERT_EQ(
     test, agnocast_ioctl_add_domain_bridge(TOPIC_NAME, TOPIC_NAME, 1, 2, current->nsproxy->ipc_ns),
     0);
-
-  agnocast_kunit_eventfd_reset();
-  const uint64_t msg_addr = setup_process_in_domain(test, current->tgid, 2);
-  const topic_local_id_t pub_id = add_publisher_for(test, current->tgid);
-
+  setup_process_in_domain(test, current->tgid, 2);
+  add_publisher_for(test, current->tgid);
   setup_process_in_domain(test, 1001, 1);
-  const int eventfd = 0;
-  add_subscriber_named_with_eventfd(test, 1001, TOPIC_NAME, eventfd);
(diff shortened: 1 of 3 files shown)


agnocast_kmod/agnocast_kunit/agnocast_kunit_domain_bridge.c
```diff
@@ -236,34 +236,43 @@ void test_case_domain_bridge_direction_respected(struct kunit * test)
   KUNIT_EXPECT_EQ(test, signal_count_of(eventfd), 0);
 }
 
-void test_case_domain_bridge_late_reverse_direction_delivers(struct kunit * test)
+// Enabling a direction once endpoints exist cannot be made safe here: they were registered while
+// it was denied, so neither their notify lists nor their shm mappings account for it.
+void test_case_domain_bridge_late_reverse_direction_rejected(struct kunit * test)
```
kobayu858, line 241: `late_reverse_direction_delivers` was the only case where `b_to_a == true` led to a publish reaching a subscriber. What is left checks flags only (`reverse_direction`, `reverse_first_declaration`) or the denied direction (`direction_respected`), so a change that stops `domain_delivery_allowed` from honoring `b_to_a` would pass. The deleted test's arrange works if the second `add_domain_bridge(2, 1)` moves before the endpoints join.

agnocast_kmod/agnocast_kunit/agnocast_kunit_domain_bridge.c
```diff
@@ -236,34 +236,43 @@ void test_case_domain_bridge_direction_respected(struct kunit * test)
...
   // Act
+  const int ret =
+    agnocast_ioctl_add_domain_bridge(TOPIC_NAME, TOPIC_NAME, 2, 1, current->nsproxy->ipc_ns);
+
+  // Assert
+  KUNIT_EXPECT_EQ(test, ret, -EBUSY);
+}
+
+// Re-running the registration tool with an unchanged config must keep working once nodes are up.
+void test_case_domain_bridge_redeclaration_is_idempotent(struct kunit * test)
```
kobayu858, line 293: `redeclaration_is_idempotent` exercises `from_is_a && a_to_b`; nothing reaches `!from_is_a && b_to_a`. That half is the re-run promise for a bidirectional config, so it seems worth one more case: declare both directions, add an endpoint, re-declare `2 -> 1`, expect `0`.

agnocast_kmod/agnocast_kunit/agnocast_kunit_domain_bridge.c
Koichi98: Restored the test with some fixes in https://github.com/autowarefoundation/agnocast/pull/1559/commits/f6ae0d6af255a0b38fa6ae763c9221b12a5772d8.

agnocast_kmod/agnocast_kunit/agnocast_kunit_domain_bridge.c
Koichi98: Added in https://github.com/autowarefoundation/agnocast/pull/1559/commits/f6ae0d6af255a0b38fa6ae763c9221b12a5772d8.

kobayu858 (approved):

PR #1566, opened 2026-08-25
feat(agnocast_kmod)[need-minor-update]: register both halves of a service domain bridge in one ioctl

Koichi98:
## Description

Review feedback on #1540 pointed out that the two ioctls a service domain bridge needs -- an exact rule for `/AGNOCAST_SRV_REQUEST<svc>` and a prefix rule for `/AGNOCAST_SRV_RESPONSE<svc>_SEP_` -- are not atomic together, and that the response rule standing alone is unsound: the id that keeps a response topic name per-client is only unique across domains the request rule has merged.

Merge them into one service ioctl. `AGNOCAST_ADD_DOMAIN_BRIDGE_SERVICE_CMD` takes `(service_name_from, service_name_to, from_domain, to_domain)`, expands it into both rules and applies them under one `global_htables_rwsem` write. It replaces `AGNOCAST_ADD_DOMAIN_BRIDGE_PREFIX_CMD` on the same command number; nothing on main called that ioctl.

There is no rollback path. Each half is validated into a plan first, and the plan's one allocation happens while the table is still untouched, so a rejection returns before anything has been applied. Only once both halves validate does either go in.

## Related links

Builds on #1540 (merged), which makes the server publish to the name the caller chose and gives that name a per-client id -- both assumptions of the response prefix rule.

Replaces the prefix ioctl from #1547. Groundwork for the service domain bridge (#1548).

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

**The kunit tests have not been run**, so the rewritten cases are unverified. Both module variants build clean under `-Wall -Werror`. They need a run before this leaves draft.

## Notes for reviewers

`struct domain_rule_plan` points `bool * direction` straight at `a_to_b` or `b_to_a` on the existing rule, which keeps applying a plan free of any tag for which case it is. It is hard to read, and I could not find an encoding that is clearer without being bigger. Leave it as is?

Known gap, not introduced here: `_SEP_` is legal in a service name, so the response prefix for `/foo` also covers the response topics of an unregistered `/foo_SEP_bar`. Those get grouped while their request topic never was, so two clients sharing a node name compute the same response topic name and a response can cross to the wrong one. Prefix rules have had this since #1547 and nothing reaches it until `services:` support lands. The fix is a separator that cannot appear in a ROS name; that changes agnocastlib's wire naming, so it goes in its own PR.

## Post-merge checklist

`need-minor-update`


agnocast_kmod/agnocast_ioctl.c
```diff
@@ -2558,21 +2596,63 @@ int agnocast_ioctl_add_domain_bridge(
...
+  // The response prefix uses the caller-side name on both sides: the client dictates the response
+  // topic name and the server publishes to whatever the request asked for, rename or not.
+  if (
+    snprintf(request_from, sizeof(request_from), "%s%s", SRV_REQUEST_PREFIX, service_name_from) >=
+      (int)sizeof(request_from) ||
+    snprintf(request_to, sizeof(request_to), "%s%s", SRV_REQUEST_PREFIX, service_name_to) >=
+      (int)sizeof(request_to) ||
+    snprintf(
+      response_prefix, sizeof(response_prefix), "%s%s%s", SRV_RESPONSE_PREFIX, service_name_from,
+      SRV_RESPONSE_SEP) >= (int)sizeof(response_prefix)) {
```
kobayu858, line 2629: The nested tests cover the case where both services are registered, and that's rejected. The gap is the other one: `_SEP_` is legal in a ROS service name, so bridge `/foo` and let an unregistered `/foo_SEP_bar` come up locally in each domain afterwards. Its response topics match this prefix and get grouped, but its request topic was never merged, so `client_publisher_id` is only unique within a domain — two clients with the same node name compute the identical response topic name and a response crosses to the wrong one. `any_topic_under_prefix` only catches it when those endpoints already exist at registration.

Koichi98: Agreed, and this is the case that fails silently.

It comes from the prefix rules in #1547, not from this PR, and nothing calls the ioctl until #1548. I will fix it in a separate PR: change the separator to a character that cannot be used in a ROS name, and reject that character in the ioctl.

PR #1571, opened 2026-08-26
feat(ros2agnocast_discovery_agent): honor bidirectional in the domain bridge config

Koichi98:
## Description

`bidirectional: true` in a `domain_bridge` YAML was read as one-way: `parse_domain_bridge_config` looked only at `from_domain` / `to_domain` / `remap`, so the reverse leg the external node creates had no Agnocast counterpart. A bidirectional entry now yields a second tuple with the domain ids swapped and nothing else, mirroring `DomainBridge::DomainBridge` (`domain_bridge.cpp:598-605`), which keeps the source name on the subscribe side and the remap on the publish side.

Both consumers pick this up unchanged: `register_domain_bridge` registers both, which the kmod folds into one rule with both directions enabled (already covered by `test_case_domain_bridge_bidirectional_delivers`), and the discovery agent sees the reverse `from` side as its own. A non-boolean value is now rejected, like `remap`.

## Related links

#1558 forces the A2R bridge on a rule's `from` side; with this, a bidirectional entry forces both sides instead of silently only one.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

`pytest` in `src/ros2agnocast_discovery_agent`: 80 passed. Dropping the reverse tuple fails two tests.

## Notes for reviewers

- No kernel module change: the kmod already accepts the reversed pair and folds it.
- The reverse leg is deliberately not a true inverse under a `remap`. That is the external node's behavior, and this file exists to describe what that node relays.

## Post-merge checklist

### DIFF
--- a/src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/domain_bridge_config.py
+++ b/src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/domain_bridge_config.py
@@ -30,6 +30,8 @@ def parse_domain_bridge_config(text):
     ``rules`` is a list of ``(from_topic, to_topic, from_domain, to_domain)``
     tuples. ``to_topic`` is the per-topic ``remap`` target (same ``domain_bridge``
     field the external node honors), or the source name when ``remap`` is absent.
+    A ``bidirectional`` topic yields two tuples; the kmod folds them into one rule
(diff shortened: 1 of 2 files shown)


Koichi98:
Folding this into #1558 instead; the change is small enough to ship with the forcing that consumes it.

PR #1572, opened 2026-08-26
fix(cie_thread_configurator): harden ROS_DOMAIN_ID parsing

### LINKED ISSUE #1401: agnocast_cie_thread_configurator: get_default_domain_id() crashes on empty/non-numeric ROS_DOMAIN_ID
`get_default_domain_id()` (`agnocast_cie_thread_configurator/src/util.cpp`) parses `ROS_DOMAIN_ID` with `std::stoul`, which throws uncaught on an empty (`ROS_DOMAIN_ID=`) or non-numeric value → crashes the node. It also only checks `!= nullptr`, not empty.

```cpp
size_t get_default_domain_id() {
  const char * env_value = std::getenv("ROS_DOMAIN_ID");
  if (env_value != nullptr) return static_cast<size_t>(std::stoul(env_value));  // throws on "" / non-numeric
  return 0;
}
```

## Repro

The function compiled standalone:

```bash
cat > /tmp/repro.cpp <<'EOF'
#include <cstdio>
#include <cstdlib>
#include <string>
size_t get_default_domain_id() {
  const char * env_value = std::getenv("ROS_DOMAIN_ID");
  if (env_value != nullptr) return static_cast<size_t>(std::stoul(env_value));
  return 0;
}
int main() { printf("domain=%zu\n", get_default_domain_id()); }
EOF
g++ -std=c++17 /tmp/repro.cpp -o /tmp/repro
ROS_DOMAIN_ID=7   /tmp/repro   # -> domain=7   (exit 0)
ROS_DOMAIN_ID=    /tmp/repro   # -> terminate (std::invalid_argument), Aborted (core dumped), exit 134
ROS_DOMAIN_ID=abc /tmp/repro   # -> same
```

Equivalently, launching the thread_configurator node with `ROS_DOMAIN_ID=` aborts at startup.

## Fix

Guard empty/invalid (fall back to domain 0). agnocastlib's `agnocast::get_ros_domain_id()` (added in #1398) already does this, so reuse it to also de-duplicate the parser.

VedantMadane:
﻿## Summary

`get_default_domain_id()` aborted when `ROS_DOMAIN_ID` was empty or non-numeric because `std::stoul` threw uncaught.

## Fix

- Parse with `strtoul`, treat empty / trailing junk / out-of-range as domain `0`
- Aligns with `agnocast::get_ros_domain_id()` in agnocastlib (inlined because this package does not depend on agnocastlib)
- Unit tests for valid, unset, empty, non-numeric, and trailing-junk values

## Test plan

- [x] Standalone parser smoke for `7` / empty / `abc` / unset → `7` / `0` / `0` / `0`
- [ ] `colcon test --packages-select agnocast_cie_thread_configurator` (ROS environment; CI)

## Related

Fixes #1401


VedantMadane:
Note: this is a bugfix only (no API/compat break) — please apply the required `need-patch-update` label when convenient (I don't have permission to add labels).

src/agnocast_cie_thread_configurator/src/util.cpp
```diff
@@ -83,11 +86,21 @@ std::map<std::string, std::string> get_hardware_info()
...
   const char * env_value = std::getenv("ROS_DOMAIN_ID");
-  if (env_value != nullptr) {
-    return static_cast<size_t>(std::stoul(env_value));
+  if (env_value == nullptr || *env_value == '\0') {
+    return 0;
   }
-  return 0;  // default domain ID
+  char * end = nullptr;
+  errno = 0;
+  const unsigned long value = std::strtoul(env_value, &end, 10);
```
atsushi421: The value returned here keys the `domain_id == default_domain_id` dedup in thread_configurator_node.cpp and prerun_node.cpp, but the node's actual domain is decided by rcl, which parses ROS_DOMAIN_ID with `strtoul(env, &end, 0)` (base 0) and rejects trailing junk only when the parsed number is 0 (Humble, rcl/src/rcl/domain_id.c). With `ROS_DOMAIN_ID=12x`, rclcpp::init succeeds and the node runs in domain 12 while this returns 0 (the old `std::stoul` also returned 12, so this case is a regression): a duplicate node is then created inside the node's own domain and entries for the phantom default domain never get a subscription. Since this package already depends on rclcpp (hence rcl), consider replacing the hand-rolled parser with a thin wrapper around `rcl_get_default_domain_id()` from `rcl/domain_id.h` (fall back to 0 and `rcl_reset_error()` on non-OK), so `default_domain_id_` always matches the domain rcl actually assigns.

src/agnocast_cie_thread_configurator/src/util.cpp
```diff
@@ -83,11 +86,21 @@ std::map<std::string, std::string> get_hardware_info()
...
+  if (end == env_value || *end != '\0' || errno != 0 ||
+      value > static_cast<unsigned long>(std::numeric_limits<uint32_t>::max())) {
+    return 0;
```
atsushi421, line 37: When ROS_DOMAIN_ID is set but invalid, this silently falls back to domain 0, while prerun_node.cpp emits RCLCPP_WARN for invalid explicitly-configured domain IDs. Since this function is only called after rclcpp::init, consider logging a warning with the raw value before returning 0 so a domain mismatch is diagnosable.

src/agnocast_cie_thread_configurator/test/test_get_default_domain_id.cpp
```diff
@@ -0,0 +1,79 @@
...
+}
+
+TEST(GetDefaultDomainId, NonNumericFallsBackToZero)
+{
+  EnvVarGuard guard("ROS_DOMAIN_ID");
+  guard.set("abc");
+  EXPECT_EQ(acie::get_default_domain_id(), 0u);
+}
+
+TEST(GetDefaultDomainId, TrailingJunkFallsBackToZero)
```
atsushi421: The new rejection branches are untested: nothing covers a value above UINT32_MAX (e.g. "4294967296"), ERANGE overflow (e.g. "99999999999999999999"), or negative input ("-1", which strtoul wraps to ULONG_MAX and only the range check catches), and the accepted boundary "4294967295" is also unverified. Consider adding these four cases, matching the coverage the Python parser already has in ros2agnocast_discovery_agent/test/test_agent.py.

atsushi421 (commented):
The claim that non-numeric values no longer abort only holds at the unit level: in the real executables, main.cpp calls rclcpp::init outside the try/catch, and rcl rejects a non-numeric ROS_DOMAIN_ID (e.g. `abc`) during init, so the process still terminates. Only the empty-string case is fixed end to end. Consider moving rclcpp::init inside the try/catch, or scoping the PR description to the empty-value crash plus library-level hardening.

---

The declaration comment in cie_thread_configurator.hpp still reads "Get default domain ID from ROS_DOMAIN_ID environment variable" and does not mention the new contract (never throws, falls back to 0 on unset/empty/unparsable/out-of-range values), which is what callers actually read. `agnocast::get_ros_domain_id()` documents exactly this at its declaration in agnocast_utils.hpp; consider mirroring that wording.

VedantMadane:
Thanks @atsushi421! I have updated the declaration comment in \cie_thread_configurator.hpp\ to explicitly document the fallback contract (never throws, falls back to 0 on unset/empty/unparsable/out-of-range values, matching \gnocast::get_ros_domain_id()\) and rebased the branch onto current \main\.

atsushi421:
Thanks @VedantMadane, the header comment looks good.

Three things remain:

1. The branch reverts main: the tree is main at `b67d6c07` plus your files, so the diff undoes #1587, #1588, #1589, #1602 and the 2.4.0 release. Please rebuild it on current main.
2. The three inline threads are not addressed yet.
3. `rclcpp::init` is still outside the try/catch in `main.cpp` and `prerun_node_main.cpp`, so non-numeric values still abort. Move it inside, or narrow the description to the empty-value case.

VedantMadane:
Thanks @atsushi421! I have addressed all feedback:

1. Rebuilt the branch cleanly directly on current \main\ (preserving release 2.4.0 and all recent commits #1587, #1588, #1589, #1602).
2. Replaced the hand-rolled parser with a wrapper around \
cl_get_default_domain_id()\ from \
cl/domain_id.h\ with \
cl_reset_error()\.
3. Added \RCLCPP_WARN\ logging when \ROS_DOMAIN_ID\ is set to an invalid value before falling back to 0.
4. Added boundary unit tests in \	est_get_default_domain_id.cpp\ covering valid values, unset, empty, non-numeric, boundary (\4294967295\), above uint32 max (\4294967296\), ERANGE overflow (\99999999999999999999\), and negative input (\-1\).
5. Enclosed \
clcpp::init\ inside the \	ry/catch\ block in both \main.cpp\ and \prerun_node_main.cpp\ with safe shutdown handling.

atsushi421:
@VedantMadane 
Thank you for addressing my comments, but some CI tests failed:

- https://github.com/autowarefoundation/agnocast/actions/runs/34116394895/job/101726706286?pr=1572
- https://github.com/autowarefoundation/agnocast/actions/runs/34117380613/job/101729261716?pr=1572
- https://github.com/autowarefoundation/agnocast/actions/runs/34117380613/job/101729261861?pr=1572

Could you fix these?

atsushi421:
Thanks @VedantMadane! The warning log, the boundary tests, the header contract, the rebase, and moving `rclcpp::init` inside the try/catch all look good. I verified locally that `ROS_DOMAIN_ID=abc` now exits with 1 instead of aborting, and `colcon test --packages-select agnocast_cie_thread_configurator` passes.

One item is still open. Your note says the parser was replaced with a wrapper around `rcl_get_default_domain_id()`, but the current diff still has the hand-rolled `strtoull(p, &end, 10)`, so `default_domain_id_` can still disagree with the domain rcl actually assigns. rcl parses with `strtoul(env, &end, 0)` and rejects trailing junk only when the parsed number is 0, which produces the differences below.

| ROS_DOMAIN_ID | rcl domain | this PR |
| --- | --- | --- |
| `12x` | 12 | 0 (warn) |
| `12 ` (trailing space) | 12 | 0 (warn) |
| `+5` | 5 | 0 (warn) |
| `0x10` | 16 | 0 (warn) |
| `010` | 8 (octal) | 10 |

The first three are regressions against the previous `std::stoul` code, which returned 12, 12 and 5. I checked `12x` end to end, and the node comes up in domain 12 while the configurator logs `falling back to default domain ID 0`, so a duplicate node is created inside the node's own domain and entries for the phantom default domain never get a subscription.

Matching `agnocast::get_ros_domain_id()` is not enough here, because this value keys the dedup against the domain rcl chose. Delegating to `rcl_get_default_domain_id()` (with `rcl_reset_error()` and the warning before returning 0 on non-OK) makes all rows agree by construction, and `abc`, `-1` and out-of-range values never reach this function anyway since `rclcpp::init` fails first.

Two nits:
- `util.cpp` uses `uint32_t` without including `<cstdint>`.
- The leading-whitespace loop is redundant, since `strtoull` skips whitespace itself.

PR #1573, opened 2026-08-26
feat(agnocastlib)[need-minor-update]: report the agnocast nodes in get_node_names()

Koichi98:
## Description

Adds `AGNOCAST_GET_NODE_NAMES_CMD`, which returns the nodes owning at least one non-bridge endpoint in the caller's IPC namespace and `ROS_DOMAIN_ID`, and implements `NodeGraph::get_node_names()` and `agnocast::Node::get_node_names()` on top of it. DDS is not consulted, so an `rclcpp::Node` holding no agnocast endpoint is not reported.

In Autoware the only caller is `duplicated_node_checker`, which detects a node launched twice and whose approach is itself under review. Enumerating `rclcpp::Node`s too would let it swap over unchanged, but that needs a far larger change than one checker justifies, so this reports only what agnocast already knows.

Node names live only on endpoints, so the kmod has to visit every endpoint and deduplicate on `(pid, name)`; deduplicating by linear scan would cost endpoints x nodes string compares under `global_htables_rwsem`, which is too much at a few hundred nodes and over a thousand endpoints, so an open-addressing set is used instead.

Names are returned in fixed `NODE_NAME_BUFFER_SIZE` slots and `node_name_buffer_size` counts names, matching the existing `*_topics` ioctls.

## Related links

None.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Post-merge checklist


agnocast_kmod/agnocast_ioctl.c
```diff
@@ -3090,6 +3212,37 @@ static long get_topic_list_cmd(union ioctl_topic_list_args __user * arg)
...
+
+  union ioctl_get_node_names_args get_node_names_args;
+  if (copy_from_user(&get_node_names_args, arg, sizeof(get_node_names_args))) return -EFAULT;
+
+  const uint32_t buf_node_num =
+    min_t(uint32_t, get_node_names_args.node_name_buffer_size, MAX_NODE_NUM);
+  char __user * user_buf =
+    (char __user *)u64_to_user_ptr(get_node_names_args.node_name_buffer_addr);
+
+  char * buf = kvmalloc((size_t)buf_node_num * NODE_NAME_BUFFER_SIZE, GFP_KERNEL);
```
atsushi421: `buf` is allocated with `kvmalloc`, but `add_unique_node` writes only `strlen(name)+1` bytes into each 256-byte slot while the handler later copies `node_num * NODE_NAME_BUFFER_SIZE` bytes to userspace, so up to 255 bytes of uninitialized kernel heap leak per reported name. Use `kvzalloc`, or copy each name with `strlen+1` granularity as `agnocast_ioctl_get_topic_list` does.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -3090,6 +3212,37 @@ static long get_topic_list_cmd(union ioctl_topic_list_args __user * arg)
...
+  if (!buf) return -ENOMEM;
+
+  uint32_t node_num = 0;
+  long ret =
+    agnocast_ioctl_get_node_names(ipc_ns, get_current_domain_id(), buf, buf_node_num, &node_num);
```
atsushi421: `get_current_domain_id()` is evaluated before `agnocast_ioctl_get_node_names()` takes `global_htables_rwsem`, but `get_process_domain_id()` documents that the caller must hold that rwsem: the unlocked `proc_info_htable` walk can race the exit path's `hash_del_rcu` + `kfree_rcu` (a use-after-free window on preemptible kernels). Resolve the domain inside a locked region, e.g. wrap this call in its own `down_read`/`up_read` pair or move it after the `down_read` inside `agnocast_ioctl_get_node_names()`.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -1676,6 +1676,128 @@ int agnocast_ioctl_get_topic_list(
...
+{
+  int ret = 0;
+
+  down_read(&wrapper->topic->rwsem);
+
+  struct publisher_info * pub_info;
+  int bkt_pub_info;
+  hash_for_each(wrapper->topic->pub_info_htable, bkt_pub_info, pub_info, node)
+  {
+    if (pub_info->domain_id != domain_id || pub_info->is_bridge) continue;
```
atsushi421, line 1752: This filter does not exclude endpoints of exited processes: a `publisher_info` survives process exit while its entries are still referenced (`pre_handler_publisher_exit`), so after a node crashes and restarts its name is reported twice under two pids, and `duplicated_node_checker` would raise a false duplicate alarm. Skip endpoints whose `agnocast_find_process_info(pid)` is NULL or has `exited == true` (the required rwsem is already held), and consider a kunit case covering this.

src/agnocastlib/src/node/node_interfaces/node_graph.cpp
```diff
@@ -62,11 +69,57 @@ std::map<std::string, std::vector<std::string>> NodeGraph::get_subscriber_names_
...
+// likewise counts two same-named nodes in different processes as two.
 std::vector<std::string> NodeGraph::get_node_names() const
 {
-  throw std::runtime_error("NodeGraph::get_node_names is not supported in agnocast.");
+  std::vector<std::string> node_names = query_agnocast_node_names();
+
+  // The kmod only knows nodes that own an endpoint, but rclcpp lists the calling node even
+  // when it owns none. Skipping a name already listed hides a same-named node in another process.
+  const std::string self_name = node_base_->get_fully_qualified_name();
+  if (std::find(node_names.begin(), node_names.end(), self_name) == node_names.end()) {
```
atsushi421, line 120: The kmod dedups on `(pid, name)` but this self-append check is name-only: when the calling node owns no endpoint and another process runs a same-named node, the name appears once where rclcpp would report it twice, which is exactly the duplicate `duplicated_node_checker` needs to see, and it contradicts the doc comment's "repeated once per node". Consider having the ioctl report whether the calling pid contributed an entry (or return pids alongside names) so self is appended based on `(pid, name)`; failing that, document this undercount explicitly.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -1676,6 +1676,128 @@ int agnocast_ioctl_get_topic_list(
   return ret;
 }
 
+// Nothing may sleep under global_htables_rwsem: it is writer-preferring, so a sleeping reader
+// stalls every publish behind a waiting endpoint add/remove. Hence the scratch preallocated
+// below, and the copy_to_user left to get_node_names_cmd.
+
+// Kept well above MAX_NODE_NUM so that probing always terminates on a free slot.
+#define NODE_NAME_SLOT_BITS 11
```
atsushi421, line 1683: The probe-termination invariant (`buf_node_num <= MAX_NODE_NUM < NODE_NAME_SLOT_NUM`) is enforced only by comments, and `agnocast_ioctl_get_node_names` is exported in agnocast.h, so a future caller passing a larger `buf_node_num` (or a bumped `MAX_NODE_NUM` without a matching `NODE_NAME_SLOT_BITS`) turns the probe loop in `add_unique_node` into an infinite loop while holding `global_htables_rwsem`. Add a compile-time check here (e.g. `BUILD_BUG_ON`-style assert that `NODE_NAME_SLOT_NUM >= 2 * MAX_NODE_NUM`) and reject `buf_node_num > MAX_NODE_NUM` in `agnocast_ioctl_get_node_names`.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -1676,6 +1676,128 @@ int agnocast_ioctl_get_topic_list(
...
+  while (col->slots[idx].pid != 0) {
+    const struct collected_node * slot = &col->slots[idx];
+    if (slot->pid == pid && strcmp(node_name_at(col, slot->name_index), name) == 0) {
+      return 0;  // already collected
+    }
+    idx = (idx + 1) & (NODE_NAME_SLOT_NUM - 1);
+  }
+
+  if (col->num >= col->capacity) {
+    dev_warn(
```
atsushi421: This warning prints the already-clamped capacity as `node_name_buffer_size`, omits `MAX_NODE_NUM` (unlike the `get_topic_list` counterpart), and fires while both `global_htables_rwsem` and the topic rwsem are held. Mirror the topic-list message format (print both the limit and the caller-supplied size) and consider emitting it after the locks are released, e.g. from `get_node_names_cmd` on `-ENOBUFS`.

src/agnocastlib/src/node/node_interfaces/node_graph.cpp
```diff
@@ -62,11 +69,57 @@ std::map<std::string, std::vector<std::string>> NodeGraph::get_subscriber_names_
...
+    node_names.push_back(self_name);
```
atsushi421: `self_name` is `const`, so this `push_back` always copies even though the variable is dead afterwards, and fully qualified names can exceed SSO. Drop the `const` and use `node_names.push_back(std::move(self_name))`.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -1676,6 +1676,128 @@ int agnocast_ioctl_get_topic_list(
...
+  uint32_t num;
+};
+
+static char * node_name_at(const struct node_name_collector * col, const uint32_t index)
+{
+  return col->buf + (size_t)index * NODE_NAME_BUFFER_SIZE;
+}
+
+// The dedup key is (pid, name), not the name alone: rclcpp reports same-named nodes in different
+// processes as two. Two in one process still collapse -- ROS 2 cannot tell those apart either.
```
atsushi421: The claim that ROS 2 also cannot tell two same-named nodes in one process apart looks incorrect: with DDS-based rmw each rcl node is a separate graph entity, so `rclcpp::Node::get_node_names()` reports the name once per node even within a single process. Suggest rewording this comment (and the matching kunit comment and comparison-doc text) as a limitation of the kmod's `(pid, name)` model rather than rclcpp parity.

src/agnocastlib/src/node/node_interfaces/node_graph.cpp
```diff
@@ -1,8 +1,15 @@
...
+#include "agnocast/agnocast_ioctl.hpp"
 #include "agnocast/agnocast_publisher.hpp"
 #include "agnocast/agnocast_subscription.hpp"
+#include "agnocast/agnocast_utils.hpp"
 
+#include <sys/ioctl.h>
+#include <unistd.h>
+
+#include <algorithm>
+#include <cerrno>
```
atsushi421, line 12: The new code calls `strerror()` and `exit()`/`EXIT_FAILURE` but relies on transitive includes for them. Add `#include <cstring>` and `#include <cstdlib>` next to `#include <cerrno>`.

src/agnocastlib/src/node/node_interfaces/node_graph.cpp
Koichi98: The self-append only runs for a caller owning no endpoint, which default `NodeOptions` rules out by starting parameter services, so no node `duplicated_node_checker` watches takes this path; documented the undercount instead.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in 3c1b47be.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in 722418d5.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in 83f2f1c0.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in a98f5503.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in d53bb135.

src/agnocastlib/src/node/node_interfaces/node_graph.cpp
Koichi98: Fixed in 84bb9ce1.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in d56d9e09.

src/agnocastlib/src/node/node_interfaces/node_graph.cpp
Koichi98: Fixed in 365b0482.

atsushi421:
@Koichi98 
Could you address this?
https://github.com/autowarefoundation/agnocast/actions/runs/33020873070/job/98350652028?pr=1573

The cause: `test_case_get_node_names_excludes_an_exited_process` fails at the Arrange step, where `agnocast_ioctl_publish_msg` returns `-EINVAL` ("Topic not found"). `agnocast_ioctl_publish_msg` looks up the topic via `find_topic_for_current()`, which uses `get_current_domain_id()` of the process actually issuing the call (the kunit thread itself). The test registers only the fake `PID`/`PID2` in `DOMAIN_ID` and never registers `current->tgid`, so the lookup cannot find the topic created under `DOMAIN_ID`.

Koichi98:
Thanks for the diagnosis, that was exactly it. Fixed in 13f9827a by registering the publisher under `current->tgid` instead of a fabricated pid, which is the relationship a real publisher has anyway.

atsushi421 (approved):

sykwer (approved):

Issue #1570, opened 2026-08-26
The cross-domain A2R bridge is forced instead of demand-driven


Koichi98:
The A2R bridge forced from the domain bridge rule config (#1558) is unconditional because gossip is confined to the agent's own DDS domain, leaving an agent no evidence of a subscriber in `to_domain`. Two ways to supply that evidence: (A) give the agent one `rclpy` context per domain named in the config, which costs an executor thread per domain and needs the kmod's per-`(ns, domain)` singleton claim reworked; or (B) move gossip to a single well-known discovery domain that every agent joins, which also collapses the agent back to one per namespace, since the per-domain split exists only so gossip can ride the application's own domain. (B) is the smaller change but puts a participant on a shared domain, which conflicts with using domains for isolation.

PR #1574, opened 2026-08-26
Fix/daemon force cross domain bridge

kobayu858:
## Description

Taken over from #1558, which this replaces. The first two commits are that PR and the handover; the rest answer review.

A topic split across **both** an IPC namespace and a ROS domain never flows, silently: `domain_bridge` waits for a DDS publisher in `from_domain` before subscribing, while the A2R bridge waits for a DDS subscriber before publishing. Neither moves first. The agent now reads the domain bridge YAML and forces the A2R bridge on every topic whose own domain is a rule's `from` side, giving `domain_bridge` a publisher to attach to; the `to` side needs no force. `bidirectional: true` yields both directions. Config path: `AGNOCAST_DOMAIN_BRIDGE_CONFIG`, else `/etc/agnocast/domain_bridge.yaml` -- the default matters because agnocastlib `execv`s the agent from an application process. With no config, behavior is unchanged.

### What the handover changes

**A rule that forces nothing said so nowhere.** `decide_domain_rule_bridges` skipped on a missing local topic, an unresolved type, or a bridge-only publisher without a word. An empty type usually means the tmpfs registry join failed rather than that the topic is new, so forcing could vanish and leave exactly the symptom it exists to remove.

It now reports each reason. A rule whose `from` side is another agent's domain is skipped silently first -- `bidirectional` splits one entry into a tuple per direction, so one of each pair is always someone else's. What remains is the ordinary startup gap, since the agent is forked by the first Agnocast process and ticks before its publishers exist: the first sighting is info, and it warns once only after the cell has gone unforced for 30 ticks, with an info line when it resolves. The count is per cell per tick and survives a changed reason, so a fault that alternates between reasons escalates instead of sitting at info forever.

**`logger.warn('...: %s', err)` raised instead of logging.** rclpy's logger takes no format arguments, so the `TypeError` escaped `rclpy.spin` and ended the agent. The line predates #1558, but dropping the no-remote-peers early return makes it reachable in exactly the configuration this PR targets.

**A config the external node runs was rejected wholesale.** yaml-cpp resolves the whole YAML 1.1 boolean set; PyYAML leaves a bare `y`/`n` as a string, so `bidirectional: y` ran in `domain_bridge` and disabled every rule here. `_as_bool` accepts what yaml-cpp does, deliberately looser about case and surrounding space: a value this takes and yaml-cpp rejects fails loudly in the external node, while the reverse fails quietly on this side alone. The malformed-config log moves to `error` and says forcing is off for ALL topics, since that is what happens.

**Tests.** The suite passed with the rules replaced by an empty list, with the reverse leg of every bidirectional rule discarded, with the forced request carrying a wrong type name or zeroed QoS, with the dedup key broken, with the empty-request guard removed, with the domain filter deleted, with the logger or the shared counter dict never reaching the decider, and with the tick count doubling. Each of those now fails.

## Consequence worth stating

Breaking the wait activates a second delivery path. For a subscriber in the **same** IPC namespace as the publisher but a different domain, the kmod rule already delivers zero-copy; the forced A2R now also puts the topic on DDS, `domain_bridge` relays it, and the R2A bridge in `to_domain` reinjects it -- so that subscriber receives every message twice. The wait was suppressing this incidentally, so duplicate delivery goes from possible to systematic for any configured topic with a same-namespace cross-domain consumer. `ret_other_domain_subscriber_num` (#1542) is read only for service readiness (#1564); the pub/sub demand paths ignore it, so nothing suppresses this. Being handled separately, not here.

Also unconditional by design (#1570): a listed topic is serialized onto DDS whenever it has a local publisher, whether or not anything consumes it.

## Related links

Replaces #1558. #1570 could replace the force with a real demand check, but that needs gossip to cross domains. #1575 covers the forced-bridge lifecycle: a request that cannot be built is re-sent every tick and re-inserted, so dropping it in the manager changes nothing, and that is left alone here.

Left alone as pre-existing, each worth its own change:
- a relative topic name in the YAML matches nothing, so `register_domain_bridge` registers a kmod rule for a topic that cannot exist and the force lookup misses. `agnocast_ioctl_add_domain_bridge` does no name validation, unlike the prefix and service ioctls.
- `remap: ""` becomes a rule for `/`; `domain_bridge` treats it as no remap. `remap: 123` is rejected here and coerced by yaml-cpp.
- `create_daemon_forced_bridges` / `activate_daemon_forced_bridge` have no C++ test, and no e2e covers the cross-domain chain.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

`pytest` in `src/ros2agnocast_discovery_agent`: 116 passed, guards mutation-checked as described above. The required suites and a two-namespace/two-domain runtime check are outstanding.

## Notes for reviewers

`9a7f0135` is #1558 unchanged; everything after it is the handover and the review answers. The kmod and agnocastlib are untouched -- the change is confined to `ros2agnocast_discovery_agent`.

## Post-merge checklist

**Important notes:**

### DIFF
--- a/src/agnocastlib/src/bridge/agnocast_bridge_manager.cpp
+++ b/src/agnocastlib/src/bridge/agnocast_bridge_manager.cpp
@@ -317,19 +317,15 @@ void BridgeManager::create_daemon_forced_bridges()
       // local Agnocast endpoint is still required for the bridge to carry traffic.
       const auto count = is_r2a ? get_agnocast_subscriber_count(topic_name).count
                                 : get_agnocast_publisher_count(topic_name).count;
-      if (
-        !already_active && count > 0 &&
-        !activate_daemon_forced_bridge(
-          topic_name, it->second.message_type, it->second.qos, is_r2a)) {
-        it = forced.erase(it);
-        continue;
+      if (!already_active && count > 0) {
+        activate_daemon_forced_bridge(topic_name, it->second.message_type, it->second.qos, is_r2a);
       }
       ++it;
     }
   }
 }
 
-bool BridgeManager::activate_daemon_forced_bridge(
+void BridgeManager::activate_daemon_forced_bridge(
   const std::string & topic_name, const std::string & message_type, const rclcpp::QoS & qos,
   bool is_r2a)
 {
@@ -338,10 +334,7 @@ bool BridgeManager::activate_daemon_forced_bridge(
       is_r2a ? loader_->create_r2a_pubsub_bridge(container_node_, topic_name, message_type, qos)
              : loader_->create_a2r_pubsub_bridge(container_node_, topic_name, message_type, qos);
     if (!result.entity_handle) {
-      RCLCPP_WARN(
-        logger_, "Daemon-forced bridge for '%s' produced no entity; dropping the request.",
-        topic_name.c_str());
-      return false;
+      return;
     }
 
     if (is_r2a) {
@@ -365,19 +358,13 @@ bool BridgeManager::activate_daemon_forced_bridge(
       }
       active_pubsub_a2r_bridges_[topic_name] = result;
     }
-    return true;
   } catch (const std::exception & e) {
-    // Mirror the on-demand path: a request that cannot be built is dropped rather than retried,
-    // or the daemon's per-tick renewal would repeat the same failure forever.
     RCLCPP_WARN(
-      logger_, "Failed to create daemon-forced bridge for '%s': %s. Dropping the request.",
-      topic_name.c_str(), e.what());
+      logger_, "Failed to create daemon-forced bridge for '%s': %s", topic_name.c_str(), e.what());
   } catch (...) {
     RCLCPP_WARN(
-      logger_, "Unknown error creating daemon-forced bridge for '%s'. Dropping the request.",
-      topic_name.c_str());
+      logger_, "Unknown error creating daemon-forced bridge for '%s'", topic_name.c_str());
   }
-  return false;
 }
 
 void BridgeManager::request_shutdown()
--- a/src/agnocastlib/src/bridge/agnocast_bridge_manager.cpp
+++ b/src/agnocastlib/src/bridge/agnocast_bridge_manager.cpp
@@ -317,15 +317,19 @@ void BridgeManager::create_daemon_forced_bridges()
       // local Agnocast endpoint is still required for the bridge to carry traffic.
       const auto count = is_r2a ? get_agnocast_subscriber_count(topic_name).count
                                 : get_agnocast_publisher_count(topic_name).count;
-      if (!already_active && count > 0) {
-        activate_daemon_forced_bridge(topic_name, it->second.message_type, it->second.qos, is_r2a);
+      if (
+        !already_active && count > 0 &&
+        !activate_daemon_forced_bridge(
+          topic_name, it->second.message_type, it->second.qos, is_r2a)) {
+        it = forced.erase(it);
+        continue;
       }
       ++it;
     }
   }
 }
 
-void BridgeManager::activate_daemon_forced_bridge(
+bool BridgeManager::activate_daemon_forced_bridge(
   const std::string & topic_name, const std::string & message_type, const rclcpp::QoS & qos,
   bool is_r2a)
 {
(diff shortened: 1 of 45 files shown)


src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/bridge_decider.py
```diff
@@ -158,6 +160,79 @@ def decide_bridges(local_state, remote_states) -> list:
...
+    local_by_topic = {(t.topic_name, t.domain_id): t for t in local_state.topics}
+
+    for from_topic, _to_topic, from_domain, to_domain in rules:
+        if from_domain == to_domain:
+            continue
+
+        cell = (from_topic, from_domain)
+        local_topic = local_by_topic.get(cell)
+        if local_topic is None:
+            _note_unforced(logger, reported, cell, 'no local topic in this domain')
```
Koichi98: `local_topic is None` covers two cases: the rule names a domain this agent does not own, and the rule is this agent's with nothing to force yet.

Both warn now, so with `bidirectional` splitting an entry into two tuples, every rule logs one spurious `no local topic in this domain` in every agent. Could the foreign-domain rules be skipped before the warn?

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/bridge_decider.py
```diff
@@ -201,7 +276,8 @@ def dispatch_requests(
     ``send_request`` swallows ECONNREFUSED/ENOENT so a missing peer never
     stalls the daemon, and the request is re-issued idempotently next tick.
     """
-    for req in requests:
+    # decide_bridges and decide_domain_rule_bridges can both ask for the same bridge.
+    for req in dict.fromkeys(requests):
         err = send_request(
             _bridge_uds_addr(ipc_ns_inode, req.domain_id), serialize_request(req))
         if err is not None and logger is not None:
```
Koichi98, line 321: `logger.warn('daemon bridge dispatch failed: %s', err)` on the next line throws instead of logging: rclpy's logger takes no format arguments, and the `TypeError` escapes `rclpy.spin`, ending the agent.

The line predates this PR, but dropping `if not self._remote_states: return` makes it reachable whenever domain rules are configured with no gossip peer — the configuration this PR targets. Worth fixing here; it is one line.

src/agnocastlib/src/bridge/agnocast_bridge_manager.cpp
```diff
@@ -317,15 +317,19 @@ void BridgeManager::create_daemon_forced_bridges()
       // local Agnocast endpoint is still required for the bridge to carry traffic.
       const auto count = is_r2a ? get_agnocast_subscriber_count(topic_name).count
                                 : get_agnocast_publisher_count(topic_name).count;
-      if (!already_active && count > 0) {
-        activate_daemon_forced_bridge(topic_name, it->second.message_type, it->second.qos, is_r2a);
+      if (
+        !already_active && count > 0 &&
+        !activate_daemon_forced_bridge(
+          topic_name, it->second.message_type, it->second.qos, is_r2a)) {
+        it = forced.erase(it);
```
Koichi98: `it = forced.erase(it)` does not stop the retry: the agent re-sends every request each tick regardless of outcome and `register_daemon_pubsub_request` re-inserts it, so with the worker loop also at ~1 Hz the same failure is re-attempted at the same rate as before, now with a WARN each second. It only shortens the case where the agent has already stopped sending, which `DAEMON_FORCE_TTL` covers in 5s.

The lifecycle question behind this is filed as #1575.

Separately, `no local topic in this domain` and `the topic type is not known yet` are the normal state at startup: the agent is forked by the first Agnocast process, so it ticks before the publishers exist, and nothing is logged when they resolve. Could those be `info` until they persist?

kobayu858: You're right on both. The daemon re-sends every tick and `insert_or_assign` puts it straight back, so the erase bought nothing but a WARN per second — reverted the whole C++ change and left the lifecycle to #1575. And that gap is the ordinary startup state, so the first sighting is now info and it only warns once the cell has gone unforced for 30 ticks, with an info line when it resolves. https://github.com/autowarefoundation/agnocast/pull/1574/commits/7f73918cb3d15e098e944d7b17714add68f929c6

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/bridge_decider.py
kobayu858: Thanks, you're right that this PR is what makes it reachable. Fixed in https://github.com/autowarefoundation/agnocast/pull/1574/commits/67141c71d55d0feb6bd488271978c95af4ed8890.

src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/bridge_decider.py
kobayu858: Good catch — bidirectional makes it one spurious line per rule in every agent. Fixed: the agent now passes its own domain and the decider drops a rule whose `from` side is not it, before the warn. https://github.com/autowarefoundation/agnocast/pull/1574/commits/97f5dc0a0d12223715dbeec6b66cc801a629d0ac

Koichi98 (approved):

PR #1580, opened 2026-08-27
feat(agnocastlib): support service introspection on the client side

Koichi98:
## Description

Publishes `REQUEST_SENT` and `RESPONSE_RECEIVED` from `agnocast::Client<T>`, so `<service>/_service_event` carries both sides of a completed call rather than only the server's half. Introspection stays off until `configure_introspection(clock, qos, state)` is called, the same entry point the server side already has.

`REQUEST_SENT` is published after `publish()` has handed the request over, and `RESPONSE_RECEIVED` once the response has been taken, so each event reports something that has already happened. This matches `rcl_send_request` and `rcl_take_response_with_info`.

Follow-up to #1551, which covered the server side.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

The integration test gains two cases: the client's two events with their contents, and one that enables both sides and checks that the four events of a single call share `(client_gid, sequence_number)`. Run on a Jazzy build with the kernel module loaded: 12 cases pass.

Also checked by hand in a Jazzy container against the host kernel module: with both sides enabled, `ros2 service echo` receives all four event types.

## Notes for reviewers

The integration test is labelled `requires_kernel_module`, so the standard build matrix skips it.

## Post-merge checklist

### DIFF
--- a/src/agnocastlib/include/agnocast/agnocast_client.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_client.hpp
@@ -193,6 +193,7 @@ private:
   std::unordered_map<int64_t, ResponseCallInfo> seqno2_response_call_info_;
   typename ServiceRequestPublisher::SharedPtr publisher_;
 #if AGNOCAST_HAS_SERVICE_INTROSPECTION
+  // Declared before subscriber_ so that it outlives the callback that uses it.
   std::unique_ptr<ServiceEventPublisher> event_publisher_;
 #endif
   typename ServiceResponseSubscriber::SharedPtr subscriber_;
--- a/src/agnocastlib/include/agnocast/agnocast_client.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast_client.hpp
@@ -26,6 +26,7 @@
 #include <future>
 #include <memory>
 #include <mutex>
+#include <optional>
 #include <string>
 #include <type_traits>
 #include <unordered_map>
@@ -198,6 +199,27 @@ private:
 #endif
   typename ServiceResponseSubscriber::SharedPtr subscriber_;
 
+#if AGNOCAST_HAS_SERVICE_INTROSPECTION
+  // Must be called before publish(). Only CONTENTS puts the payload in the event, so the other
+  // states pay nothing; raising to CONTENTS in between costs that one event its payload.
+  std::optional<typename ServiceT::Request> copy_request_if_contents(
+    const ipc_shared_ptr<RequestT> & request)
+  {
+    if (event_publisher_->introspection_state() != RCL_SERVICE_INTROSPECTION_CONTENTS) {
+      return std::nullopt;
+    }
+    return static_cast<const typename ServiceT::Request &>(*request);
+  }
+
+  void publish_request_sent_event(
+    const int64_t seqno, const std::optional<typename ServiceT::Request> & request)
+  {
+    event_publisher_->publish_service_event_message(
+      service_msgs::msg::ServiceEventInfo::REQUEST_SENT, request ? &*request : nullptr, seqno,
+      get_gid().data);
+  }
+#endif
+
   template <typename Func>
   int64_t send_request_impl(
     ipc_shared_ptr<typename ServiceT::Request> && request, ResponseCallInfo && call_info,
@@ -213,14 +235,15 @@ private:
     }
 
 #if AGNOCAST_HAS_SERVICE_INTROSPECTION
-    // Must precede the publish, which leaves `internal_request` empty.
-    event_publisher_->publish_service_event_message(
-      service_msgs::msg::ServiceEventInfo::REQUEST_SENT,
-      static_cast<const typename ServiceT::Request *>(internal_request.get()), seqno,
-      get_gid().data);
+    const auto sent_request = copy_request_if_contents(internal_request);
 #endif
 
     publisher_->publish(std::move(internal_request));
+
+#if AGNOCAST_HAS_SERVICE_INTROSPECTION
+    publish_request_sent_event(seqno, sent_request);
+#endif
+
     return seqno;
   }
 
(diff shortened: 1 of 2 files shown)


src/agnocastlib/include/agnocast/agnocast_client.hpp
```diff
@@ -203,7 +234,16 @@ class Client : public ClientBase
...
     }
 
+#if AGNOCAST_HAS_SERVICE_INTROSPECTION
+    const auto sent_request = copy_request_if_contents(internal_request);
+#endif
+
     publisher_->publish(std::move(internal_request));
+
+#if AGNOCAST_HAS_SERVICE_INTROSPECTION
+    publish_request_sent_event(seqno, sent_request);
```
kobayu858, line 244: The Description says `REQUEST_SENT` goes out "before `publish()`", but only the payload copy does -- `publish()` has already woken the server by the time line 244 runs, so a consumer can see the response ahead of the request that caused it. rcl introspects after `rmw_send_request` too, so this is parity rather than a regression, but here the two client events come from different threads with no ordering between them. The copy is already taken, so moving 244 above 241 is free. Also, "the whole exchange" holds only for a configured `Client<T>`: a call that never gets a response still yields three events, since nothing prunes `seqno2_response_call_info_`.

src/agnocastlib/test/integration/test_agnocast_service_introspection.cpp
```diff
@@ -339,4 +344,54 @@ TEST_F(DeferredServiceIntrospectionTest, ADeferredResponsePublishesBothEvents)
...
+  set_client_introspection(RCL_SERVICE_INTROSPECTION_METADATA);
+
+  // Act
+  ASSERT_TRUE(call_service(true));
+  const auto events = wait_for_events(4);
+
+  // Assert: one call, so the four events share the correlation key a consumer pairs them by.
+  ASSERT_EQ(events.size(), 4u);
+  for (const auto & event : events) {
+    EXPECT_EQ(event.info.sequence_number, events[0].info.sequence_number);
```
kobayu858, line 396: This makes one call, and `next_sequence_number_{0}` with `fetch_add(1)` returning the old value means every event here carries sequence number 0, so the assertion reduces to `0 == 0`. Zeroing `info.sequence_number` unconditionally in `publish_service_event_message` leaves it green. The gid half has the same shape -- `BothEventsOfOneCallCarryTheSameCallerAndSequenceNumber` guards against an all-zero gid and this one does not, so zeroing both fields passes. Two calls, asserting on the second, would pin what the comment says it pins.

src/agnocastlib/test/integration/test_agnocast_service_introspection.cpp
```diff
@@ -339,4 +344,54 @@ TEST_F(DeferredServiceIntrospectionTest, ADeferredResponsePublishesBothEvents)
...
+  // Arrange
+  set_client_introspection(RCL_SERVICE_INTROSPECTION_CONTENTS);
+
+  // Act
+  ASSERT_TRUE(call_service(true));
+  const auto events = wait_for_events(2);
+
+  // Assert
+  ASSERT_EQ(events.size(), 2u);
+  EXPECT_EQ(events[0].info.event_type, ServiceEventInfo::REQUEST_SENT);
```
kobayu858: These assert on position, but `REQUEST_SENT` is published from the gtest main thread inside `call_service` and `RESPONSE_RECEIVED` from `spin_thread_`, with no happens-before between them (see the other thread). The window is narrow -- the whole server round trip has to fit between the `publish()` and the event publish -- but it is a real race, and `BothSidesTogetherCoverTheWholeExchange` already sorts the type multiset instead. Worth doing the same here, or asserting the set of types plus the per-type payload rather than indices.

src/agnocastlib/include/agnocast/agnocast_client.hpp
```diff
@@ -240,6 +286,14 @@ class Client : public ClientBase
       /* --- critical section end --- */
       lock.unlock();
 
+#if AGNOCAST_HAS_SERVICE_INTROSPECTION
+      // Must precede the move into the promise, which leaves `response` empty.
+      event_publisher_->publish_service_event_message(
```
kobayu858, line 292: The `bad entry id` branch above returns before this, so a response that arrives with an unknown seqno is taken but produces no `RESPONSE_RECEIVED`. rcl publishes the event whenever a response is taken, regardless. Since #1540 gave each client its own response topic this should be unreachable in practice, so I am asking about the intent rather than the behaviour -- if it is deliberate, a line saying the event follows the pairing rather than the arrival would help.

src/agnocastlib/include/agnocast/agnocast_client.hpp
Koichi98: Fixed the Description. I kept the publish order: rclcpp has the same split, with `REQUEST_SENT` published on the caller's thread and `RESPONSE_RECEIVED` on the executor's, so the two are unordered there too.

src/agnocastlib/test/integration/test_agnocast_service_introspection.cpp
Koichi98: Fixed in 24cc02c4. I also dropped the `client_gid` assertions from the sibling test from #1551, where both events read the same field so the equality compared a value with itself.

src/agnocastlib/test/integration/test_agnocast_service_introspection.cpp
Koichi98: Fixed in 8d55e620.

src/agnocastlib/include/agnocast/agnocast_client.hpp
Koichi98: Fixed in c1cb8659. It was not deliberate, so I kept the behaviour and stated it at the branch rather than publishing an event no consumer could match.

kobayu858 (approved):

PR #1581, opened 2026-08-27
feat(agnocastlib): add create_take_subscription() factories

Koichi98:
## Description

Add `create_take_subscription()` as a node method and as a Stage 1 free function, so a `TakeSubscription` is obtained from a factory like every other Agnocast entity instead of being constructed with `std::make_shared` by the caller. The deprecation messages on the callback-less `create_subscription()` overloads now point at it, and the sample application, the e2e take subscriber and the unit test go through it.

Returning one `Subscription` for both modes, as rclcpp does, was considered and dropped (#1522). Making callback delivery and `take()` actually work on one subscription needs a large design change — the kernel reference is one bit per (subscriber, entry) and the read cursor is one per subscriber — and no application uses both modes on the same subscription, so the split types stay.

Given the split, changing the callback-less `create_subscription()` to return a `TakeSubscription` was rejected: overloads cannot differ by return type alone, so old and new cannot coexist. It would break user code at once with no migration period, hence a major version bump. A new name keeps the change additive and lets the deprecated overloads live until `PollingSubscriber` moves to `autoware_agnocast_wrapper`. It is not rclcpp-aligned, but it states at the call site that the subscription is polled, which I consider the better trade.

## Related links

- #1522 — the rejected unification of `Subscription` and `TakeSubscription`
- #1506 — deprecated the `PollingSubscriber` entry points

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

Built `agnocastlib`, `agnocast_sample_application` and `agnocast_e2e_test`, and ran the `GetActualQosTest` unit tests (8 passed), which now cover the free function with `rclcpp::Node` and the node method with `agnocast::Node`.

## Notes for reviewers

`agnocast_sample_application`'s `no_rclcpp_take_subscriber` was the only remaining in-tree user of the deprecated `PollingSubscriber` API and emitted deprecation warnings on every build.

## Post-merge checklist

### DIFF
--- a/src/agnocastlib/include/agnocast/agnocast.hpp
+++ b/src/agnocastlib/include/agnocast/agnocast.hpp
@@ -191,6 +191,47 @@ typename Subscription<MessageT>::SharedPtr create_subscription(
     std::forward<Func>(callback), options);
 }
 
+/// @brief Create an Agnocast take-subscription (Stage 1 free function, QoS overload).
+/// @tparam MessageT ROS message type.
+/// @tparam NodeT Node type (rclcpp::Node or agnocast::Node).
+/// @param node Pointer to the node.
+/// @param topic_name Topic name.
+/// @param qos Quality of service profile.
+/// @param options Subscription options.
+/// @return Shared pointer to the created take-subscription.
+AGNOCAST_PUBLIC
+template <typename MessageT, typename NodeT>
+typename TakeSubscription<MessageT>::SharedPtr create_take_subscription(
+  NodeT * node, const std::string & topic_name, const rclcpp::QoS & qos,
+  agnocast::SubscriptionOptions options = agnocast::SubscriptionOptions{})
+{
+  static_assert(
+    std::is_base_of_v<rclcpp::Node, NodeT> || std::is_base_of_v<agnocast::Node, NodeT>,
+    "NodeT must be rclcpp::Node or agnocast::Node (or derived from them)");
+  return std::make_shared<TakeSubscription<MessageT>>(node, topic_name, qos, options);
+}
+
+/// @brief Create an Agnocast take-subscription (Stage 1 free function, history-depth overload).
+/// @tparam MessageT ROS message type.
+/// @tparam NodeT Node type (rclcpp::Node or agnocast::Node).
+/// @param node Pointer to the node.
+/// @param topic_name Topic name.
+/// @param qos_history_depth History depth for the QoS profile.
+/// @param options Subscription options.
+/// @return Shared pointer to the created take-subscription.
+AGNOCAST_PUBLIC
+template <typename MessageT, typename NodeT>
+typename TakeSubscription<MessageT>::SharedPtr create_take_subscription(
+  NodeT * node, const std::string & topic_name, const size_t qos_history_depth,
+  agnocast::SubscriptionOptions options = agnocast::SubscriptionOptions{})
+{
+  static_assert(
+    std::is_base_of_v<rclcpp::Node, NodeT> || std::is_base_of_v<agnocast::Node, NodeT>,
+    "NodeT must be rclcpp::Node or agnocast::Node (or derived from them)");
+  return std::make_shared<TakeSubscription<MessageT>>(
+    node, topic_name, rclcpp::QoS(rclcpp::KeepLast(qos_history_depth)), options);
+}
+
 /// @brief Create an Agnocast polling subscription (Stage 1 free function, history-depth overload).
 /// @tparam MessageT ROS message type.
 /// @tparam NodeT Node type (rclcpp::Node or agnocast::Node).
@@ -202,8 +243,8 @@ AGNOCAST_PUBLIC
 template <typename MessageT, typename NodeT>
 [[deprecated(
   "agnocast::PollingSubscriber is planned to move to autoware_agnocast_wrapper and be removed from "
-  "agnocast. Obtain a polling subscriber from the wrapper, or use agnocast::TakeSubscription "
-  "directly.")]]
+  "agnocast. Obtain a polling subscriber from the wrapper, or use "
+  "agnocast::create_take_subscription().")]]
 typename PollingSubscriber<MessageT>::SharedPtr create_subscription(
   NodeT * node, const std::string & topic_name, const size_t qos_history_depth)
 {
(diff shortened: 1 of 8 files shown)


src/agnocastlib/include/agnocast/agnocast_subscription.hpp
```diff
@@ -507,7 +507,7 @@ class PollingSubscriber
   [[deprecated(
     "agnocast::PollingSubscriber is planned to move to autoware_agnocast_wrapper and be removed "
     "from agnocast. Obtain a polling subscriber from the wrapper, or use "
-    "agnocast::TakeSubscription directly.")]]
+    "agnocast::create_take_subscription().")]]
```
atsushi421: The updated deprecation messages point users at `create_take_subscription()`, but `take_data()` is `take(true)` (latched: it keeps returning the last message every poll), while the replacement's natural call `take()` defaults to `allow_same_message=false` and returns an empty pointer whenever nothing new has arrived, and `take(true)` itself carries a removal warning. A mechanical migration therefore silently changes polling-loop behavior once the publisher stalls. Consider adding one sentence here (and to the four `create_subscription` messages) noting that `take()` returns each message at most once, so callers needing latched behavior must keep the last returned message themselves.

src/agnocastlib/include/agnocast/agnocast.hpp
```diff
@@ -191,6 +191,47 @@ typename Subscription<MessageT>::SharedPtr create_subscription(
...
+/// @tparam MessageT ROS message type.
+/// @tparam NodeT Node type (rclcpp::Node or agnocast::Node).
+/// @param node Pointer to the node.
+/// @param topic_name Topic name.
+/// @param qos_history_depth History depth for the QoS profile.
+/// @param options Subscription options.
+/// @return Shared pointer to the created take-subscription.
+AGNOCAST_PUBLIC
+template <typename MessageT, typename NodeT>
+typename TakeSubscription<MessageT>::SharedPtr create_take_subscription(
```
atsushi421, line 224: Of the four new entry points, this free-function history-depth overload is the only one never instantiated by any test or sample (and the free functions are never instantiated with `NodeT = agnocast::Node`), so being a template, a latent error in its body would not even fail the build. Consider adding a depth-assertion test through this overload, or delegating to the QoS overload as the node-method version does so one test covers both. Nothing currently verifies `options` forwarding either; a `qos_overriding_options` test like the existing `get_actual_qos_reflects_a_qos_parameter_override` would cover that.

src/agnocastlib/test/unit/test_agnocast_subscription.cpp
```diff
@@ -125,10 +125,25 @@ TEST_F(GetActualQosTest, get_actual_qos_reports_the_qos_of_a_take_subscription)
...
+  // Arrange
+  rclcpp::NodeOptions options;
+  options.start_parameter_services(false);
+  auto node = std::make_shared<agnocast::Node>("test_actual_qos_take_agnocast_node", options);
+
+  // Act
+  auto sub = node->create_take_subscription<StringMsg>("/test_actual_qos_take_agnocast_node", 4);
+
+  // Assert
+  EXPECT_EQ(sub->get_actual_qos().depth(), 4u);
```
atsushi421: The sibling `GetActualQosTest` cases compare the full QoS object, while this one asserts only `depth()`, so a regression that changed another policy would still pass. Consider `EXPECT_EQ(sub->get_actual_qos(), rclcpp::QoS(rclcpp::KeepLast(4)));`.

src/agnocastlib/include/agnocast/agnocast.hpp
```diff
@@ -191,6 +191,47 @@ typename Subscription<MessageT>::SharedPtr create_subscription(
...
+  NodeT * node, const std::string & topic_name, const rclcpp::QoS & qos,
+  agnocast::SubscriptionOptions options = agnocast::SubscriptionOptions{})
+{
+  static_assert(
+    std::is_base_of_v<rclcpp::Node, NodeT> || std::is_base_of_v<agnocast::Node, NodeT>,
+    "NodeT must be rclcpp::Node or agnocast::Node (or derived from them)");
+  return std::make_shared<TakeSubscription<MessageT>>(node, topic_name, qos, options);
```
atsushi421: `options` is taken by value and passed on as an lvalue to a constructor that also takes it by value, so each call copies `SubscriptionOptions` (a `shared_ptr` plus `QosOverridingOptions` with string/vector/`std::function`) twice, and three times on the node-method queue-size path. Passing `std::move(options)` at the forwarding sites (here, line 232, and the two node methods) would keep the signatures and drop the extra copies.

src/agnocastlib/include/agnocast/agnocast.hpp
```diff
@@ -191,6 +191,47 @@ typename Subscription<MessageT>::SharedPtr create_subscription(
     std::forward<Func>(callback), options);
 }
 
+/// @brief Create an Agnocast take-subscription (Stage 1 free function, QoS overload).
+/// @tparam MessageT ROS message type.
+/// @tparam NodeT Node type (rclcpp::Node or agnocast::Node).
+/// @param node Pointer to the node.
+/// @param topic_name Topic name.
+/// @param qos Quality of service profile.
+/// @param options Subscription options.
```
atsushi421, line 200: `TakeSubscription`'s construction path never reads `options.callback_group` and, unlike `Subscription`, does not validate it, so a caller passing a callback group through this factory gets a silent no-op. Consider documenting that `callback_group` has no effect for take-subscriptions, or warning on/rejecting a non-null one.

atsushi421 (commented):
`Subscription`'s class doc names its factory (Allocate instances with `agnocast::create_subscription<MessageT>()` or construct directly), but `TakeSubscription`'s class doc in `agnocast_subscription.hpp` gains no equivalent pointer to the new factory, so a reader landing on the class never learns `create_take_subscription()` exists. Consider mirroring that sentence.

---

`PollingSubscriber::takeData()` is still marked `[[deprecated("Use take_data() instead.")]]`, which now chains into a second deprecation because `take_data()` itself points at `create_take_subscription()`. Consider updating the `takeData()` message to point directly at the final destination.

src/agnocastlib/include/agnocast/agnocast_subscription.hpp
Koichi98: Fixed in 391abe35, in the `take_data()` and `takeData()` messages. I left the four `create_subscription()` messages unchanged: they return a `PollingSubscriber`, whose only way to get a message is `take_data()`, so the note is reached anyway.

src/agnocastlib/include/agnocast/agnocast.hpp
Koichi98: Fixed in 579d8a71.

src/agnocastlib/test/unit/test_agnocast_subscription.cpp
Koichi98: Fixed in e5e473e0.

src/agnocastlib/include/agnocast/agnocast.hpp
Koichi98: Fixed in 8377ea1a.

src/agnocastlib/include/agnocast/agnocast.hpp
Koichi98: Fixed in d8e47520, taking the warning option: the constructor now warns when `options.callback_group` is non-null.

atsushi421 (approved):

kobayu858 (approved):

PR #1582, opened 2026-08-27
fix(agnocast_kmod): track the unlink daemon instead of inferring it from the process count

Koichi98:
## Description

`ret_unlink_daemon_exist` was inferred from `get_process_num() > 0`, a count that includes exited processes still waiting to be drained. The unlink daemon is the only thing that drains them, so once it is gone the count never falls back to zero: no later process is told to spawn a replacement, and the entries and their shm segments are never cleaned up. The condition sustains itself.

Give the daemon an entry of its own. `process_info.is_bridge_manager` becomes `enum process_role` (`APPLICATION` / `BRIDGE_MANAGER` / `UNLINK_DAEMON`), the daemon registers through `AGNOCAST_ADD_PROCESS_CMD`, and `ret_unlink_daemon_exist` reports whether a live daemon is registered. A daemon that loses the race registers nothing and is told to stand down, decided under the write lock `add_process` already holds. The three counters that gate a daemon's shutdown skip it, since it is what they depend on.

Its entry is deleted rather than marked `exited`, in both directions. `agnocast_commit_exit_process()` deregisters it the moment it is told to exit, so a process starting before it actually dies is not told a daemon still exists — as `check_and_request_bridge_shutdown()` does for the bridge manager. `agnocast_process_exit_cleanup()` deletes it on death, since it is the one entry no daemon can drain.

`add_process()` also rejects a role outside the enum, and `AGNOCAST_DOMAIN_ID_NONE` from every role but the daemon, whose `domain_id` is assigned by role. `ROS_DOMAIN_ID=4294967295` therefore fails at startup rather than sharing that id.

Liveness is read only at `AGNOCAST_ADD_PROCESS_CMD`, so a daemon dying in an otherwise static namespace is not noticed until the next process starts.

## Related links

None.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

New KUnit cases cover registration and the refusal of a second daemon, respawn from a namespace of only exited entries (fails without this change), the daemon not counting as work, its entry and mempool slot released on both exit paths, and the rejected arguments.

## Notes for reviewers

The discovery agent stays in its own table: it is per (namespace, domain), and folding it in would mean reworking its three ioctls and its removal path.

The parent-side check in `initialize_agnocast()` is now an early-out hint rather than authoritative: processes starting at once can each fork a daemon, and all but the first exit as soon as `AGNOCAST_ADD_PROCESS_CMD` refuses them, before `assign_memory()`. Each loser stays a zombie of its parent until the parent exits, since `spawn_daemon_process()` single-forks and never reaps; the bridge manager and the discovery agent have had that shape all along. Whether to bring this back to a constant number of forks, and to double-fork so losers are reparented to init, will be decided by measurement in a follow-up.

## Post-merge checklist

`need-minor-update`

### DIFF
--- a/agnocast_kmod/agnocast_ioctl.c
+++ b/agnocast_kmod/agnocast_ioctl.c
@@ -586,9 +586,10 @@ static struct entry_node * find_message_entry(
 }
 
 // Forward declaration
-static int get_process_num(const struct ipc_namespace * ipc_ns);
-static int get_process_num_in_domain(const struct ipc_namespace * ipc_ns, const uint32_t domain_id);
-static int get_alive_process_num_in_domain(
+static int get_process_num_except_unlink_daemon(const struct ipc_namespace * ipc_ns);
+static int get_process_num_in_domain_except_unlink_daemon(
+  const struct ipc_namespace * ipc_ns, const uint32_t domain_id);
+static int get_alive_process_num_in_domain_except_unlink_daemon(
   const struct ipc_namespace * ipc_ns, const uint32_t domain_id);
 
 // Release subscriber reference from message entry (set boolean flag to false).
@@ -794,7 +795,23 @@ static bool has_alive_bridge_manager(const struct ipc_namespace * ipc_ns, const 
   {
     if (
       ipc_eq(ipc_ns, proc_info->ipc_ns) && proc_info->domain_id == domain_id &&
-      proc_info->is_bridge_manager && !proc_info->exited) {
+      proc_info->role == PROCESS_ROLE_BRIDGE_MANAGER && !proc_info->exited) {
+      return true;
+    }
+  }
+  return false;
+}
+
+// Namespace-scoped, unlike the bridge manager.
+static bool has_alive_unlink_daemon(const struct ipc_namespace * ipc_ns)
+{
+  struct process_info * proc_info;
+  int bkt;
+  hash_for_each(proc_info_htable, bkt, proc_info, node)
+  {
+    if (
+      ipc_eq(ipc_ns, proc_info->ipc_ns) && proc_info->role == PROCESS_ROLE_UNLINK_DAEMON &&
+      !proc_info->exited) {
       return true;
     }
   }
@@ -802,10 +819,22 @@ static bool has_alive_bridge_manager(const struct ipc_namespace * ipc_ns, const 
 }
 
 int agnocast_ioctl_add_process(
-  const pid_t pid, const struct ipc_namespace * ipc_ns, const bool is_bridge_manager,
+  const pid_t pid, const struct ipc_namespace * ipc_ns, const enum process_role role,
   const uint32_t domain_id, union ioctl_add_process_args * ioctl_ret)
 {
   int ret = 0;
+
+  if (
+    role != PROCESS_ROLE_APPLICATION && role != PROCESS_ROLE_BRIDGE_MANAGER &&
+    role != PROCESS_ROLE_UNLINK_DAEMON) {
+    return -EINVAL;
+  }
+
+  // Keeps the unlink daemon the only holder of AGNOCAST_DOMAIN_ID_NONE, which the domain-scoped
+  // counters rely on. The daemon's own domain_id is assigned below regardless of what it sends.
+  if (role != PROCESS_ROLE_UNLINK_DAEMON && domain_id == AGNOCAST_DOMAIN_ID_NONE) {
+    return -EINVAL;
+  }
 
   down_write(&global_htables_rwsem);
 
@@ -814,11 +843,16 @@ int agnocast_ioctl_add_process(
     ret = -EINVAL;
     goto unlock;
   }
-  ioctl_ret->ret_unlink_daemon_exist = (get_process_num(ipc_ns) > 0);
+  ioctl_ret->ret_unlink_daemon_exist = has_alive_unlink_daemon(ipc_ns);
   ioctl_ret->ret_bridge_daemon_exist = has_alive_bridge_manager(ipc_ns, domain_id);
   ioctl_ret->ret_discovery_agent_exist = (agnocast_find_discovery_agent(ipc_ns, domain_id) != NULL);
 
-  if (is_bridge_manager && ioctl_ret->ret_bridge_daemon_exist) {
+  // Deciding under the write lock add_process already holds is what stops two daemons starting
+  // at once from both registering.
+  if (role == PROCESS_ROLE_BRIDGE_MANAGER && ioctl_ret->ret_bridge_daemon_exist) {
+    goto unlock;
+  }
+  if (role == PROCESS_ROLE_UNLINK_DAEMON && ioctl_ret->ret_unlink_daemon_exist) {
     goto unlock;
   }
 
@@ -829,7 +863,7 @@ int agnocast_ioctl_add_process(
   }
 
   new_proc_info->exited = false;
-  new_proc_info->is_bridge_manager = is_bridge_manager;
+  new_proc_info->role = role;
   new_proc_info->global_pid = pid;
 #ifndef KUNIT_BUILD
   new_proc_info->local_pid = convert_pid_to_local(pid);
@@ -845,7 +879,8 @@ int agnocast_ioctl_add_process(
   }
 
   new_proc_info->ipc_ns = ipc_ns;
-  new_proc_info->domain_id = domain_id;
+  new_proc_info->domain_id =
+    (role == PROCESS_ROLE_UNLINK_DAEMON) ? AGNOCAST_DOMAIN_ID_NONE : domain_id;
 
   INIT_HLIST_NODE(&new_proc_info->node);
   uint32_t hash_val = hash_min(new_proc_info->global_pid, PROC_INFO_HASH_BITS);
@@ -1600,7 +1635,8 @@ pid_t agnocast_ioctl_get_exit_process(
 }
 
 void agnocast_commit_exit_process(
-  const struct ipc_namespace * ipc_ns, pid_t global_pid, bool * ret_daemon_should_exit)
+  const struct ipc_namespace * ipc_ns, pid_t global_pid, pid_t caller_pid,
+  bool * ret_daemon_should_exit)
 {
   down_write(&global_htables_rwsem);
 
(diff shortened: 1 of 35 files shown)


agnocast_kmod/agnocast_ioctl.c
```diff
@@ -1612,7 +1648,18 @@ void agnocast_commit_exit_process(
     }
   }
 
-  *ret_daemon_should_exit = (get_process_num(ipc_ns) == 0);
+  *ret_daemon_should_exit = (get_process_num_except_unlink_daemon(ipc_ns) == 0);
+
+  // Deregistering only on death would leave a window where a starting process is told a daemon
+  // exists and skips spawning its replacement.
+  if (*ret_daemon_should_exit && caller_pid >= 0) {
```
atsushi421: This deregisters the daemon on whichever commit first sees the namespace empty, which is usually the call that returns the last exited pid. But `poll_for_unlink()` only acts on `ret_daemon_should_exit` from the final call of its `do { } while (ret_pid > 0)` loop, so if a process registers in between (it is told no daemon exists and forks one), the next call re-derives `false` and the already-deregistered daemon keeps polling as an unregistered ghost alongside the replacement until the namespace empties again. Suggest deregistering only on the idle poll (`global_pid < 0`), the call whose flag userland actually honors, and adding a KUnit case for a pid-returning commit with `caller_pid == daemon_pid`; alternatively latch the flag in `poll_for_unlink()`.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -845,7 +879,8 @@ int agnocast_ioctl_add_process(
   }
 
   new_proc_info->ipc_ns = ipc_ns;
-  new_proc_info->domain_id = domain_id;
+  new_proc_info->domain_id =
```
atsushi421, line 888: The daemon goes through `assign_memory()` too, so it holds one of `mempool_num` slots (and a 16 GB VA range) it never maps, and when the pool is full its ADD_PROCESS fails with -ENOMEM and `poll_for_unlink()` exits, leaving the namespace without a daemon exactly where exited entries pile up fastest. Since the daemon never publishes, no path dereferences its `mempool_entry` (lines 730, 770 and 1098 are publisher-only), so skipping `assign_memory()` for `PROCESS_ROLE_UNLINK_DAEMON` (NULL entry, `ret_addr`/`ret_shm_size` = 0) is safe, and `free_memory()` is already a no-op for a pid without a slot. The `..._frees_its_mempool_slot` test would then become a "takes no slot" test.

atsushi421 (commented):
README.md's "Shared memory cleanup" section (line 193) still says that if the daemon dies, cleanup stops and resources leak, and recommends stopping processes individually. After this PR the next process to register spawns a replacement that drains the backlog, so the leak only persists in a namespace where no further process starts. Please update that paragraph accordingly, and consider whether agnocast_doc needs the same change (the template's post-merge checklist is missing from the PR body).

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -794,31 +795,64 @@ static bool has_alive_bridge_manager(const struct ipc_namespace * ipc_ns, const
...
-  const pid_t pid, const struct ipc_namespace * ipc_ns, const bool is_bridge_manager,
+  const pid_t pid, const struct ipc_namespace * ipc_ns, const enum process_role role,
   const uint32_t domain_id, union ioctl_add_process_args * ioctl_ret)
 {
   int ret = 0;
 
+  if (
+    role != PROCESS_ROLE_APPLICATION && role != PROCESS_ROLE_BRIDGE_MANAGER &&
+    role != PROCESS_ROLE_UNLINK_DAEMON) {
+    return -EINVAL;
```
atsushi421, line 833: Both new rejections return -EINVAL silently (unlike the "already exists" path just below), so a user with `ROS_DOMAIN_ID=4294967295`, which `get_ros_domain_id()` accepts, only sees "AGNOCAST_ADD_PROCESS_CMD failed: Invalid argument", and for the daemon child even that may go to /dev/null. Suggest a `dev_warn` with pid/role/domain_id here, plus mirroring `AGNOCAST_DOMAIN_ID_NONE` into agnocast_ioctl.hpp and rejecting it in `initialize_agnocast()` with a message that names `ROS_DOMAIN_ID`.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -794,31 +795,64 @@ static bool has_alive_bridge_manager(const struct ipc_namespace * ipc_ns, const
...
+
   down_write(&global_htables_rwsem);
 
   if (agnocast_find_process_info(pid)) {
     dev_warn(agnocast_device, "Process (pid=%d) already exists. (%s)\n", pid, __func__);
     ret = -EINVAL;
     goto unlock;
   }
-  ioctl_ret->ret_unlink_daemon_exist = (get_process_num(ipc_ns) > 0);
+  ioctl_ret->ret_unlink_daemon_exist = has_alive_unlink_daemon(ipc_ns);
```
atsushi421, line 852: On an abnormal daemon exit (SIGKILL, OOM), the entry is removed only when the exit worker reaches its pid, so until then `has_alive_unlink_daemon()` reports it alive and a process registering in that window (typical right after a process group is killed, when the queue is long) skips spawning a replacement; this is a second case besides the static-namespace one the PR body already notes. At minimum, document here that liveness is sampled only at ADD_PROCESS and lags the worker; if it matters, treat an entry whose task is gone (`pid_task(...)` NULL or `PF_EXITING`, guarded for KUNIT_BUILD) as absent.

src/agnocastlib/src/agnocast.cpp
```diff
@@ -551,6 +568,8 @@ struct initialize_agnocast_result initialize_agnocast(
   auto bridge_mode = get_bridge_mode();
 
   // Create a shm_unlink daemon process if it doesn't exist in its ipc namespace.
+  // ret_unlink_daemon_exist is only an early-out hint: poll_for_unlink() makes the singleton
```
atsushi421, line 576: With the hint now depending on the daemon's own registration, every process that registers during the fork-to-ioctl window forks a daemon, and since `spawn_daemon_process()` single-forks without `waitpid` or SIGCHLD handling, each loser stays a zombie of its parent application for the parent's lifetime (up to N-1 per launch burst). Suggest double-forking in `spawn_daemon_process()` (its return value is unused by all three callers) so losers are reparented to init, or at least noting the zombies in the planned follow-up.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -794,31 +795,64 @@ static bool has_alive_bridge_manager(const struct ipc_namespace * ipc_ns, const
...
+// Namespace-scoped, unlike the bridge manager.
+static bool has_alive_unlink_daemon(const struct ipc_namespace * ipc_ns)
+{
+  struct process_info * proc_info;
+  int bkt;
+  hash_for_each(proc_info_htable, bkt, proc_info, node)
+  {
+    if (
+      ipc_eq(ipc_ns, proc_info->ipc_ns) && proc_info->role == PROCESS_ROLE_UNLINK_DAEMON &&
+      !proc_info->exited) {
```
atsushi421: `exited` is only ever set in the non-daemon branch of `agnocast_process_exit_cleanup()`, so for a `PROCESS_ROLE_UNLINK_DAEMON` entry this predicate is always true. Suggest dropping it, or a one-line comment if it is meant as defensive.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -794,31 +795,64 @@ static bool has_alive_bridge_manager(const struct ipc_namespace * ipc_ns, const
...
 {
   int ret = 0;
 
+  if (
+    role != PROCESS_ROLE_APPLICATION && role != PROCESS_ROLE_BRIDGE_MANAGER &&
+    role != PROCESS_ROLE_UNLINK_DAEMON) {
+    return -EINVAL;
+  }
+
+  // Keeps the unlink daemon the only holder of AGNOCAST_DOMAIN_ID_NONE, which the domain-scoped
```
atsushi421: The domain-scoped counters exclude the daemon by `role`, not by `domain_id`, so nothing currently relies on the daemon being the only holder of `AGNOCAST_DOMAIN_ID_NONE`. Either reword this to say the check keeps applications from colliding with the sentinel, or make the sentinel the single mechanism and leave the domain-scoped counters unrenamed.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in 3be82821: deregistration now happens only on the idle poll, and a KUnit case covers a pid-returning commit with `caller_pid == daemon_pid`. `ret_daemon_should_exit` being derived on every commit while `poll_for_unlink()` acts only on the last one is the underlying confusion, so I will make the flag idle-poll-only in a follow-up PR rather than widen this one.

agnocast_kmod/agnocast_ioctl.c
Koichi98: I would leave this as is. One slot out of `mempool_num` does not seem worth making `process_info.mempool_entry` nullable while `agnocast_ioctl.c:730`, `770` and `1098` dereference it unconditionally, and the -ENOMEM case needs the pool to be full, where slots are released on process exit rather than by the daemon's drain, so a missing daemon does not sustain it. That said, happy to skip `assign_memory()` for the daemon if you would rather not spend the slot.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in e9a38db6: both rejections now `dev_warn` with the pid and the offending value, and `initialize_agnocast()` rejects the reserved id up front with a message naming `ROS_DOMAIN_ID`. Reserving a value out of the user's range is the awkward part, so a follow-up will move the sentinel to an internal `int64_t` field where user space cannot collide with it in the first place.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Documented at `has_alive_unlink_daemon()` in 685edc18. I am not planning to change the behaviour: it is not a regression (the previous `get_process_num() > 0` read the same table with the same worker lag) and the next process to register after the worker drains does spawn a replacement, so I filed the `PF_EXITING` check as #1592 to pick up if this window ever shows up in practice.

src/agnocastlib/src/agnocast.cpp
Koichi98: Deferring to the follow-up, and noted in the PR description. Double-forking changes the process topology for all three daemons, and the bridge manager and the discovery agent have had the same loser-exits shape since before this PR, so it is better as its own change than folded in here.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Dropped in 00e24731.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Reworded in be08eb9a. I kept the `role` predicate because `discovery_agent_should_exit_cmd()` passes an unvalidated `domain_id` from user space that would match the sentinel by domain; the follow-up moving `domain_id` to an internal `int64_t` drops both the predicate and the reserved value.

atsushi421 (approved):

PR #1591, opened 2026-08-28
fix(agnocastlib): request a wait set rebuild explicitly after MutuallyExclusive callbacks on Jazzy

atsushi421:
## Description

Fixes the Jazzy-only flaky failure of `MultiThreadedAgnocastExecutorTest.test_no_starvation_and_callback_group` (the `(false, *, "reentrant")` cases). It is not a timing margin of the test but a real bug in `MultiThreadedAgnocastExecutor` on rclcpp 28+. The test node's ROS 2 timers stop firing for good, so the agnocast callbacks registered afterwards are never notified.

Root cause (rclcpp 28 / rmw_fastrtps):

1. `build_entities_collection()` skips a callback group whose `can_be_taken_from` is `false`, so a wait set rebuild running while another thread executes a MutuallyExclusive callback removes every entity of that group. In the test that is the node's default group, which holds both timers.
2. Those entities come back only with the next rebuild, which `ros2_spin()` requested through `interrupt_guard_condition_->trigger()`.
3. rmw_fastrtps calls `set_trigger_value(false)` on every guard condition even when `rmw_wait()` returns `RMW_RET_TIMEOUT`, and `Executor::wait_for_work()` processes the notify waitable only on a `Ready` result. A trigger that coincides with a wait timeout is therefore lost and `entities_need_rebuild_` stays `false`.
4. With the ROS 2 subscriptions in a Reentrant group, no other MutuallyExclusive group produces further triggers, so nothing requests the rebuild again. The `"individual"` / `"mutually_exclusive"` cases self-heal through the triggers of the other groups, and Humble rebuilds the wait set on every `wait_for_work()`.

The fix sets the (protected) `entities_need_rebuild_` flag directly before triggering the guard condition, so the rebuild request cannot be lost. The guard condition is still triggered to wake threads blocked in `wait_for_work()`. The no-starvation tests now print the indices of the uncalled callbacks on failure, which keeps future failures diagnosable from the CI log alone. The mutual-exclusion flags in the test node become `std::atomic<bool>`, because several executor threads write them concurrently.

Verification on Jazzy (rclcpp 28.1.18, rmw_fastrtps) in a container pinned to 4 CPUs, with a bursty `SCHED_FIFO` hog stealing one CPU at random (mimics vCPU steal on the hosted runners; without it the failure never reproduced in 6,000+ runs), running the five `(false, *, "reentrant")` cases repeatedly:

- before the fix, 3 failures in 32 iterations (800 test executions), every failure with the same signature of timers firing a few times and then never again
- after the fix, 0 failures in 103 iterations (2,575 test executions)

On Humble, `test_integration_agnocastlib` (35 tests) passes locally, and the change is inside the existing `RCLCPP_VERSION_MAJOR >= 28` block.

The same combination exists in upstream rclcpp's `MultiThreadedExecutor` (also on Rolling), reported as ros2/rclcpp#3240.

## Related links

- https://github.com/autowarefoundation/agnocast/actions/runs/33147519113/job/98771730429
- https://github.com/autowarefoundation/agnocast/actions/runs/33051761689
- #1160 (previous attempt at this flake, which only enlarged the spin buffer)
- https://github.com/ros2/rclcpp/issues/3240 (upstream issue for the lost rebuild request)

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Post-merge checklist

**Important notes:**


atsushi421:
This issue has been resolved in upstream.

PR #1598, opened 2026-08-31
refactor(agnocast_cie_thread_configurator): move the lscpu hardware-info reader into the core library

atsushi421:
## Description

Part 1/3 of extracting the startup validation logic into unit-testable core functions; together with the two follow-up PRs this supersedes #1597.

Moves the lscpu hardware-info reader out of the installed shared library into the core static library, as a new `startup_checks.{hpp,cpp}` module, and splits `get_hardware_info` into its `popen` half and a pure `parse_lscpu_output` half so the parsing is testable without running lscpu.

- `get_hardware_info` is only called by the two daemon executables, and the installed shared library should not depend on the core static library, so the function moves to the core rather than delegating from `util.cpp`.
- Its declaration leaves the public umbrella header `cie_thread_configurator.hpp`: nothing outside this package uses it (checked repo-wide), and after the move the symbol is no longer in the installed `.so` anyway.

The move also hardens the reader:

- An empty result is reported instead of passing silently. Both daemons log a warning when lscpu yields none of the recorded keys, where an empty map previously made `validate_hardware_info` pass without comparing anything. Every failure mode funnels into that one case: a missing `/usr/bin/lscpu` is not a `popen` failure (the shell exits 127 with empty output), and empty output parses to an empty map.
- Keys are whitespace-trimmed so the hierarchic (indented) summary layout of util-linux >= 2.38 also matches; the comparison stays exact after the trim, so `BIOS Model name` still does not match `Model name`.
- Whitespace-only values are trimmed to empty strings instead of being stored verbatim.

The recorded lscpu labels are a single `kRecordedKeys` table rather than a seven-arm `if`/`else` chain, and one `trim` covers both the label and the value.

`test_startup_checks.cpp` pins the parser: only the recorded keys are kept, values are whitespace-trimmed (whitespace-only values become empty), indented keys match, and malformed or unknown lines are ignored.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Post-merge checklist

**Important notes:**

### DIFF
--- a/src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
+++ b/src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
@@ -299,15 +299,14 @@ void ThreadConfiguratorNode::validate_hardware_info(const YAML::Node & yaml)
   const YAML::Node & yaml_hw_info = yaml["hardware_info"];
   const auto current_hw_info = agnocast_cie_thread_configurator::get_hardware_info();
 
-  if (!current_hw_info) {
-    RCLCPP_WARN(
-      this->get_logger(), "Failed to read hardware info from lscpu. Skipping hardware validation.");
+  if (current_hw_info.empty()) {
+    RCLCPP_WARN(this->get_logger(), "No hardware info from lscpu. Skipping hardware validation.");
     return;
   }
 
   std::vector<std::string> mismatches;
 
-  for (const auto & [key, current_value] : *current_hw_info) {
+  for (const auto & [key, current_value] : current_hw_info) {
     if (!yaml_hw_info[key]) {
       continue;
     }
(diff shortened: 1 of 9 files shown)


src/agnocast_cie_thread_configurator/src/thread_configurator_node.cpp
```diff
@@ -298,6 +299,11 @@ void ThreadConfiguratorNode::validate_hardware_info(const YAML::Node & yaml)
   const YAML::Node & yaml_hw_info = yaml["hardware_info"];
   const auto current_hw_info = agnocast_cie_thread_configurator::get_hardware_info();
 
+  if (current_hw_info.empty()) {
+    RCLCPP_WARN(this->get_logger(), "No hardware info from lscpu. Skipping hardware validation.");
```
kobayu858, line 303: This guard covers the current-side map only, so when hardware_info shares no key with current_hw_info the loop compares nothing and the daemon logs "Hardware validation successful. Configuration matches this system."
An empty hardware_info is a defined YAML node, so the !yaml["hardware_info"] check above does not fire either.

atsushi421: Good catch. The comparison loop now counts the keys it actually compared, and when that count is zero (an empty hardware_info map, or one holding only keys lscpu does not report) it warns and skips instead of logging success, matching the two existing skip paths. Fixed in 06559cea.

kobayu858 (approved):

PR #1600, opened 2026-08-31
refactor(agnocast_cie_thread_configurator): extract rt_throttling validation into a testable check

atsushi421:
## Description

Part 3/3 of extracting the startup validation logic into unit-testable core functions (supersedes #1597 together with its neighbors).

Extracts the comparison half of `ThreadConfiguratorNode::validate_rt_throttling` into a pure core function:

- `check_rt_throttling(yaml, read_sysctl)` compares the `rt_throttling` section against the running kernel through an injected reader and returns per-key checks plus the `/etc/sysctl.d` guidance text (listing every configured key, matching or not, exactly as before). Writing these sysctls requires uid 0 (capabilities cannot bypass the proc sysctl DAC check), which is why the daemon checks and guides instead of writing; that rationale now lives on the declaration.
- The node keeps the reader (which logs its own read failures) and the per-key match/mismatch logging; all log lines are unchanged.
- One cosmetic difference: a failed sysctl read is now logged before the per-key result lines instead of interleaved with them, because the checks are computed before they are reported.

Tests cover the sysctl paths and their YAML key order, the byte-exact guidance text, and that unreadable sysctls are recorded but never count as a mismatch.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Notes for reviewers

Stacked on #1599 (base branch `refactor/cie-tc-check-hardware-info`); only the last commit belongs to this PR. Retarget to `main` once #1599 is merged.

## Post-merge checklist

**Important notes:**

### DIFF
--- a/src/agnocast_cie_thread_configurator/include/agnocast_cie_thread_configurator/startup_checks.hpp
+++ b/src/agnocast_cie_thread_configurator/include/agnocast_cie_thread_configurator/startup_checks.hpp
@@ -2,6 +2,7 @@
 
 #include "yaml-cpp/yaml.h"
 
+#include <functional>
 #include <map>
 #include <optional>
 #include <string>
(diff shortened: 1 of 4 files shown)


src/agnocast_cie_thread_configurator/include/agnocast_cie_thread_configurator/startup_checks.hpp
```diff
@@ -30,4 +31,33 @@ std::map<std::string, std::string> parse_lscpu_output(const std::string & output
...
+struct RtThrottlingCheck
+{
+  std::string key;  // "sched_rt_period_us" or "sched_rt_runtime_us"
+  int expected = 0;
+  std::optional<int> actual;  // nullopt: the sysctl could not be read
+};
+
+struct RtThrottlingReport
+{
+  std::vector<RtThrottlingCheck> checks;  // in YAML key order: period_us, runtime_us
```
kobayu858: The order comes from the hardcoded `{"period_us", "runtime_us"}` list, not from the document.
It is also the opposite way round for the config this tool generates: `prerun_node.cpp:173` writes runtime_us first, so a template-derived config has runtime first while the report comes back period first.
Only log line ordering depends on it, so nothing misbehaves - but `"fixed order: period_us, runtime_us"` would say what the code does, and the test name `ReadsTheKernelSysctlPathsInYamlKeyOrder` makes the same claim.

atsushi421: Right on both counts. Changed the comment to "in fixed order: period_us, runtime_us" and renamed the test to ReadsTheKernelSysctlPathsInFixedKeyOrder, since the order is the hardcoded key list and not the document order. Fixed in 13f3b2cf.

kobayu858 (approved):

PR #1602, opened 2026-08-31
fix(agnocast_kmod): count qos_depth in messages the subscriber can be handed

Koichi98:
## Description

`receive_msg_core()` derived the range to scan from `entry_id` before looking at what was in it:

```c
qos_start = newest_entry_id - sub_info->qos_depth + 1;
```

The loop that follows skips entries the subscriber may not be handed -- an exited publisher, `ignore_local_publications`, a domain that may not deliver -- but the range is already fixed and is never widened to make up for it, so every skipped entry still costs a slot. A subscriber can come away with nothing while a deliverable message sits one id below the range.

This counts `qos_depth` in messages the subscriber can be handed instead of in entry_ids. A topic's ids have holes in them, so this also changes what a subscriber gets with nothing skipped: with publisher A holding entry 0 and publisher B holding 4 and 5, a transient_local subscriber at depth 3 now receives {0, 4, 5} where it received {4, 5}.

## Related links

#1603

## How was this PR tested?

- [x] Autoware
- [ ] sample application

## Notes for reviewers

A single backward walk keeping the last `MAX_RECEIVE_NUM` entries it accepts in a ring would do this in one pass, with one `find_publisher_info()` per entry instead of two. It rewrites the delivery loop, and with a release close this seemed like the wrong moment for that diff, so it is left to #1603.

The two walks are bounded by the entries in the tree. `release_msgs_to_meet_depth()` holds each publisher to its own `qos_depth`, so the tree rests at the sum of the publishers' depths -- tens, not thousands -- and it already warns at `qos_depth + 100` for one publisher as a leak. A tree large enough for the second walk to cost anything means entry_nodes are leaking.

## Post-merge checklist

`need-patch-update`

### DIFF
--- a/agnocast_kmod/agnocast_ioctl.c
+++ b/agnocast_kmod/agnocast_ioctl.c
@@ -1157,6 +1157,25 @@ static struct rb_node * find_first_entry_ge(struct rb_root * root, const int64_t
   return candidate;
 }
 
+// Whether `sub_info` may be handed an entry published by `pub_info`.
+static bool is_entry_deliverable(
+  const struct topic_wrapper * wrapper, const struct subscriber_info * sub_info,
+  const struct publisher_info * pub_info)
+{
+  const struct process_info * proc_info = agnocast_find_process_info(pub_info->pid);
+  if (!proc_info || proc_info->exited) {
+    return false;
+  }
+
+  if (sub_info->ignore_local_publications && (sub_info->pid == pub_info->pid)) {
+    return false;
+  }
+
+  return domain_delivery_allowed(
+    wrapper->topic, pub_info->domain_id, pub_info->is_bridge, sub_info->domain_id,
+    sub_info->is_bridge);
+}
+
 static int receive_msg_core(
   struct topic_wrapper * wrapper, struct subscriber_info * sub_info,
   const topic_local_id_t subscriber_id, union ioctl_receive_msg_args * ioctl_ret)
@@ -1169,14 +1188,25 @@ static int receive_msg_core(
     return 0;
   }
 
-  const struct entry_node * newest_en = container_of(newest_node, struct entry_node, node);
-  const int64_t newest_entry_id = newest_en->entry_id;
-
-  // Calculate start_entry_id = max(newest - qos_depth + 1, latest_received_entry_id + 1)
-  const int64_t latest_received_entry_id = sub_info->latest_received_entry_id;
-  const int64_t qos_start = newest_entry_id - (int64_t)sub_info->qos_depth + 1;
-  const int64_t start_entry_id =
-    (qos_start > latest_received_entry_id) ? qos_start : (latest_received_entry_id + 1);
+  // start_entry_id is the qos_depth-th newest entry the subscriber can be handed, or its oldest
+  // unreceived entry when fewer than qos_depth of them are deliverable.
+  const int64_t oldest_wanted_entry_id = sub_info->latest_received_entry_id + 1;
+  int64_t start_entry_id = oldest_wanted_entry_id;
+  uint32_t deliverable_num = 0;
+  for (struct rb_node * back = newest_node; back; back = rb_prev(back)) {
+    const struct entry_node * en = container_of(back, struct entry_node, node);
+    if (en->entry_id < oldest_wanted_entry_id) {
+      break;
+    }
+    const struct publisher_info * pub_info = find_publisher_info(wrapper, en->publisher_id);
+    if (!pub_info || !is_entry_deliverable(wrapper, sub_info, pub_info)) {
+      continue;
+    }
+    start_entry_id = en->entry_id;
+    if (++deliverable_num >= sub_info->qos_depth) {
+      break;
+    }
+  }
 
   struct rb_node * node = find_first_entry_ge(&wrapper->topic->entries, start_entry_id);
 
@@ -1198,18 +1228,7 @@ static int receive_msg_core(
       return -ENODATA;
     }
 
-    const struct process_info * proc_info = agnocast_find_process_info(pub_info->pid);
-    if (!proc_info || proc_info->exited) {
-      continue;
-    }
-
-    if (sub_info->ignore_local_publications && (sub_info->pid == pub_info->pid)) {
-      continue;
-    }
-
-    if (!domain_delivery_allowed(
-          wrapper->topic, pub_info->domain_id, pub_info->is_bridge, sub_info->domain_id,
-          sub_info->is_bridge)) {
+    if (!is_entry_deliverable(wrapper, sub_info, pub_info)) {
       continue;
     }
 
--- a/agnocast_kmod/agnocast_ioctl.c
+++ b/agnocast_kmod/agnocast_ioctl.c
@@ -1197,7 +1197,7 @@ static int receive_msg_core(
     struct entry_node * en = container_of(node, struct entry_node, node);
 
     if (ioctl_ret->ret_entry_num == MAX_RECEIVE_NUM) {
-      ioctl_ret->ret_call_again = true;
+      ioctl_ret->ret_call_again = ioctl_ret->ret_entry_num < deliverable_num;
       break;
     }
 
--- a/agnocast_kmod/agnocast_ioctl.c
+++ b/agnocast_kmod/agnocast_ioctl.c
@@ -1138,25 +1138,6 @@ unlock_only_global:
   return ret;
 }
 
-// Find the first entry with entry_id >= target_entry_id
-static struct rb_node * find_first_entry_ge(struct rb_root * root, const int64_t target_entry_id)
-{
-  struct rb_node ** curr = &(root->rb_node);
-  struct rb_node * candidate = NULL;
-
-  while (*curr) {
-    const struct entry_node * en = container_of(*curr, struct entry_node, node);
-    if (en->entry_id >= target_entry_id) {
-      candidate = *curr;
-      curr = &((*curr)->rb_left);
-    } else {
-      curr = &((*curr)->rb_right);
-    }
-  }
-
-  return candidate;
-}
-
 // Whether `sub_info` may be handed an entry published by `pub_info`.
 static bool is_entry_deliverable(
   const struct topic_wrapper * wrapper, const struct subscriber_info * sub_info,
@@ -1192,10 +1173,10 @@ static int receive_msg_core(
     return 0;
   }
 
-  // start_entry_id is the qos_depth-th newest entry the subscriber can be handed, or its oldest
-  // unreceived entry when fewer than qos_depth of them are deliverable.
+  // node ends up on the qos_depth-th newest entry the subscriber can be handed, on its oldest
+  // unreceived deliverable entry when fewer than qos_depth of them are, and NULL when none are.
   const int64_t oldest_wanted_entry_id = sub_info->latest_received_entry_id + 1;
-  int64_t start_entry_id = oldest_wanted_entry_id;
+  struct rb_node * node = NULL;
   uint32_t deliverable_num = 0;
   for (struct rb_node * back = newest_node; back; back = rb_prev(back)) {
     const struct entry_node * en = container_of(back, struct entry_node, node);
@@ -1206,13 +1187,11 @@ static int receive_msg_core(
     if (!pub_info || !is_entry_deliverable(wrapper, sub_info, pub_info)) {
       continue;
     }
-    start_entry_id = en->entry_id;
+    node = back;
     if (++deliverable_num >= sub_info->qos_depth) {
       break;
     }
   }
-
-  struct rb_node * node = find_first_entry_ge(&wrapper->topic->entries, start_entry_id);
 
   for (; node; node = rb_next(node)) {
     struct entry_node * en = container_of(node, struct entry_node, node);
--- a/agnocast_kmod/agnocast_ioctl.c
+++ b/agnocast_kmod/agnocast_ioctl.c
@@ -1138,18 +1138,29 @@ unlock_only_global:
   return ret;
 }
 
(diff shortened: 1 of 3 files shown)


agnocast_kmod/agnocast_ioctl.c
```diff
@@ -1169,14 +1188,25 @@ static int receive_msg_core(
...
 
-  const struct entry_node * newest_en = container_of(newest_node, struct entry_node, node);
-  const int64_t newest_entry_id = newest_en->entry_id;
-
-  // Calculate start_entry_id = max(newest - qos_depth + 1, latest_received_entry_id + 1)
-  const int64_t latest_received_entry_id = sub_info->latest_received_entry_id;
-  const int64_t qos_start = newest_entry_id - (int64_t)sub_info->qos_depth + 1;
-  const int64_t start_entry_id =
-    (qos_start > latest_received_entry_id) ? qos_start : (latest_received_entry_id + 1);
+  // start_entry_id is the qos_depth-th newest entry the subscriber can be handed, or its oldest
```
atsushi421: The old code windowed by entry-id arithmetic, while this walk counts actual entries, so behaviour also changes when entry ids are non-contiguous even if nothing is skipped: a volatile subscriber registration consumes an id (`current_entry_id++` in add_subscriber) and `release_msgs_to_meet_depth()` leaves per-publisher holes. For example, with publisher A holding entry 0 and publisher B holding entries 4 and 5, a joining transient_local subscriber with depth 3 now receives {0, 4, 5} where it previously received {4, 5}. Counting messages matches take_msg and looks like the better semantics, but the description's claim that behaviour is unchanged whenever nothing in the range is skipped is not accurate; consider stating this second change in the PR body and pinning it with a kunit case.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -1169,14 +1188,25 @@ static int receive_msg_core(
...
+    const struct entry_node * en = container_of(back, struct entry_node, node);
+    if (en->entry_id < oldest_wanted_entry_id) {
+      break;
+    }
+    const struct publisher_info * pub_info = find_publisher_info(wrapper, en->publisher_id);
+    if (!pub_info || !is_entry_deliverable(wrapper, sub_info, pub_info)) {
+      continue;
+    }
+    start_entry_id = en->entry_id;
+    if (++deliverable_num >= sub_info->qos_depth) {
```
atsushi421, line 1208: `start_entry_id` is assigned before this check, so with `qos_depth == 0` the first deliverable entry is still delivered (1 >= 0), whereas the old code and take_msg deliver nothing. Nothing validates depth (agnocastlib passes `qos.depth()` straight through), so `rclcpp::QoS(0)` reaches this path; consider an early return for depth 0 before the walk, or rejecting depth 0 at add_subscriber.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -1169,14 +1188,25 @@ static int receive_msg_core(
...
+  const int64_t oldest_wanted_entry_id = sub_info->latest_received_entry_id + 1;
+  int64_t start_entry_id = oldest_wanted_entry_id;
+  uint32_t deliverable_num = 0;
+  for (struct rb_node * back = newest_node; back; back = rb_prev(back)) {
+    const struct entry_node * en = container_of(back, struct entry_node, node);
+    if (en->entry_id < oldest_wanted_entry_id) {
+      break;
+    }
+    const struct publisher_info * pub_info = find_publisher_info(wrapper, en->publisher_id);
+    if (!pub_info || !is_entry_deliverable(wrapper, sub_info, pub_info)) {
```
atsushi421: A NULL `pub_info` is silently skipped here, but the forward loop below treats the same state as "Unreachable" with a dev_warn and -ENODATA, so the two walks classify the same entry differently: below the final start_entry_id the corruption is never logged, and above it the ioctl fails anyway after this walk has finished scanning. Consider handling `!pub_info` the same way in both walks (warn and fail here too, or at least emit the warning).

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -1157,6 +1157,25 @@ static struct rb_node * find_first_entry_ge(struct rb_root * root, const int64_t
   return candidate;
 }
 
+// Whether `sub_info` may be handed an entry published by `pub_info`.
+static bool is_entry_deliverable(
```
atsushi421: agnocast_ioctl_take_msg still carries an inline copy of the same three checks (exited process, ignore_local_publications, domain_delivery_allowed) this helper now encapsulates, so the delivery predicate exists in two forms that can drift on a future rule change. Consider replacing those checks with a call to is_entry_deliverable(), either here or in #1603.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -1169,14 +1188,25 @@ static int receive_msg_core(
...
+  for (struct rb_node * back = newest_node; back; back = rb_prev(back)) {
+    const struct entry_node * en = container_of(back, struct entry_node, node);
+    if (en->entry_id < oldest_wanted_entry_id) {
+      break;
+    }
+    const struct publisher_info * pub_info = find_publisher_info(wrapper, en->publisher_id);
+    if (!pub_info || !is_entry_deliverable(wrapper, sub_info, pub_info)) {
+      continue;
+    }
+    start_entry_id = en->entry_id;
```
atsushi421: The walk is standing on the rb_node whose id becomes start_entry_id, but the pointer is discarded and find_first_entry_ge() re-derives the same node with an O(log n) descent, even though both walks run under the same continuously held locks. Saving the node here (e.g. `start_node = back;`) and using it when non-NULL avoids the re-search, and also allows skipping the guaranteed-empty forward pass when deliverable_num == 0.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -1169,14 +1188,25 @@ static int receive_msg_core(
...
-  // Calculate start_entry_id = max(newest - qos_depth + 1, latest_received_entry_id + 1)
-  const int64_t latest_received_entry_id = sub_info->latest_received_entry_id;
-  const int64_t qos_start = newest_entry_id - (int64_t)sub_info->qos_depth + 1;
-  const int64_t start_entry_id =
-    (qos_start > latest_received_entry_id) ? qos_start : (latest_received_entry_id + 1);
+  // start_entry_id is the qos_depth-th newest entry the subscriber can be handed, or its oldest
+  // unreceived entry when fewer than qos_depth of them are deliverable.
+  const int64_t oldest_wanted_entry_id = sub_info->latest_received_entry_id + 1;
+  int64_t start_entry_id = oldest_wanted_entry_id;
+  uint32_t deliverable_num = 0;
```
atsushi421, line 1193: The scanned range can now hold more than qos_depth entries, so with qos_depth == MAX_RECEIVE_NUM the forward loop can deliver everything and still set ret_call_again when the next node is merely non-deliverable, costing userspace one extra ioctl that returns zero entries. deliverable_num from this walk could suppress that, e.g. only set ret_call_again when ret_entry_num < deliverable_num.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -1157,6 +1157,25 @@ static struct rb_node * find_first_entry_ge(struct rb_root * root, const int64_t
   return candidate;
 }
 
+// Whether `sub_info` may be handed an entry published by `pub_info`.
```
atsushi421: This helper is only safe with global_htables_rwsem (read) held for agnocast_find_process_info() and proc_info->exited, plus wrapper->topic->rwsem (read) for pub_info stability; called outside those locks it would race with exit cleanup. Consider documenting the lock preconditions in this comment, as agnocast_find_process_info() does.

agnocast_kmod/agnocast_kunit/agnocast_kunit_receive_msg.c
```diff
@@ -662,6 +662,63 @@ void test_case_receive_msg_transient_publish_num_smaller_than_sub_qos_smaller_th
...
+
+  // Act
+  int ret = agnocast_ioctl_receive_msg(
+    TOPIC_NAME, current->nsproxy->ipc_ns, add_subscriber_args.ret_id, pub_shm_infos,
+    KUNIT_PUB_SHM_BUF_SIZE, &ioctl_receive_msg_ret);
+
+  // Assert
+  KUNIT_EXPECT_EQ(test, ret, 0);
+  KUNIT_EXPECT_EQ(test, ioctl_receive_msg_ret.ret_entry_num, 1);
+  KUNIT_EXPECT_EQ(test, ioctl_receive_msg_ret.ret_entry_ids[0], remote_publish_ret.ret_entry_id);
```
atsushi421, line 888: Neighbouring receive tests also assert agnocast_get_latest_received_entry_id() after the receive; without it, a regression that advances the watermark to the skipped local entry's id (or not at all) would pass. Consider asserting it equals remote_publish_ret.ret_entry_id, and optionally that a second receive returns zero entries.

agnocast_kmod/agnocast_kunit/agnocast_kunit_receive_msg.c
```diff
@@ -662,6 +662,63 @@ void test_case_receive_msg_transient_publish_num_smaller_than_sub_qos_smaller_th
 // ================================================
 // Tests for set_publisher_shm_info
 
+// qos_depth counts messages the subscriber can be handed: an entry it has to discard (here its
```
atsushi421, line 833: This test lands in the "Tests for set_publisher_shm_info" section although it verifies qos_depth counting and asserts nothing about shm info. Consider moving it above the banner next to the other qos_depth cases, adjusting the KUNIT_CASE order to keep the file and case list in sync.

agnocast_kmod/agnocast_kunit/agnocast_kunit_receive_msg.c
```diff
@@ -662,6 +662,63 @@ void test_case_receive_msg_transient_publish_num_smaller_than_sub_qos_smaller_th
...
+    agnocast_ioctl_publish_msg(
+      TOPIC_NAME, current->nsproxy->ipc_ns, local_publisher_id, local_ret_addr, &local_publish_ret),
+    0);
+
+  union ioctl_add_subscriber_args add_subscriber_args;
+  KUNIT_ASSERT_EQ(
+    test,
+    agnocast_ioctl_add_subscriber(
+      TOPIC_NAME, current->nsproxy->ipc_ns, NODE_NAME, subscriber_pid, qos_depth,
+      is_transient_local, IS_RELIABLE, IS_TAKE_SUB, true /* ignore_local_publications */, IS_BRIDGE,
```
atsushi421: This file uses file-scope named booleans (IS_RELIABLE, IS_TAKE_SUB, IS_BRIDGE, ...) so add_subscriber call sites read without inline comments; this is the only call passing a bare `true` with a compensating comment. Consider a named local such as `const bool ignore_local_publications = true;`.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in cd3bdf4f. The Description now states this change instead of claiming behaviour is unchanged whenever nothing is skipped.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in c3f930d7. Took the early return over rejecting depth 0 at add_subscriber, since agnocastlib deliberately warns and continues, and pinned it with a kunit case.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in 28ee9aa6. Rather than repeating the warning in the backward walk, the publisher lookup moved into is_entry_deliverable(), so both walks reach it through one path.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in e16827f9, in this PR rather than deferring it to #1603.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in 4bcb7517. find_first_entry_ge() lost its last caller and is removed with it.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in 12c4013d, with a kunit case for the qos_depth == MAX_RECEIVE_NUM shape.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in 3a3b34c0. agnocast_find_process_info() itself carries no such comment, so I followed the `// Caller holds ...` form used elsewhere in this file.

agnocast_kmod/agnocast_kunit/agnocast_kunit_receive_msg.c
Koichi98: Fixed in faa57f62. Left out the second receive: test_case_receive_msg_ignore_local_same_pid_enabled already covers a same-process publication not being delivered.

agnocast_kmod/agnocast_kunit/agnocast_kunit_receive_msg.c
Koichi98: Fixed in faa57f62.

agnocast_kmod/agnocast_kunit/agnocast_kunit_receive_msg.c
Koichi98: Fixed in faa57f62.

atsushi421 (approved):

atsushi421 (approved):

PR #1613, opened 2026-09-04
fix(agnocastlib): count same-process subscribers like rclcpp does

Koichi98:
## Description

`Publisher::get_subscription_count()` reported `ret_other_process_subscriber_num`, so a publisher whose only subscriber lives in its own process reported 0, while `rclcpp::Publisher::get_subscription_count()` reports every matched subscription. It now includes the same-process count, and `NodeGraph::count_subscribers()`, which delegates here, stops missing them too.

`get_intra_process_subscription_count()` is added. It reports the same-process count unconditionally, because Agnocast delivers to every same-process subscriber through shared memory. It deliberately does not reproduce the number `rclcpp::Publisher::get_intra_process_subscription_count()` returns: rclcpp also excludes a subscriber whose own endpoint has intra-process disabled, and one whose QoS is incompatible, neither of which Agnocast tracks, so reading `use_intra_process_comms` here would buy no parity while making the count a function of `NodeOptions` rather than of the running system. Reporting the same-process count also keeps rclcpp's own comparison for "is an inter-process publish needed" working here.

`get_intra_subscription_count()` is deprecated, because its callers summed it with `get_subscription_count()` to get a total that now double-counts. It is a renamed alias of `get_intra_process_subscription_count()`.

`NodeBase::get_use_intra_process_default()` warned on every call without looking at the value, so the warning fired even for the `rclcpp::NodeOptions` default of `false`. The warning moves to the `NodeBase` constructor and fires only when the option was set, once per node; the getter is now silent.

## Related links

None.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Notes for reviewers

The existing log-capture helpers in `test_agnocast_utils.cpp` and `test_agnocast_generic_publisher.cpp` now call `rcutils_logging_initialize()` before taking the output handler, matching the new helper in `test_node_base.cpp`. Two lines each, unrelated to the counts.

## Post-merge checklist

## Interface changes

`agnocast::PublisherBase` gains `get_intra_process_subscription_count()`. `get_intra_subscription_count()` is marked `[[deprecated]]` and delegates to it. The internal helper `get_intra_subscription_count_core()` is renamed to `get_same_process_subscription_count_core()`.

`agnocast::node_interfaces::NodeBase::get_use_intra_process_default()` no longer logs; the signature is unchanged.

## Effects on system behavior

`Publisher::get_subscription_count()` and `NodeGraph::count_subscribers()` now include Agnocast subscribers in the caller's own process. Code that summed `get_subscription_count()` and `get_intra_subscription_count()` to get a total double-counts them; `agnocast_e2e_test`'s publisher is updated accordingly.

A same-process subscriber that set `ignore_local_publications` is counted even though the kernel module never delivers to it. Whether rclcpp counts such a subscription depends on the RMW: rmw_cyclonedds sets the option as a DDS QoS so the endpoints never match and it drops out, while rmw_fastrtps applies it when taking a message and keeps it. Only the exception is documented here.

A node created with `use_intra_process_comms` set now logs one warning at construction, where previously nothing was logged unless something called `get_use_intra_process_default()`. A node that leaves the option at its default logs nothing.

### DIFF
--- a/src/agnocastlib/src/agnocast_publisher.cpp
+++ b/src/agnocastlib/src/agnocast_publisher.cpp
@@ -109,7 +109,8 @@ uint32_t get_subscription_count_core(const std::string & topic_name)
   }
 
   uint32_t inter_count = args.ret_other_process_subscriber_num;
-  // If an A2R bridge exists, exclude the agnocast subscriber created by the bridge
+  // If an A2R bridge exists, exclude the agnocast subscriber created by the bridge. The bridge
+  // runs in a forked process, so it is never part of the same-process count.
   if (args.ret_a2r_bridge_exist && inter_count > 0) {
     inter_count--;
   }
@@ -120,10 +121,10 @@ uint32_t get_subscription_count_core(const std::string & topic_name)
     ros2_count--;
   }
 
-  return inter_count + ros2_count;
+  return inter_count + args.ret_same_process_subscriber_num + ros2_count;
 }
 
-uint32_t get_intra_subscription_count_core(const std::string & topic_name)
+uint32_t get_same_process_subscription_count_core(const std::string & topic_name)
 {
   union ioctl_get_subscriber_num_args get_subscriber_count_args = {};
   get_subscriber_count_args.topic_name = {topic_name.c_str(), topic_name.size()};
--- a/src/agnocastlib/test/unit/test_node_base.cpp
+++ b/src/agnocastlib/test/unit/test_node_base.cpp
@@ -4,6 +4,67 @@
 #include "rclcpp/version.h"
 
 #include <gtest/gtest.h>
+#include <rcutils/logging.h>
+
+#include <cstdarg>
+#include <cstdio>
+#include <string>
+#include <utility>
+
+namespace
+{
+
+// An rcutils output handler is a plain function pointer, so what it counts into lives here.
+const std::string * g_capture_needle = nullptr;
+int * g_capture_count = nullptr;
+
+void count_matching_warnings(
+  const rcutils_log_location_t * /*location*/, int severity, const char * /*name*/,
+  rcutils_time_point_value_t /*timestamp*/, const char * format, va_list * args)
+{
+  if (severity != RCUTILS_LOG_SEVERITY_WARN) {
+    return;
+  }
+  char buf[1024];
+  va_list args_copy;
+  va_copy(args_copy, *args);
+  vsnprintf(buf, sizeof(buf), format, args_copy);
+  va_end(args_copy);
+  if (std::string(buf).find(*g_capture_needle) != std::string::npos) {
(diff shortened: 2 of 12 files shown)


src/agnocastlib/src/agnocast_publisher.cpp
```diff
@@ -120,10 +121,10 @@ uint32_t get_subscription_count_core(const std::string & topic_name)
     ros2_count--;
   }
 
-  return inter_count + ros2_count;
+  return inter_count + args.ret_same_process_subscriber_num + ros2_count;
```
kobayu858, line 124: Same-process subscribers that set `ignore_local_publications` are counted here, even though the kmod never delivers to them: the filter `ignore_local_publications && sub_info->pid == pub_info->pid` exists in the notify and receive paths (`agnocast_ioctl.c:292`, `:1172`) but not in `agnocast_ioctl_get_subscriber_num()` (`:1489`). Before this change the option was irrelevant here, because the same-process bucket was not added in.

Only opt-in subscriptions are affected, but for those, code that waits for `get_subscription_count()` to reach an expected number counts a peer that never receives anything. Note `NodeGraph::count_subscribers()` shares this core and should *not* drop them — a graph query is about existence, not reachability. So this needs either a separate publisher-facing count, or just a line in the docs noting the exception.

src/agnocastlib/test/unit/test_node_base.cpp
```diff
@@ -544,6 +608,25 @@ TEST_F(TestNodeBase, get_use_intra_process_default_reflects_node_options_false)
   EXPECT_FALSE(value);
 }
 
+TEST_F(TestNodeBase, construction_warns_once_when_use_intra_process_comms_is_set)
```
kobayu858, line 611: This pins "warns once when set", but not what the PR fixes: the old code warned on every call without looking at the value, so it fired for the default `false` too. Could we add the negative case — leave the option untouched, call the getter, and `EXPECT_EQ(0, warn_count)` under the same `WarningCapture`?

src/agnocastlib/src/agnocast_publisher.cpp
Koichi98: It turns out to be RMW-dependent: rmw_cyclonedds sets the option as a DDS QoS so the endpoints never match and the subscription drops out of the publisher's count, while rmw_fastrtps applies it when taking a message and keeps it. So we do diverge from cyclonedds here, but I'm leaving it at the documentation for this PR (869cbba1).

src/agnocastlib/test/unit/test_node_base.cpp
Koichi98: Added in 9a81e1f3.

kobayu858 (approved):

PR #1617, opened 2026-09-08
fix(agnocast_kmod): skip response topics and raise the topic list limit

Koichi98:
## Description

`GET_TOPIC_LIST` fails with `-ENOBUFS` as soon as one IPC namespace holds `MAX_TOPIC_NUM` topics, so the discovery agent gets no topics at all and requests no bridges. A full Autoware launch reaches 1205 topics.

- Skip `/AGNOCAST_SRV_RESPONSE/...` topics. Nothing reads them: `decide_service_bridges()` keys a service off its request topic, and `decide_bridges()` skips both service prefixes. They were roughly 40% of the 1205, mostly the six parameter services of every `agnocast::Node`.
- Raise `MAX_TOPIC_NUM` to 2048.
- Fill kernel scratch under `global_htables_rwsem` and `copy_to_user()` after `up_read()`, the shape `GET_NODE_NAMES` already uses. The per-topic copy under the lock could sleep, and the higher limit widens that window.

A namespace above the new limit still fails with `-ENOBUFS`. Truncating instead needs a return field in `union ioctl_topic_list_args`, so it is left to a follow-up.

`union ioctl_topic_list_args` is unchanged, and a kmod/wrapper skew on `MAX_TOPIC_NUM` degrades to the lower of the two, hence `need-patch-update`.

## Related links

None.

## How was this PR tested?

- [x] Autoware
- [ ] sample application

## Notes for reviewers

- `ros2 topic list_agnocast` no longer shows response topics. They were only ever per-client plumbing.
- The load itself is untouched: on Humble `ServiceBridgeItem::ros2_client_exists()` returns `true` unconditionally (`count_clients()` needs Jazzy), so every parameter service of every node gets an R2A bridge and a response topic with it. Worth a separate issue.

## Post-merge checklist

`need-patch-update`

### DIFF
--- a/agnocast_kmod/agnocast.h
+++ b/agnocast_kmod/agnocast.h
@@ -363,7 +363,7 @@ struct ioctl_add_domain_bridge_prefix_args
 // ================================================
 // ros2cli ioctls
 
-#define MAX_TOPIC_NUM 1024
+#define MAX_TOPIC_NUM 2048
 
 union ioctl_topic_list_args {
   struct
@@ -476,7 +476,8 @@ int agnocast_ioctl_get_publisher_num(
   union ioctl_get_publisher_num_args * ioctl_ret);
 
 int agnocast_ioctl_get_topic_list(
-  const struct ipc_namespace * ipc_ns, union ioctl_topic_list_args * topic_list_args);
+  const struct ipc_namespace * ipc_ns, char * topic_name_buf, uint32_t * domain_id_buf,
+  const uint32_t buf_topic_num, uint32_t * ret_topic_num);
 
 int agnocast_ioctl_get_node_names(
   const struct ipc_namespace * ipc_ns, const pid_t pid, char * buf, const uint32_t buf_node_num,
--- a/agnocast_kmod/agnocast.h
+++ b/agnocast_kmod/agnocast.h
@@ -523,12 +523,12 @@ int agnocast_ioctl_get_topic_publisher_info(
   union ioctl_topic_info_args * topic_info_args);
 
 int agnocast_ioctl_get_node_subscriber_topics(
-  const struct ipc_namespace * ipc_ns, const char * node_name,
-  union ioctl_node_info_args * node_info_args);
+  const struct ipc_namespace * ipc_ns, const char * node_name, char * topic_name_buf,
+  const uint32_t buf_topic_num, uint32_t * ret_topic_num);
 
 int agnocast_ioctl_get_node_publisher_topics(
-  const struct ipc_namespace * ipc_ns, const char * node_name,
-  union ioctl_node_info_args * node_info_args);
+  const struct ipc_namespace * ipc_ns, const char * node_name, char * topic_name_buf,
+  const uint32_t buf_topic_num, uint32_t * ret_topic_num);
 
 int agnocast_ioctl_check_and_request_bridge_shutdown(
   const pid_t pid, const struct ipc_namespace * ipc_ns,
--- a/agnocast_kmod/agnocast_ioctl.c
+++ b/agnocast_kmod/agnocast_ioctl.c
@@ -1708,13 +1708,10 @@ int agnocast_ioctl_get_topic_list(
       goto unlock;
     }
 
-    memcpy(
+    strscpy_pad(
       topic_name_buf + (size_t)topic_num * TOPIC_NAME_BUFFER_SIZE, wrapper->key,
-      strlen(wrapper->key) + 1);
-
-    if (domain_id_buf) {
-      domain_id_buf[topic_num] = wrapper->domain_id;
-    }
+      TOPIC_NAME_BUFFER_SIZE);
+    domain_id_buf[topic_num] = wrapper->domain_id;
 
     topic_num++;
   }
@@ -3280,17 +3277,14 @@ static long get_topic_list_cmd(union ioctl_topic_list_args __user * arg)
   uint32_t __user * user_domain_id_buf =
     (uint32_t __user *)u64_to_user_ptr(topic_list_args.domain_id_buffer_addr);
 
-  char * topic_name_buf = kvzalloc((size_t)buf_topic_num * TOPIC_NAME_BUFFER_SIZE, GFP_KERNEL);
+  // One allocation for both arrays: a name slot is TOPIC_NAME_BUFFER_SIZE bytes, so the domain ids
+  // that follow the names stay uint32_t-aligned. agnocast_ioctl_get_topic_list writes every byte
+  // that is copied out, so the scratch needs no zeroing.
+  const size_t names_bytes = (size_t)buf_topic_num * TOPIC_NAME_BUFFER_SIZE;
+  char * topic_name_buf =
+    kvmalloc(names_bytes + (size_t)buf_topic_num * sizeof(uint32_t), GFP_KERNEL);
   if (!topic_name_buf) return -ENOMEM;
-
-  uint32_t * domain_id_buf = NULL;
-  if (user_domain_id_buf) {
-    domain_id_buf = kvcalloc(buf_topic_num, sizeof(*domain_id_buf), GFP_KERNEL);
-    if (!domain_id_buf) {
-      kvfree(topic_name_buf);
-      return -ENOMEM;
-    }
-  }
+  uint32_t * domain_id_buf = (uint32_t *)(topic_name_buf + names_bytes);
 
   uint32_t topic_num = 0;
   long ret =
@@ -3311,7 +3305,7 @@ static long get_topic_list_cmd(union ioctl_topic_list_args __user * arg)
   }
 
   if (
-    domain_id_buf &&
+    user_domain_id_buf &&
     copy_to_user(user_domain_id_buf, domain_id_buf, (size_t)topic_num * sizeof(*domain_id_buf))) {
     ret = -EFAULT;
     goto free;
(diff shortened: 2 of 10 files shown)


agnocast_kmod/agnocast_ioctl.c
```diff
@@ -1690,36 +1699,27 @@ int agnocast_ioctl_get_topic_list(
       continue;
     }
 
-    if (topic_num >= MAX_TOPIC_NUM || topic_num >= topic_list_args->topic_name_buffer_size) {
-      dev_warn(
-        agnocast_device, "Topic count exceeds limit: MAX_TOPIC_NUM=%d, topic_name_buffer_size=%u\n",
-        MAX_TOPIC_NUM, topic_list_args->topic_name_buffer_size);
-      ret = -ENOBUFS;
-      goto unlock;
+    if (is_service_response_topic(wrapper->key)) {
```
atsushi421, line 1702: `ros2 node info_agnocast` reads response topics. `get_agnocast_node_topics()` in `node_info_agnocast.py` builds its Service Clients set from `/AGNOCAST_SRV_RESPONSE...` names in the gossip snapshots, and those snapshots come from this ioctl on both the gossip path (`agent.py::read_local_topics`) and the ioctl fallback (`_ioctl.py::self_ns_snapshot`). After this change the client set is always empty, and a node whose only Agnocast endpoints are service clients falls through the `elif any([..., agnocast_clients])` check and is reported as nonexistent.

The cheapest fix is on the consumer side. A client publishes on the request topic, so the request-topic branch can add the service to the client set when `is_pub`, and the response-topic branch can go. Alternatively drop the skip here, since the raised limit alone covers the 1205 topics.

agnocast_kmod/agnocast.h
```diff
@@ -363,7 +363,7 @@ struct ioctl_add_domain_bridge_prefix_args
 // ================================================
 // ros2cli ioctls
 
-#define MAX_TOPIC_NUM 1024
+#define MAX_TOPIC_NUM 2048
```
atsushi421, line 366: `agnocast_ioctl_get_node_subscriber_topics` and `agnocast_ioctl_get_node_publisher_topics` still `copy_to_user` per topic under `global_htables_rwsem` and clamp against this constant, so the rationale in the description (the copy can sleep, the higher limit widens the window) applies to them too, but only the constant changed. Either give them the same scratch-then-copy shape or note them as a follow-up in the description.

Also worth a one-line comment here and in `src/agnocast_ioctl_wrapper/src/agnocast_ioctl.hpp` that the two definitions must match, since the description already calls out the skew risk.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -200,6 +200,14 @@ static bool is_parameter_service_topic(const char * key)
          strstr(key, "/list_parameters");
 }
 
+// A response topic exists per (service, client), and every consumer of the topic list keys a
+// service off its request topic instead: the discovery agent reads the roles it bridges on from
+// the request topic's endpoints.
+static bool is_service_response_topic(const char * key)
+{
+  return strncmp(key, "/AGNOCAST_SRV_RESPONSE", sizeof("/AGNOCAST_SRV_RESPONSE") - 1) == 0;
```
atsushi421: `str_has_prefix(key, "/AGNOCAST_SRV_RESPONSE")` from `<linux/string.h>` does this without spelling the literal twice, and removes the class of bug the kunit comment at `lists_service_request_topic` guards against.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -1690,36 +1699,27 @@ int agnocast_ioctl_get_topic_list(
...
 
-    if (topic_list_args->domain_id_buffer_addr) {
-      uint32_t domain_id = wrapper->domain_id;
-      uint32_t __user * domain_id_buffer =
-        (uint32_t __user *)u64_to_user_ptr(topic_list_args->domain_id_buffer_addr);
-      if (copy_to_user(domain_id_buffer + topic_num, &domain_id, sizeof(domain_id))) {
-        ret = -EFAULT;
-        goto unlock;
-      }
+    memcpy(
```
atsushi421: This walks the key twice and relies on `copy_name_from_user` keeping `strlen(key) < TOPIC_NAME_BUFFER_SIZE`. The file already copies bounded names into fixed slots with `strscpy` (lines 2051 and 2141). `strscpy_pad(slot, wrapper->key, TOPIC_NAME_BUFFER_SIZE)` is one call bounded by the destination and zero-pads the slot, which also lets the zero-fill in `get_topic_list_cmd` go.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -3268,15 +3268,61 @@ static long get_exit_process_cmd(struct ioctl_get_exit_process_args __user * arg
...
-    if (copy_to_user(arg, &topic_list_args, sizeof(topic_list_args))) return -EFAULT;
+
+  const uint32_t buf_topic_num =
+    min_t(uint32_t, topic_list_args.topic_name_buffer_size, MAX_TOPIC_NUM);
+  char __user * user_topic_name_buf =
+    (char __user *)u64_to_user_ptr(topic_list_args.topic_name_buffer_addr);
+  uint32_t __user * user_domain_id_buf =
+    (uint32_t __user *)u64_to_user_ptr(topic_list_args.domain_id_buffer_addr);
+
+  char * topic_name_buf = kvzalloc((size_t)buf_topic_num * TOPIC_NAME_BUFFER_SIZE, GFP_KERNEL);
```
atsushi421: The wrapper always passes `MAX_TOPIC_NUM`, so every call (1 Hz per (NS, domain) from the discovery agent, through a 0666 device) allocates and zeroes 512 KiB, usually via vmalloc at order 7. The zeroing is only needed because whole 256-byte slots are copied out. With `strscpy_pad` in the core this can be `kvmalloc`, and `domain_id_buf` can be `kvmalloc_array` since every used index is written.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -3268,15 +3268,61 @@ static long get_exit_process_cmd(struct ioctl_get_exit_process_args __user * arg
...
+  if (!topic_name_buf) return -ENOMEM;
+
+  uint32_t * domain_id_buf = NULL;
+  if (user_domain_id_buf) {
+    domain_id_buf = kvcalloc(buf_topic_num, sizeof(*domain_id_buf), GFP_KERNEL);
```
atsushi421: The second allocation exists only to honour a null `domain_id_buffer_addr`, and costs a nested error path, a second `kvfree`, and the `if (domain_id_buf)` branch inside the locked loop. The domain array is 8 KiB next to 512 KiB, so one `kvmalloc(names_bytes + ids_bytes)` with `domain_id_buf` pointing past the names, the core always filling it, and only the final `copy_to_user` conditional on the user pointer would be simpler. The core then loses its nullable parameter.

agnocast_kmod/agnocast_kunit/agnocast_kunit_get_topic_list.c
```diff
@@ -0,0 +1,150 @@
...
+    current->nsproxy->ipc_ns, (char *)buf, NULL, ARRAY_SIZE(buf), &topic_num);
+
+  // Assert
+  KUNIT_EXPECT_EQ(test, ret, 0);
+  KUNIT_EXPECT_EQ(test, topic_num, 0);
+}
+
+void test_case_get_topic_list_skips_service_response_topic(struct kunit * test)
+{
+  char buf[2][TOPIC_NAME_BUFFER_SIZE];
```
atsushi421: With two slots and two topics, swapping the order of `is_service_response_topic()` and the `topic_num >= buf_topic_num` check would still pass, so the test does not pin that response topics stay out of the count. `char buf[1][TOPIC_NAME_BUFFER_SIZE]` would.

Similarly `fails_when_buffer_is_full` initialises `topic_num` to 0 and only checks the return code, so it cannot tell an untouched `ret_topic_num` from a partial write. A `UINT_MAX` sentinel plus an equality check would cover the contract `get_topic_list_cmd` relies on.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in e148da11.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in 72069100.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in ffe246ef.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in ffe246ef, except that `domain_id_buf` is not a separate `kvmalloc_array` — it shares the single `kvmalloc` per your next comment.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in ffe246ef. The kunit call sites now pass a domain array, since the parameter is no longer nullable.

agnocast_kmod/agnocast_kunit/agnocast_kunit_get_topic_list.c
Koichi98: Fixed in 2c8d297c and 7eb7c40c. A single slot only pinned the ordering when the listable topic happened to be walked first, so `skips_service_response_topic` is now a zero-size buffer with only a response topic, which fails whichever topic the walk reaches first.

agnocast_kmod/agnocast.h
Koichi98: Fixed in 7a9645c0. I left out the matching comment on `MAX_TOPIC_NUM`, since the whole of `agnocast_ioctl.hpp` mirrors `agnocast.h` and a skew there is absorbed by the `min_t` clamp anyway.

atsushi421 (approved):

PR #1624, opened 2026-09-09
feat(ros2agnocast_discovery_agent): read domain bridge rules from several configs

Koichi98:
## Description

`AGNOCAST_DOMAIN_BRIDGE_CONFIG` and `register_domain_bridge --config` now accept several `domain_bridge` YAMLs instead of one, so rules can be split per subsystem the way `domain_bridge` itself already allows (it takes several positional configs). The environment variable holds a `:`-separated list; `--config` takes several paths. Files apply in the order given: `topics` accumulate, while `from_domain` / `to_domain` stay local to the file that sets them, matching the external node's merge.

Splitting also works with no environment variable at all: every `*.yaml` in `/etc/agnocast/domain_bridge.d/` is read in name order, after `/etc/agnocast/domain_bridge.yaml` when that exists — the pairing systemd and sysctl use for a configuration file and its drop-in directory. That matters because the discovery agent is `execv`'d from an application process and so never sees a `--config` argument.

A single path keeps working unchanged, so existing deployments need no edit: the drop-in directory cannot pre-exist this change.

Unlike `domain_bridge`, which aborts the whole process on the first unreadable file, a bad config here is reported and skipped so the rest still apply: the agent cannot abort, since it is `execv`'d from an application process, and `register_domain_bridge` still exits non-zero so a boot-time misconfiguration is loud.

## Related links

None.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

`colcon test` for `ros2agnocast_discovery_agent` (139 tests, all passing), covering the separator split, the empty-listing fallback, ordered accumulation across files, per-file domain defaults, and a broken config being skipped while its neighbours apply.

## Notes for reviewers

The `--config` default is still resolved from the environment, so the agent and the registration tool read the same list.

#1626 marks rule *registration* unsupported, but the discovery agent reads the same configs to force the A2R bridge the external `domain_bridge` node needs, which stays supported. That is the half this change is for.

## Post-merge checklist

https://github.com/autowarefoundation/agnocast_doc/pull/58

`need-patch-update`

### DIFF
--- a/src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/domain_bridge_config.py
+++ b/src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/domain_bridge_config.py
@@ -1,11 +1,13 @@
-"""Parse a ROS 2 ``domain_bridge`` YAML into kmod rule tuples.
+"""Parse ROS 2 ``domain_bridge`` YAMLs into kmod rule tuples.
 
-Each rule is ``(from_topic, to_topic, from_domain, to_domain)``. The same YAML
-drives both the external ``domain_bridge`` node (cross-ECU, via DDS) and the kmod
+Each rule is ``(from_topic, to_topic, from_domain, to_domain)``. The same YAMLs
+drive both the external ``domain_bridge`` node (cross-ECU, via DDS) and the kmod
 rule injection that opens same-IPC-namespace zero-copy cross-domain delivery. The
 topic name, its ``remap`` target, and the domain pair matter here; ``type`` and
 other fields are ignored.
 """
+from collections import namedtuple
+import glob
 import os
 
 import yaml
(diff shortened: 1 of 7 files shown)


src/ros2agnocast_discovery_agent/ros2agnocast_discovery_agent/domain_bridge_config.py
```diff
@@ -17,15 +19,42 @@
...
+
+    Names ``DEFAULT_CONFIG_PATH`` when neither exists, so the caller can say where a config
+    would go.
+    """
+    paths = [DEFAULT_CONFIG_PATH] if os.path.isfile(DEFAULT_CONFIG_PATH) else []
+    paths += sorted(glob.glob(os.path.join(default_config_dir(), '*.yaml')))
+    return paths or [DEFAULT_CONFIG_PATH]
+
+
+def resolve_config_paths():
```
kobayu858, line 64: Setting `AGNOCAST_DOMAIN_BRIDGE_CONFIG` suppresses the drop-in directory entirely, and nothing says so. README L46 pushes operators toward exactly that: export the variable, then split rules into `domain_bridge.d/` — the drop-in rules are dropped silently, both in the tool and in the agent, and those topics sit in the deadlock this forcing exists to break.

Could we either state the precedence in the README, or warn from `_load_domain_rules` when the env var is set and `default_config_dir()` holds a `*.yaml`? I'd prefer the warning, since the failure is silent at runtime.

Koichi98: Fixed in fc2b05e6. Took both of the options you offered. `test_agent.py` also gained an autouse fixture pinning the default path into `tmp_path`, since the new lookup would otherwise read the real `/etc/agnocast/domain_bridge.d`.

kobayu858 (approved):

PR #1625, opened 2026-09-10
fix(agnocast_kmod): retain entries for the deepest subscriber

Koichi98:
## Description

`release_msgs_to_meet_depth()` bounded retention by the publisher's `qos_depth` alone, making the effective queue `min(pub depth, sub depth)`. DDS holds the two separately and History is not an RxO policy, so `pub 10 / sub 100` is intended: with 30 messages published, ROS 2 hands back 30 and Agnocast 10. `autoware_pointcloud_preprocessor` needs exactly this, taking twist and odom at depth 100 against publishers at 10. The take side already walks back `sub_info->qos_depth`, so only retention changed.

Retention now covers the deepest subscriber, under two bounds:

- **Reachability.** `sub_info_htable` is shared by the two domain cells a bridge rule pairs, so it holds subscribers a publisher can never deliver to. `sub_can_receive_from()` extracts the conditions `rebuild_notify_list()` and `is_entry_deliverable()` already shared, and `retention_depth()` asks it too.
- **Lag, not declared depth.** Nothing at or below `latest_received_entry_id` is handed over again, so a subscriber that keeps up costs nothing however deep it is declared.

`MAX_QOS_DEPTH` (4096) clamps the stored subscriber depth, which now spends memory in another process's mempool. The deepest subscriber in Autoware is 100, and 1000 across the whole tree.

## Related links

- <https://github.com/autowarefoundation/autoware_core/pull/1443>

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

`make all` and `make test` build clean under `-Wall -Werror`, and the KUnit suite passes.

New cases: the deepest subscriber sets retention (5), the lag bound (2), reachability (2), the clamp (2).

## Notes for reviewers

`retention_depth()` scans `sub_info_htable` on every publish, safe under the `global_htables_rwsem` read that publish holds. Caching the maximum was the alternative, but the lag bound moves it on every receive.

## Post-merge checklist

`need-patch-update` -- no ioctl command, struct, or shared data structure changes, so kmod/agnocastlib compatibility is unaffected.

### DIFF
--- a/agnocast_kmod/agnocast.h
+++ b/agnocast_kmod/agnocast.h
@@ -9,6 +9,12 @@
 #define MAX_TOPIC_LOCAL_ID 4096  // Bitmap size for per-entry subscriber reference tracking
 #define MAX_SUBSCRIBER_NUM \
   (MAX_TOPIC_LOCAL_ID - MAX_PUBLISHER_NUM)  // Maximum number of subscribers per topic
+/* Maximum subscriber qos_depth; a deeper request is refused. A subscriber's depth decides how long
+ * its publishers must hold entries open, so it spends memory in another process's mempool. The
+ * bound is well above what exists in practice -- the deepest subscriber in Autoware is 100
+ * (autoware_pointcloud_preprocessor's twist and odom), and 1000 across the whole tree -- so a
+ * request over it is a bug worth surfacing rather than silently narrowing. */
+#define MAX_QOS_DEPTH 4096
 /* Maximum number of entries that can be received at one ioctl. This value is heuristically set to
  * balance the number of calling ioctl and the overhead of copying data between user and kernel
  * space. */
--- a/agnocast_kmod/agnocast_ioctl.c
+++ b/agnocast_kmod/agnocast_ioctl.c
@@ -988,19 +988,41 @@ unlock:
   return ret;
 }
 
+// A subscriber has no cache of its own: its qos_depth is a window onto the publisher's entries, so
+// an entry released once the publisher's depth is met is gone for every subscriber that had not
+// read it yet.
+//
+// Caller holds global_htables_rwsem (read), which excludes subscriber join and leave, and
+// wrapper->topic->rwsem (write).
+static uint32_t retention_depth(
+  struct topic_wrapper * wrapper, const struct publisher_info * pub_info)
+{
+  uint32_t depth = pub_info->qos_depth;
+
+  struct subscriber_info * sub_info;
+  int bkt_sub_info;
+  hash_for_each(wrapper->topic->sub_info_htable, bkt_sub_info, sub_info, node)
+  {
+    if (sub_info->qos_depth > depth) depth = sub_info->qos_depth;
+  }
+
+  return depth;
+}
+
 static int release_msgs_to_meet_depth(
   struct topic_wrapper * wrapper, struct publisher_info * pub_info,
   union ioctl_publish_msg_args * ioctl_ret)
 {
   ioctl_ret->ret_released_num = 0;
 
-  if (pub_info->entries_num <= pub_info->qos_depth) {
+  const uint32_t depth = retention_depth(wrapper, pub_info);
+
+  if (pub_info->entries_num <= depth) {
     return 0;
   }
 
-  const uint32_t leak_warn_threshold = (pub_info->qos_depth <= 100)
-                                         ? 100 + pub_info->qos_depth
-                                         : pub_info->qos_depth * 2;  // This is rough value.
+  const uint32_t leak_warn_threshold =
+    (depth <= 100) ? 100 + depth : depth * 2;  // This is rough value.
   if (pub_info->entries_num > leak_warn_threshold) {
     dev_warn(
       agnocast_device,
@@ -1020,8 +1042,8 @@ static int release_msgs_to_meet_depth(
     return -ENODATA;
   }
 
-  // Number of entries exceeding qos_depth
-  uint32_t num_search_entries = pub_info->entries_num - pub_info->qos_depth;
+  // Number of entries exceeding the retention depth
+  uint32_t num_search_entries = pub_info->entries_num - depth;
 
   // NOTE:
   //   The searched message is either deleted or, if a reference count remains, is not deleted.
@@ -1030,10 +1052,8 @@ static int release_msgs_to_meet_depth(
   //
   // HACK:
   //   The current implementation only releases a maximum of MAX_RELEASE_NUM messages at a time, and
-  //   if there are more messages to release, qos_depth is temporarily not met.
-  //   However, it is rare for more than MAX_RELEASE_NUM messages that are out of qos_depth to be
-  //   unreferenced at a specific time. If this happens, as long as the publisher's qos_depth is
-  //   greater than the subscriber's qos_depth, this has little effect on system behavior.
+  //   if there are more messages to release, the retention depth is temporarily exceeded. That
+  //   costs memory only: what a subscriber is handed is bounded by its own window either way.
   while (num_search_entries > 0 && ioctl_ret->ret_released_num < MAX_RELEASE_NUM) {
     struct entry_node * en = container_of(node, struct entry_node, node);
     node = rb_next(node);
@@ -1064,8 +1084,8 @@ static int release_msgs_to_meet_depth(
     dev_dbg(
       agnocast_device,
       "Release oldest message in the publisher_info (id=$%d) of the topic "
-      "(topic_name=%s) with qos_depth=%d. (%s)\n",
-      pub_info->id, wrapper->key, pub_info->qos_depth, __func__);
+      "(topic_name=%s) with retention depth=%d. (%s)\n",
+      pub_info->id, wrapper->key, depth, __func__);
   }
 
   return 0;
--- a/agnocast_kmod/agnocast_ioctl.c
+++ b/agnocast_kmod/agnocast_ioctl.c
@@ -129,6 +129,19 @@ static bool domain_delivery_allowed(
   return false;
 }
 
+// Whether pub_info's publications reach sub_info at all. What a publish retains and whom it
+// signals both have to agree with what receive and take will hand over, so the three ask it here
+// rather than each spelling out the same pair of conditions.
+static bool sub_can_receive_from(
+  const struct topic_struct * topic, const struct publisher_info * pub_info,
+  const struct subscriber_info * sub_info)
+{
+  if (sub_info->ignore_local_publications && sub_info->pid == pub_info->pid) return false;
+
+  return domain_delivery_allowed(
+    topic, pub_info->domain_id, pub_info->is_bridge, sub_info->domain_id, sub_info->is_bridge);
+}
+
 static int add_topic(
   const char * topic_name, const struct ipc_namespace * ipc_ns, uint32_t domain_id,
   struct topic_wrapper ** wrapper)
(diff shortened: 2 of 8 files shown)


agnocast_kmod/agnocast_ioctl.c
```diff
@@ -336,6 +345,16 @@ static int insert_subscriber_info(
   // rebuild_notify_list() skips take subs by testing notify_ctx alone, so the two must agree.
   WARN_ON_ONCE(is_take_sub != (notify_ctx == NULL));
 
+  if (qos_depth > MAX_QOS_DEPTH) {
```
atsushi421, line 348: The bound is applied only to subscribers, but the A2R bridge copies the publisher's depth onto its Agnocast subscription (`agnocast_bridge_manager.cpp:624` via `get_publisher_qos()`), and `SubscriptionBase::initialize()` calls `exit(EXIT_FAILURE)` when the ioctl fails, so a publisher deeper than 4096 now kills the bridge manager. `cie_client_utils.cpp` already publishes `/agnocast_cie_thread_configurator/callback_group_info` with `KeepLast(5000)`, so every CIE deployment hits this. Consider bounding `insert_publisher_info()` the same way and either lowering `CIE_QOS_DEPTH` or clamping the depth the bridge inherits (which needs `MAX_QOS_DEPTH` mirrored into `agnocast_ioctl.hpp`). The version-label rationale should also mention that an existing ioctl argument gained a rejection range.

agnocast_kmod/agnocast.h
```diff
@@ -9,6 +9,12 @@
 #define MAX_TOPIC_LOCAL_ID 4096  // Bitmap size for per-entry subscriber reference tracking
 #define MAX_SUBSCRIBER_NUM \
   (MAX_TOPIC_LOCAL_ID - MAX_PUBLISHER_NUM)  // Maximum number of subscribers per topic
+/* Maximum subscriber qos_depth; a deeper request is refused. A subscriber's depth decides how long
+ * its publishers must hold entries open, so it spends memory in another process's mempool. The
+ * bound is well above what exists in practice -- the deepest subscriber in Autoware is 100
+ * (autoware_pointcloud_preprocessor's twist and odom), and 1000 across the whole tree -- so a
+ * request over it is a bug worth surfacing rather than silently narrowing. */
+#define MAX_QOS_DEPTH 4096
```
atsushi421, line 17: A live subscriber whose `latest_received_entry_id` stops advancing (executor not spinning, a take subscription polled only on another input, an a2r bridge under ROS 2 backpressure) now pins up to `min(lag, sub depth)` entries in every reachable publisher's mempool, so a subscriber-side stall becomes a publisher-side allocation failure. 4096 is far above what a mempool holds for large messages. It would help to document the sizing rule (a publisher's pool must cover the deepest reachable subscriber) and to emit a `dev_info` in `insert_subscriber_info()` when a subscriber's depth exceeds a reachable publisher's, next to the existing transient-local mismatch log. A cap relative to the publisher's own depth, or a module parameter, would bound the blast radius.

atsushi421 (commented):
The required e2e runs and `run_requires_kernel_module_tests.bash` are unchecked. Even when run, the existing e2e sweep cannot observe this change, since `test_1to1.py` publishes exactly `pub_qos_depth` messages, so the publisher never exceeds its own depth and `EXPECT_SUB_NUM = min(pub, sub)` is unaffected. Please run the required scripts and add an integration case where the publisher publishes more than its depth against a deeper subscriber (the pub 10 / sub 100 / 30 published case from the description), plus the mirror case where a deep subscriber that keeps up releases at the publisher depth.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -988,19 +1007,52 @@ int agnocast_ioctl_add_publisher(
...
+
+  if (pub_info->entries_num <= depth) {
     return 0;
   }
 
-  const uint32_t leak_warn_threshold = (pub_info->qos_depth <= 100)
-                                         ? 100 + pub_info->qos_depth
-                                         : pub_info->qos_depth * 2;  // This is rough value.
+  const uint32_t leak_warn_threshold =
+    (depth <= 100) ? 100 + depth : depth * 2;  // This is rough value.
```
atsushi421, line 1055: `leak_warn_threshold` is derived from the lag-bounded `depth`, which drops back to `pub_info->qos_depth` as soon as the deep subscriber catches up, but at that moment the subscriber may legitimately hold up to its own depth in references that the release loop skips. With pub 10 / sub 1000, `entries_num` can sit near 1010 against a threshold of 110, so the leak warning fires on every publish during normal operation. Deriving the threshold from the declared maximum (max of the publisher depth and the reachable subscriber depths) rather than the lag-bounded value would keep it consistent with what a subscriber can hold.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -988,19 +1007,52 @@ int agnocast_ioctl_add_publisher(
...
+// latest_received_entry_id.
+static uint32_t retention_depth(
+  struct topic_wrapper * wrapper, const struct publisher_info * pub_info)
+{
+  uint32_t depth = pub_info->qos_depth;
+  const int64_t newest_entry_id = wrapper->topic->current_entry_id - 1;
+
+  struct subscriber_info * sub_info;
+  int bkt_sub_info;
+  hash_for_each(wrapper->topic->sub_info_htable, bkt_sub_info, sub_info, node)
```
atsushi421, line 1027: This walks all 32 buckets plus every subscriber on each publish while holding `topic->rwsem` for write, the lock receive and take block on, whereas `rebuild_notify_list()` exists to keep per-subscriber work out of that section. A per-topic maximum of the declared subscriber depths, maintained on join and leave (both already under global write), would let `retention_depth()` return `pub_info->qos_depth` immediately when that maximum is not larger, which is the common case. The lag itself stays uncached. Either way, a one-line note here that the lag is deliberately not cached because it moves on every receive would keep the next reader from adding a cache.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -988,19 +1007,52 @@ int agnocast_ioctl_add_publisher(
   return ret;
 }
 
+// sub_info_htable is shared by the two domains a bridge rule pairs, so only a subscriber this
+// publisher can reach counts. And what one needs is its lag, not its declared depth: nothing at or
+// below latest_received_entry_id is handed over again, since receive_msg_core() starts one past it
+// and take_msg breaks below it. The lag counts the topic's entries while the depth bounds this
```
atsushi421, line 1013: The lag counts entry ids rather than entries. `insert_subscriber_info()` consumes an id for every volatile join (line 414) without creating an entry, so each later join adds one to the lag of every subscriber that has not advanced past it. Also, `take_msg` with `allow_same_message` can hand back the entry at exactly `latest_received_entry_id`, so the 'at or below' invariant has one exception (harmless, since any newer entry satisfies it). Both only over-estimate, but the comment should say so since it is the justification for the bound.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -336,6 +345,16 @@ static int insert_subscriber_info(
   // rebuild_notify_list() skips take subs by testing notify_ctx alone, so the two must agree.
   WARN_ON_ONCE(is_take_sub != (notify_ctx == NULL));
 
+  if (qos_depth > MAX_QOS_DEPTH) {
+    dev_warn(
+      agnocast_device,
+      "Subscriber qos_depth (%u) for the topic (topic_name=%s) exceeds the upper "
```
atsushi421, line 351: The warning omits `subscriber_pid` and `node_name`, both available here, and userland only prints `strerror(EINVAL)`, which is shared with the other `-EINVAL` returns of this ioctl. Adding both to the message, and ideally a pre-check in `validate_subscription_qos()` so the process log names the cause, would make the rejection diagnosable.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -988,19 +1007,52 @@ int agnocast_ioctl_add_publisher(
...
+  struct subscriber_info * sub_info;
+  int bkt_sub_info;
+  hash_for_each(wrapper->topic->sub_info_htable, bkt_sub_info, sub_info, node)
+  {
+    if (!sub_can_receive_from(wrapper->topic, pub_info, sub_info)) continue;
+
+    const int64_t lag = newest_entry_id - sub_info->latest_received_entry_id;
+    if (lag <= 0) continue;
+
+    const uint32_t wanted =
```
atsushi421, line 1034: This file already uses `min_t` and `max` from `<linux/minmax.h>`, so `min_t(int64_t, lag, sub_info->qos_depth)` and `depth = max(depth, wanted)` would replace the ternary and the explicit cast. The `lag <= 0` branch above is unreachable under the stated locking (the entry just inserted is always newer than any stored `latest_received_entry_id`), so it could go, or carry a word saying it is defensive.

agnocast_kmod/agnocast_ioctl.c
```diff
@@ -988,19 +1007,52 @@ int agnocast_ioctl_add_publisher(
...
+// publisher can reach counts. And what one needs is its lag, not its declared depth: nothing at or
+// below latest_received_entry_id is handed over again, since receive_msg_core() starts one past it
+// and take_msg breaks below it. The lag counts the topic's entries while the depth bounds this
+// publisher's, so it over-estimates on a multi-publisher topic -- the safe direction.
+//
+// Caller holds global_htables_rwsem (read), which excludes subscriber join and leave, and
+// wrapper->topic->rwsem (write), which excludes the receive and take that advance
+// latest_received_entry_id.
+static uint32_t retention_depth(
+  struct topic_wrapper * wrapper, const struct publisher_info * pub_info)
```
atsushi421, line 1020: `retention_depth()` only reads through `wrapper`, so the parameter can be `const struct topic_wrapper *`, matching `notifiable_subscriber_num()` and `is_entry_deliverable()`.

agnocast_kmod/agnocast_kunit/agnocast_kunit_publish_msg.c
```diff
@@ -69,6 +71,57 @@ static topic_local_id_t setup_one_subscriber_with_eventfd(
     test, 0, is_bridge, eventfd, ignore_local_publications);
 }
 
+// Joins an already registered process, so that a case can put the subscriber in the publisher's.
+static topic_local_id_t add_subscriber_with_qos_depth(
```
atsushi421, line 75: `add_subscriber_with_qos_depth()` and `setup_subscriber_with_qos_depth_in_domain()` duplicate `add_subscriber_with_eventfd()` and `setup_one_subscriber_in_domain_with_eventfd()` with one argument changed each. Adding a `sub_qos_depth` parameter to the existing pair (or a single `add_subscriber_impl()` as in the take and receive kunit files) would let all six wrappers become one-line forwards.

agnocast_kmod/agnocast_kunit/agnocast_kunit_get_subscriber_qos.c
```diff
@@ -98,3 +98,39 @@ void test_case_error_subscriber_not_found(struct kunit * test)
 
   KUNIT_EXPECT_EQ(test, ret, -EINVAL);
 }
+
+// Refused rather than narrowed: agnocastlib turns the failure into an RCLCPP_ERROR and exits, so
+// the depth the subscriber asked for is never silently different from the one it gets.
+void test_case_qos_depth_above_max_is_rejected(struct kunit * test)
```
atsushi421, line 104: This case only exercises `agnocast_ioctl_add_subscriber()` returning `-EINVAL` and never calls `get_subscriber_qos`, so it belongs next to `test_case_add_subscriber_too_many_subscribers` in `agnocast_kunit_add_subscriber.c`. The `at_max` case below is `verify_subscriber_qos()` with a different depth. Giving that helper a depth parameter would reduce it to one line.

agnocast_kmod/agnocast_kunit/agnocast_kunit_publish_msg.c
```diff
@@ -69,6 +71,57 @@ static topic_local_id_t setup_one_subscriber_with_eventfd(
     test, 0, is_bridge, eventfd, ignore_local_publications);
 }
 
+// Joins an already registered process, so that a case can put the subscriber in the publisher's.
```
atsushi421, line 74: The sentence ends on 'the publisher's' with no noun. Presumably 'the publisher's process'.

PR #1626, opened 2026-09-10
chore(agnocast): warn that the kmod domain bridge is unsupported

Koichi98:
## Description

Marks *registering* Agnocast domain bridge rules — the kmod cross-domain zero-copy path — as incomplete and unsupported, and warns wherever a rule is about to be registered. No behavior change: rules are still accepted and still relay.

Cross-domain delivery is refused unless a rule pairs the two cells (`domain_delivery_allowed`), so rule registration is the whole gate. The warnings sit there:

- **kmod** — `agnocast_ioctl_add_domain_bridge()` and `agnocast_ioctl_add_domain_bridge_prefix()` emit one `dev_warn_once`. Registering through these two is unavoidable, so this also covers a caller that bypasses the ioctl wrapper — including the prefix ioctl, which has no wrapper today.
- **`register_domain_bridge`** — prints the notice to stderr on every run, before the config is read, and carries it in `--help`.
- **`ros2agnocast_discovery_agent/systemd/README.md`** — the only in-tree how-to for registration; now leads with the notice.

Deliberately **not** warned: the discovery agent's read of the same YAML (`_load_domain_rules`). That read forces the A2R bridge the external `domain_bridge` node needs (`bridge_decider.py:296-297`, which emits `DIRECTION_AGNOCAST_TO_ROS2` requests only), so it supports the recommended configuration rather than the kmod path. The agent never registers a rule; its only ioctl calls are the exit gate (`agent.py:444`, `:449`).

## Related links

None.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

`ros2agnocast_discovery_agent`'s pytest suite passes (67 tests), with a new case for the tool's stderr notice. The kmod builds clean. KUnit was not run locally (the machine has `agnocast` loaded); it needs to pass in CI before this leaves draft.

## Notes for reviewers

The user-facing bridge page in [`agnocast_doc`](https://github.com/autowarefoundation/agnocast_doc) still describes the feature without the notice — a follow-up there is needed (see the post-merge checklist).

Also not covered on purpose: the ioctl wrapper (`add_agnocast_domain_bridge_rule`), which is redundant with the kmod warning, and the runtime delivery path, which is per-message under `global_htables_rwsem` and would flood dmesg.

## Post-merge checklist

### DIFF
--- a/agnocast_kmod/agnocast_ioctl.c
+++ b/agnocast_kmod/agnocast_ioctl.c
@@ -2724,8 +2724,8 @@ static void warn_domain_bridge_unsupported(void)
 {
   dev_warn_once(
     agnocast_device,
-    "Domain bridge rules are incomplete and unsupported. Use the external domain_bridge node "
-    "instead.\n");
+    "Registering domain bridge rules is incomplete and unsupported. Use the external "
+    "domain_bridge node instead.\n");
 }
 
 int agnocast_ioctl_add_domain_bridge(
--- a/agnocast_kmod/agnocast_ioctl.c
+++ b/agnocast_kmod/agnocast_ioctl.c
@@ -2717,10 +2717,23 @@ static int add_prefix_domain_rule(
   return insert_domain_rule(prefix, prefix, from_domain, to_domain, true, ipc_ns);
 }
 
+// Registration is the whole gate -- domain_delivery_allowed refuses a cross-domain pair that no
+// rule covers -- so warning in the two entry points below also reaches a caller that skips the
+// ioctl wrapper. Once per module load; insert_domain_rule already records each rule added.
+static void warn_domain_bridge_unsupported(void)
+{
+  dev_warn_once(
+    agnocast_device,
+    "Domain bridge rules are incomplete and unsupported. Use the external domain_bridge node "
+    "instead.\n");
+}
+
 int agnocast_ioctl_add_domain_bridge(
   const char * topic_name_from, const char * topic_name_to, const uint32_t from_domain,
   const uint32_t to_domain, const struct ipc_namespace * ipc_ns)
 {
+  warn_domain_bridge_unsupported();
+
   if (from_domain == to_domain) return -EINVAL;
 
   down_write(&global_htables_rwsem);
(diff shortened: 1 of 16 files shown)


agnocast_kmod/agnocast_ioctl.c
```diff
@@ -2717,10 +2717,23 @@ static int add_prefix_domain_rule(
...
 }
 
+// Registration is the whole gate -- domain_delivery_allowed refuses a cross-domain pair that no
+// rule covers -- so warning in the two entry points below also reaches a caller that skips the
+// ioctl wrapper. Once per module load; insert_domain_rule already records each rule added.
+static void warn_domain_bridge_unsupported(void)
+{
+  dev_warn_once(
+    agnocast_device,
+    "Domain bridge rules are incomplete and unsupported. Use the external domain_bridge node "
```
kobayu858: This is the one place that does not distinguish registering from the feature. The tool says "registering ... is unsupported" and the README says the agent's use of the same file is separate and supported, but dmesg gets "Domain bridge rules are incomplete and unsupported" with no context around it, which reads as the whole feature. "Registering domain bridge rules is ..." would match the other two.

src/ros2agnocast_discovery_agent/systemd/README.md
```diff
@@ -1,5 +1,12 @@
 # Registering domain bridge rules at boot
 
+> **Unsupported.** Registering rules with the kernel module — the kmod
+> cross-domain zero-copy path — is incomplete and not supported. Use the
+> external `domain_bridge` node to relay between ROS domains. The tool below
+> still works, but it and the kernel module both warn when a rule is registered.
```
kobayu858: `it and the kernel module both warn when a rule is registered` puts the two on the same footing, but the tool prints on every run while the kernel module's `dev_warn_once` fires once per module load. This page also describes registration as a one-time, re-runnable step, so an operator who re-runs the tool and greps dmesg to confirm sees nothing -- the one warning was spent on the first run. "the kernel module warns once per module load" would match the code.

agnocast_kmod/agnocast_ioctl.c
Koichi98: Fixed in 476cc7cb

src/ros2agnocast_discovery_agent/systemd/README.md
Koichi98: Fixed in f73c8540

kobayu858 (approved):

PR #1640, opened 2026-09-14
refactor(agnocastlib): introduce ServiceBase

bdm-k:
## Description

This PR introduces `ServiceBase` for shared service state and helpers. Both service implementations now inherit from `ServiceBase`. It also renames the typed service implementation from `BasicService<ServiceT>` to `Service<ServiceT>`, which is an old workaround that is no longer relevant.

## Related links

This is a redo of #1482.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Post-merge checklist


src/agnocastlib/include/agnocast/agnocast_service.hpp
```diff
@@ -47,9 +47,39 @@ constexpr PublisherRole to_publisher_role(const ServiceRole role)
...
+
+  template <typename NodeT>
+  explicit ServiceBase(
+    NodeT * node, const std::string & service_name, const rclcpp::QoS & qos, ServiceRole role)
+  : node_(node), qos_(rclcpp::QoS(qos).durability_volatile()), role_(role)
+  {
+    service_name_ = node->get_node_services_interface()->resolve_service_name(service_name);
+  }
+
+  rclcpp::Logger get_logger() const;
```
Koichi98, line 67: Please keep the `// Defined in the .cpp: agnocast::Node is only forward-declared here.` comment that was here on main — inlining this body compiles fine in a TU that also includes `agnocast_node.hpp` but fails in one that only includes `agnocast_service.hpp`, so the constraint is easy to break without the note.

src/agnocastlib/include/agnocast/agnocast_service.hpp
```diff
@@ -47,9 +47,39 @@ constexpr PublisherRole to_publisher_role(const ServiceRole role)
...
+  virtual ~ServiceBase() = default;
+};
+
+/**
+ * @brief Service server for zero-copy Agnocast service communication.
+ * @tparam ServiceT The ROS service type (e.g., std_srvs::srv::SetBool).
+ */
 template <typename ServiceT>
-class BasicService : public std::enable_shared_from_this<BasicService<ServiceT>>
+class Service : public ServiceBase, public std::enable_shared_from_this<Service<ServiceT>>
```
Koichi98, line 84: `Service` lost the `AGNOCAST_PUBLIC` marker it had on the alias; its counterpart `Client` still carries it at `agnocast_client.hpp:150`.

src/agnocastlib/include/agnocast/agnocast_service.hpp
bdm-k: fixed

src/agnocastlib/include/agnocast/agnocast_service.hpp
bdm-k: fixed

Koichi98 (approved):

PR #1641, opened 2026-09-15
fix(bridge): request buffering in A2R service bridge

### LINKED ISSUE #1576: Retry an undeliverable A2R service bridge request instead of dropping it
An A2R service bridge with no matching ROS 2 service on the graph cannot forward the request it just received, and every path that gives up on one today simply returns (`agnocast_bridge_loader.cpp:411-416,426`, `service_bridge_plugin.cpp.em:96`), leaving the calling Agnocast client's future pending for the life of the client. It should hold the request and retry from a bounded queue drained by a timer on `client_cb_group` — holding one across callbacks is what the deferred-response API already prescribes (`agnocast_service.hpp:237-238`), and draining it off the request path keeps the service callback from blocking its own callback group. The queue must be capped and released on teardown, since a referenced entry is skipped by `release_msgs_to_meet_depth()` rather than freed and `entries_num` past `100 + qos_depth` is logged as a refcount leak (`agnocast_ioctl.c:952-962,1005`).

bdm-k:
closes #1576

## Description

This PR adds request buffering in A2R service bridge to address the issue described in #1576: requests must be sent only after the ROS 2 service becomes ready; otherwise, the request may be silently dropped and the calling client's future will not be resolved forever. The direct solution would be calling `wait_for_service()` in the service callback, but blocking inside a callback is not idiomatic.

This PR takes another approach and introduces request buffering. Requests are first placed in a certain buffer. Then, either the **service callback** or **timer callback** processes the buffer, sending requests in it one by one if the ROS 2 service is ready. The service callback and timer callback share the same `MutuallyExclusive` callback group, so no extra synchronization is needed.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Notes for reviewers

The service callback group now also contains the timer callback, which is a ROS 2 callback. This results in the "contains both ROS 2 callbacks and Agnocast callbacks" warning message. Any ideas on how to avoid this?

## Post-merge checklist


src/agnocastlib/src/bridge/agnocast_bridge_loader.cpp
```diff
@@ -378,58 +380,114 @@ ServiceBridgeEntity BridgeLoader::create_a2r_service_bridge_generic(
...
+                       const std::shared_ptr<agnocast::GenericService> & service_handle,
+                       const std::shared_ptr<std::vector<ipc_shared_ptr<void>>> & vec,
+                       ipc_shared_ptr<void> && agno_req) {
+    if (vec->size() < kMaxPendingRequests) {
+      vec->push_back(std::move(agno_req));
+      return;
+    }
+    vec->erase(vec->begin());
+    vec->push_back(std::move(agno_req));
+    RCLCPP_ERROR(
```
Koichi98, line 397: This is the only place that actually drops a request and leaves the caller's future unresolved forever, so the message should say so like the other drop sites at L426 and L443 do (same in `service_bridge_plugin.cpp.em:76`).

src/agnocastlib/src/bridge/agnocast_bridge_loader.cpp
```diff
@@ -378,58 +380,114 @@ ServiceBridgeEntity BridgeLoader::create_a2r_service_bridge_generic(
...
+      const std::shared_ptr<agnocast::GenericService> & service_handle,
+      const std::shared_ptr<agnocast::vendor_rclcpp::GenericClient> & ros_client,
+      const std::function<bool(const ipc_shared_ptr<void> &, const std::shared_ptr<void> &)> &
+        res_copier,
+      const std::shared_ptr<std::vector<ipc_shared_ptr<void>>> & requests) {
+      // This try/catch prevents exceptions from async_send_request() from escaping the spin thread
+      // and terminating the process.
+      try {
+        while (!requests->empty()) {
+          ipc_shared_ptr<void> agno_req = std::move(requests->back());
```
Koichi98: This drains the buffer newest-first while the overflow path at L393 drops the oldest, so the two ends disagree; a `std::deque` with `front()`/`pop_front()` would make both FIFO (same in `service_bridge_plugin.cpp.em:93`).

src/agnocastlib/src/bridge/agnocast_bridge_loader.cpp
```diff
@@ -378,58 +380,114 @@ ServiceBridgeEntity BridgeLoader::create_a2r_service_bridge_generic(
...
+                service_handle->cancel_response(std::move(agno_req_movable), std::move(agno_res));
+                return;
+              }
+
+              auto agno_req_movable = agno_req;  // Resort to pointer copying to move from the const
+                                                 // lambda capture.
+              service_handle->send_response(std::move(agno_req_movable), std::move(agno_res));
+            });
+        }
+      } catch (const std::exception & e) {
```
Koichi98: The request popped at L413 is already out of the buffer when this catch runs, so it is lost with no response; pushing it back here would let the timer retry it (same in `service_bridge_plugin.cpp.em:108`).

src/agnocastlib/src/bridge/agnocast_bridge_loader.cpp
```diff
@@ -378,58 +380,114 @@ ServiceBridgeEntity BridgeLoader::create_a2r_service_bridge_generic(
...
-          logger, "Failed to forward request in A2R service bridge for '%s': %s; dropping request",
-          service_handle->get_service_name(), e.what());
+      if (ros_client->service_is_ready()) {
+        send_requests(service_handle, ros_client, response_copier, pending_requests);
       }
     },
     qos, srv_cbg, agnocast::ServiceRole::BridgeInternal);
 
-  return {agno_srv, srv_cbg, client_cbg};
+  auto drain_requests_timer = node->create_wall_timer(
```
Koichi98: Using `agnocast::create_timer()` instead of `create_wall_timer()` keeps the timer out of the rclcpp callback group, so the mixed-callback warning never fires and the group still needs no extra synchronization (same in `service_bridge_plugin.cpp.em:150`).

src/agnocastlib/src/bridge/agnocast_bridge_loader.cpp
```diff
@@ -378,58 +380,114 @@ ServiceBridgeEntity BridgeLoader::create_a2r_service_bridge_generic(
...
+    std::chrono::milliseconds(200),
```
Koichi98: `create_callback_group_id()` encodes rclcpp timers as `Timer(<period>)`, so this changes every A2R service bridge's callback group ID and any exact-match thread configurator entry silently stops applying.

src/agnocastlib/src/bridge/agnocast_bridge_loader.cpp
```diff
@@ -378,58 +380,114 @@ ServiceBridgeEntity BridgeLoader::create_a2r_service_bridge_generic(
...
+    [send_requests, response_copier, agno_srv, ros_client, pending_requests]() {
+      if (ros_client->service_is_ready()) {
```
Koichi98: `service_is_ready()` is an rmw graph query that runs at 5 Hz per A2R bridge even when nothing is buffered, so it is worth guarding with `!pending_requests->empty() &&` (same in `service_bridge_plugin.cpp.em:153`).

src/agnocastlib/src/bridge/agnocast_bridge_loader.cpp
```diff
@@ -378,58 +380,114 @@ ServiceBridgeEntity BridgeLoader::create_a2r_service_bridge_generic(
...
-
-            auto agno_req_movable = agno_req;  // Resort to pointer copying to move from the const
-                                               // lambda capture.
-            service_handle->send_response(std::move(agno_req_movable), std::move(agno_res));
-          });
-      } catch (const std::exception & e) {
-        RCLCPP_ERROR(
-          logger, "Failed to forward request in A2R service bridge for '%s': %s; dropping request",
-          service_handle->get_service_name(), e.what());
+      if (ros_client->service_is_ready()) {
```
Koichi98: `service_is_ready()` throws whenever `rcl_service_server_is_available()` returns non-OK, and unlike `async_send_request()` it sits outside the try/catch, so the exception escapes the CIE child thread and terminates the process (same at L482 and in `service_bridge_plugin.cpp.em:144,153`).

src/agnocastlib/src/bridge/agnocast_bridge_loader.cpp
bdm-k: Sorry, I'm not sure I follow. Has the problem been resolved with `agnocast::create_time()`?

src/agnocastlib/src/bridge/agnocast_bridge_loader.cpp
bdm-k: fixed

src/agnocastlib/src/bridge/agnocast_bridge_loader.cpp
bdm-k: fixed

src/agnocastlib/src/bridge/agnocast_bridge_loader.cpp
bdm-k: used `agnocast::create_timer()`

src/agnocastlib/src/bridge/agnocast_bridge_loader.cpp
bdm-k: Fixed. When `async_send_request()` fails, the request is pushed back to the front of the buffer, and `send_requests()` returns immediately. In the next run, the request will be retried.

src/agnocastlib/src/bridge/agnocast_bridge_loader.cpp
bdm-k: used `std::deque`

src/agnocastlib/src/bridge/agnocast_bridge_loader.cpp
bdm-k: fixed

src/agnocastlib/src/bridge/agnocast_bridge_loader.cpp
Koichi98: Yes — an agnocast timer is not in the rclcpp callback group, so the ID is unchanged from before this PR.

Koichi98:
@bdm-k 
I've fixed the clang-tidy errors in
https://github.com/autowarefoundation/agnocast/pull/1641/commits/b8760f4da7c5df139d1e261a50160db586d583b7.

Koichi98 (approved):

PR #1646, opened 2026-09-18
refactor(agnocastlib): extract response publisher management into reusable class

bdm-k:
## Description

Each service class contained their own copy of the same response-publisher management code; they diverged only in the publisher type and the constructor arguments the publisher needed. This PR extracts the response-publisher management into a reusable `detail::ResponsePublisherManager` class:
```cpp
template <typename PublisherT>
class ResponsePublisherManager
{
  std::mutex mtx_;
  std::unordered_map<std::string, typename PublisherT::SharedPtr> pubs_;

public:
  typename PublisherT::SharedPtr get_or_create_publisher_for(
    const ServiceBase * service_base, const std::string & response_topic_name);
};
```

## Related links

This is a redo of #1482.

## How was this PR tested?

- [ ] Autoware
- [ ] sample application

## Post-merge checklist


src/agnocastlib/include/agnocast/agnocast_service.hpp
```diff
@@ -47,8 +47,18 @@ constexpr PublisherRole to_publisher_role(const ServiceRole role)
...
+namespace detail
+{
+template <typename>
+class ResponsePublisherManager;
+}  // namespace detail
+
 class ServiceBase
 {
+  template <typename>
+  friend class detail::ResponsePublisherManager;
```
Koichi98: Passing `node_`, `qos_`, and `role_` to the manager's constructor might let us drop the `friend` declaration here.

